PR #20, opened 2017-08-07
SPIDER: Regional Transit Authority

jim:
I'm posting this to get some early feedback and raise some potential issues that other scrapers are likely to encounter:

* The meeting time is currently hardcoded, which is not ideal but I think OK for our immediate purposes. I think this scraper will need to do the same with the location as it is not included in an easily parsable location and is unlikely to change. Ideally we could provide some guidance on handling this situation for other devs.

* The RTA site embeds the meeting schedule using an iframe, so the tests are currently just using the source of the iframe to test item extraction. In order to obtain more thorough coverage, we'll need to consider using a different testing strategy that can stub out Requests created during the test run.

    Scrapy supports Downloader middleware, so either the HttpCacheMiddleware
or a simple custom one will probably do the job. We should be able to configure this to only be enabled during testing.

I'd also like to note that I don't do a lot of Python development, so feedback on any stylistic issues is welcomed and encouraged! 😄


eads:
Thanks, I'll take a look! Worth noting there are a few stylistic things you can fix that were [caught by automated testing](https://travis-ci.org/City-Bureau/documenters-aggregator/builds/261853021?utm_source=github_status&utm_medium=notification).

jim:
@eads Thanks for bringing those to my attention. I now know about `pyflakes` and `pep8`!

eads:
Also, let's review this tomorrow!

documenters_aggregator/spiders/rta.py
```diff
@@ -0,0 +1,112 @@
...
+        request = scrapy.Request(url, callback = self.parse_iframe)
+        request.meta['description'] = description
+
+        yield request
+
+    def _parse_id(self, item):
+        """
+        There is no publicly exposed ID.
+        """
+        return None
```
eads: This is a blocker for merging -- all events need a unique identifier.

documenters_aggregator/spiders/rta.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def _parse_start(self, item):
+        """
+        Combine start time with year, month, and day.
+        """
+        title = item.css('.committee::text').extract_first()
+        m = re.search('(\d{4})-(\d{1,2})-(\d{1,2})', title)
+        tz = timezone('America/Chicago')
+        naive_dt = datetime(int(m.group(1)), int(m.group(2)), int(m.group(3)), 8, 30)
+        dt = tz.localize(naive_dt, is_dst=False)
```
eads: I'm curious why you're setting is_dst to false -- don't we want pytz to determine dst based on day of year for us?

jim: I think you're right. I'll update the code to reflect this.

documenters_aggregator/spiders/rta.py
jim: I'm planning to generate an identifier from the date, time, and title, which will hopefully be unique enough.

jim:
@eads I made the requested changes, and I think this is good to go now.

PR #30, opened 2017-08-08
CCHHS scraper, tests fail

r-wei:
The tests fail, but the expected output is correct when I run `scrapy crawl cchhs`... not sure why... The only thing not working properly is something with time zones and daylight savings (UTC-5 vs UTC-6)

Some decisions I made:
- `status = passed` if the start_date is in the past (so I pass the start date to the _parse_status function)
- The description is a link to the agenda (which is a pdf). It looks like agendas are usually posted 1 week before the meeting. I put `status = confirmed` only if an agenda has been posted.


eads:
We figured out what was wrong! The test file had an HTML comment before the opening `<html>` tag, which isn't quite kosher.

.gitignore
```diff
@@ -86,6 +86,7 @@ celerybeat-schedule
 .venv
 venv/
 ENV/
+documenters-aggregator/
```
eads: I'm curious if it's wise to use this as the name of the virtualenv. I'd find it confusing myself. But I think it's likely others may name it similarly, so a good idea to put it in the gitignore.

documenters_aggregator/spiders/cchhs.py
```diff
@@ -0,0 +1,149 @@
...
+                'id': self._parse_id(item),
+                'name': self._parse_name(item),
+                'end_time': self._parse_end(item),
+                'all_day': self._parse_all_day(item)
+            }
+
+            aria_control = item.xpath("@aria-controls").extract_first()
+            item_uncollapsed = item.xpath("//div[@id='{}']//tbody//td[@data-title='Meeting Information']".format(aria_control))
+            for subitem in item_uncollapsed:
+                data.update({
```
eads: I'm not crazy about updating the same dict over and over here. But the flow is very clear for something relatively hairy.

documenters_aggregator/spiders/cchhs.py
```diff
@@ -0,0 +1,149 @@
...
+        passed = meeting already started
+        tentative = no agenda posted
+        confirmed = agenda posted
+        Is this the right way to determine tentative vs confirmed?
+        Looks like agendas are usually posted a week before the meeting
+        """
+        if datetime.now().isoformat() > start_time:
+            return 'passed'
+        agenda = subitem.xpath("following-sibling::td/a/@href").extract_first()
+        if agenda is None:
```
eads: Love this as a confirmation criteria.

documenters_aggregator/spiders/cchhs.py
r-wei: Making this change + using curl fixed the failing tests!

eads:
Huzzah!

PR #41, opened 2017-08-09
resolve #35

### LINKED ISSUE #35: can/should genspider pull an example html file
@r-wei suggested that the `invoke genspider` command should pull a sample file automatically for testing, which is very wise but will require some updates to the command.

r-wei:
The html files downloaded via requests.get(url).content and curl are slightly different...

If there are multiple starturls, I name the html files spiderName_partOfUrlAfterLastSlash.html


eads:
What's the difference between curl and requests.get(...)? 

requests.get(url).content returns bytes, whereas requests.get(url).text returns encoded text, so that might explain it...

r-wei:
Ah ok. Yes, using requests.get(url).text and curl gives the same output. I wrote some tests to verify it. The files that end in .example were downloaded using curl.

tasks.py
```diff
@@ -1,18 +1,20 @@
...
 from jinja2 import Environment, FileSystemLoader
+import requests
 
 import os
 
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
 SPIDERS_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'documenters_aggregator/spiders')
 TESTS_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'tests')
+FILES_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'tests/files')
```
eads: You could also do:

```
FILES_DIR = os.path.join(TESTS_DIR, 'files')
```

Either way is good, just pointing it out.

tasks.py
```diff
@@ -22,6 +24,11 @@ def genspider(ctx, name, domain):
     test_filename = _gen_tests(name, domain)
     print('Created {0}'.format(test_filename))
 
+    if start_urls:
+        html_filenames = _gen_html(name, start_urls)
```
eads: This is mostly my own curiosity -- does `invoke` automagically expand a comma separated list into a a Python list when passed as an argument?

tasks.py
```diff
@@ -43,6 +50,31 @@ def _gen_tests(name, domain):
...
+        except requests.exceptions.RequestException as e:
+            print(e)
+            continue
+
+        content = r.text.strip()
+
+        url_suffix = url.split('/')[-1]
+        if '.' in url_suffix:
+            url_suffix = url_suffix.split('.')[-2]
+        filename = '{0}/{1}_{2}.html'.format(FILES_DIR, name, url_suffix)
```
eads: Not a worry now, but what if the file isn't HTML?

tasks.py
```diff
@@ -43,6 +50,31 @@ def _gen_tests(name, domain):
...
+
+        with open(filename, 'w') as f:
+            f.write(content)
+
+        files += [filename]
```
eads: I think a more pythonic and readable way to do this would be:

```
files.append(filename)
```

tests/test_tasks.py
```diff
@@ -22,4 +26,22 @@ def test_render_spider():
     assert test_file_content == rendered_content
 
 
+def test_gen_html():
```
eads: Generally speaking, if you can split out your assertions into multiple tests, then [that's probably a good idea](https://softwareengineering.stackexchange.com/a/7829). This feels like a place where you could do that.

tasks.py
r-wei: Thanks for catching that. Looks like invoke doesn't take lists as arguments yet. I changed genspider to receive a { separated string of urls. I don't know if { is the best choice--just tried to pick a character that doesn't appear much in urls.

tests/test_tasks.py
r-wei: I separated them out :)

tasks.py
eads: Heh, what about `|`? Comma could work as well -- any commas in URLs could be required to be escaped. That said, somebody's probably gonna use an unescaped comma....

r-wei: Unfortunately `|` gets read as the pipe command in bash... I could look into seeing if there's a way around that. Oh! maybe we could do commas, and then search for the string `,http://`

eads: Ha, right! Damn. You could require the list to be quoted, as well... `"http://espn.com|http://npr.org"`... 

I do like splitting on `,http` -- I think you'd want that rather than `,http://` because then it would include https.

r-wei: splitting on `,http` now :)

PR #42, opened 2017-08-14
WIP SPIDER: Chicago Park District Board of Commissioners 

bergren2:
Wanted to get what I had into a PR in case I don't have time to finish it before going on vacation this week.

The DOM that gets shown in Chrome is very different than what's seen by the spider, so I've been using `curl` to grab the page and plopping it into a test file so I can view locally:

    $ curl https://chicagoparkdistrict.legistar.com/Calendar.aspx > test.html

There's also a copy of the DOM in the response file for the tests.


eads:
You can also use `scrapy view <url>` to open the url as the crawler sees it in a browser (on most systems this should Just Work but some recent versions of OS X seem to have a problem with this) and `scrapy crawl <url>` to work with the response object in an interactive shell.

bergren2:
Closing this one in favor of #168.

PR #48, opened 2017-08-14
Add runtests invoke task

jim:
I've added an invoke task that runs the tests and all style checkers in order to help developers follow the style guidelines used by the project. This should also help prevent the situation where a developer runs the tests locally and sees them pass only to have CI fail due to a style issue.

If there is a better name than `runtests`, I'm all ears!


eads:
I was thinking of adding tox, but this is way better and more elegant. 👍

Issue #46, opened 2017-08-14
what to do about files genspider tests downloads


eads:
@r-wei 

The genspider tests leave new files in the tests directory, somewhat out of necessity. Not a big deal but worth noting.

we could:

* ignore completely
* remove when finished
* do some mock stuff so they never get written

jim:
If we can't find a better solution, writing the downloaded files to a `tmp` directory and adding that dir to `.gitignore` would be OK. Right now running the tests leaves the repo in a dirty state, which is not a good experience for new developers to the project.

r-wei:
I'll fix this

PR #57, opened 2017-08-15
moved genspider html files to tmp/

r-wei:
resolves 46 according to Jim's suggestion. If the tests pass, then the html files created from genspider get moved to files/tmp/  (It doesn't make sense to move the files if the tests don't pass because then something went wrong in downloading or naming the files.)


jim:
Do you think it would make sense to just delete the files instead of moving them? Do we need them for anything after the test runs?

eads:
++ @jim -- seems cleanest just to delete when done to me.

r-wei:
sounds good. I updated the code to delete them.

eads:
whoops, closes #46

PR #63, opened 2017-08-16
WIP SPIDER: Chicago Public Library

maxine:
#### Not ready to be merged yet, but so far:
Since the [Chicago Public Library board events page](https://www.chipublib.org/board-of-directors/board-meeting-schedule/) is all in HTML, my spider structure is a bit different to accommodate/avoid parsing strings every time we want to extract any information from the page. Currently, code in the main `parse` function turns the HTML glob into an array of event objects, which are each an array of (HTML tag-free) strings, one per line on the CPL events page.
#### Notes on this structure:
* The first string in an event object is *always* the date
* The last line is *always* the address
* If the event has 4 lines, then line 2 and 3 should be concatenated to be the location
* If the event has 3 lines, then the middle line is the location
#### To do to finish:
* Figure out a way to create a unique ID for each event
* Get date/time in correct format
* Restructure tests to work for modified spider structure
* Agree upon a classification (board meeting?) for the events


jim:
@maxine Awesome!

If there isn't an event ID exposed, you can use the date and title to generate one. [Here is how I did it in the RTA scraper](https://github.com/City-Bureau/documenters-aggregator/blob/master/documenters_aggregator/spiders/rta.py#L53). Hopefully the date and title combo will be unique enough for our purposes.

maxine:
This is all ready to go! Tests all pass.

r-wei:
Hey @maxine, looks great.

I made a few quick changes to your code:
- removed numpy and pandas dependencies--the same thing can be accomplished with urllib and json, and won't require everyone having to download these somewhat bulky packages
- merged in the master branch--there have been some changes to the main project since you last synced with it, so I updated your code to make it consistent
- pep8 style fixes--little things like indentation and whitespace formatting

The only remaining issue is that the start dates are all in the year 1900 instead of 2017. Do you want to take a crack at fixing that? I think it should only take a couple lines of code. To make the fix, you should first pull your branch to get up-to-date my commits--let me know if you need more elaboration on that and sorry if these are things you already know.

After fixing the year, I think this is good to merge. :)

eads:
Excellent review @r-wei!

maxine:
Thanks so much @r-wei ! I fixed the year issue so now the start date should show up with the correct year instead of 1900.

eads:
This is so close! What's the word with the status field?

r-wei:
The status field test is failing for a different spider (which I built ^^;;). It's failing because, a week ago, I picked a meeting in the future whose status was tentative. But now the meeting has passed so the spider is (correctly) labeling it as "passed", but the test hasn't been updated. I'll fix this when I get in city bureau. :)

PR #65, opened 2017-08-16
Retry attempts to fetch a URL.

jim:
Our intermittent test failures are the result of the citybureau.org
server returning a 429 response when a test attempts to fetch some
HTML. In an effort to make our tests solidly green and remove the
intermittent failures, this commit adds retrying functionality to the
_gen_html function. We retry up to 5 times, with an exponentially increasing
backoff (2, 4, 8, 16, 32 seconds).

Please note that this is a temporary fix, as ultimately we should not be hitting
the network to run these tests. I can think of two solutions:

1. Have two test suites, one for local unit tests and one that tests
   remote systems.
2. Use a library to mock out these requests so we don't rely on the
   network.

In the short term, however, I think this fix will improve the experience
for new developers.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -3,7 +3,7 @@ from jinja2 import Environment, FileSystemLoader
 import requests
 
 import os
-
+import time
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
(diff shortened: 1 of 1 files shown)


eads:
I like option #1 because it is nice to have live network tests in some cases. I'm not sure how to do it with pytest, but when we used nose2 for elex, we would set an environment variable like `ELEX_NETWORK_TESTS=1` which triggered running the network tests in addition to all the offline tests.

Issue #10, opened 2017-08-01
document data model


eads:

r-wei:
we have an events_schema.md now

Issue #31, opened 2017-08-08
templates and existing scrapers need "timezone" field in returned data


eads:

jim:
For this, do we need to have the timezone name or just the offset? If just the offset, we could extract it from the ISO8601 times (`2017-08-24T08:30:00-05:00`). Or do we need the `timezone` value to be "America/Chicago"?

r-wei:
no longer needed

Issue #29, opened 2017-08-08
what should the date format be? 


eads:
The OpenCivicData spec is unclear on this point. To wit, "Starting date / time of the event. This should be fully timezone qualified." 

What does that even mean?

jim:
@eads I think I'd go with ISO 8601 (e.g. `2017-08-14T15:43:39+00:00`) unless there is a good reason to use something else.

eads:
@jim I'm with that. In the absence of more detail in the Open Civic Data spec, ISO format seems like the right way to go. Especially because libraries like pytz can calculate the correct timezone offset based on the date and Olsen timezone.

I'll add this to the docs and close when that's done.

jim:
Turns out the OCD docs mention that the datetimes should be in UTC and use the Z timezone designator (indicating that they are in plain UTC) ([source](http://docs.opencivicdata.org/en/latest/proposals/0004.html#core-fields)):

```
Starting time of the event. This field is serialized as an ISO 8601 datetime string, normalized to UTC, and must be fully datetime aware. ISO 8601 defines times expressed in UTC with a special UTC designator (“Z”) as the timezone.
```

I added this in #73, so this can be closed when that is merged.

PR #73, opened 2017-08-23
Document event schema and writing tests

jim:
I added a description of our event schema to the docs. We already have it in the comments from the generator and it is available on the [Open Civic Data site](http://docs.opencivicdata.org/en/latest/data/event.html), but I thought it made sense to have it in our docs for ease of access.

I also added some words about the dev process and opening a pull request. It's a start, and I imagine we'll continue to develop that content over time.

### DIFF
--- a/docs/contribute-a-spider.md
+++ b/docs/contribute-a-spider.md
@@ -177,4 +177,53 @@ class ChaSpider(scrapy.Spider):
 
 ## Write tests
 
-TK TK writing tests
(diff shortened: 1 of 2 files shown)


jim:
This should resolve #10 and #29.

jim:
Also addresses #31.

eads:
Excellent. I want to do some more editing and work on this, as well, so thanks for pushing it forward.

PR #85, opened 2017-08-30
Add GeocoderPipeline

jim:
Two things in this PR that are notable:

1. An initial stab at a `GeocoderPipeline` that will fill in the lat/long coordinates on events when they are missing.
2. Configuration for `betamax`, which records HTTP requests and saves them to `tests/cassettes`, which are then used to stub out network responses during future test runs. We can investigate using this in other places in the test suite where it would be useful.
3. (new) I stubbed out the network requests done by out task tests, so tests run much faster, are more reliable, and no longer depend upon having network access. I'm not a huge fan of passing around the `requests.session` as much as we are now (it feels like a missing object to me), but for now this is probably OK.


tasks.py
```diff
@@ -77,13 +77,15 @@ def _gen_tests(name, domain):
     return filename
 
 
-def _fetch_url(url, attempt=1):
+def _fetch_url(url, attempt=1, session=requests.Session()):
     """
     Attempt to fetch the specified url. If the request fails, retry it with an
     exponential backoff up to 5 times.
```
r-wei: Where does re-trying 5 times happen? Looks like the attempt variable is unused.

r-wei:
The code looks fine to me although I haven't worked with the geocoder or betamax packages before. 

Is the purpose of using betamax and sessions so that the downloaded content is more standardized from run to run? I looks to me like `session.get` is still called every time the tests (_fetch_url) are run, so I'm not sure how this avoids needing an Internet (network?) connection.

I know that geocoders can sometimes be computationally costly / slow. Maybe we should check how long it takes this geocoder to geocode a batch of addresses.

jim:
@r-wei betamax has a neat behavior where the first time it runs, network requests are made as usual, but whatever responses are received are then saved to the disk. The next time the same request is made (usually when the test is run the next time), no new network requests are made. Instead, the recorded responses from the disk are played back in a way that looks to most of the code like they are real responses. In this way, you can test in a realistic way but not actually need a network connection. It also insulates the tests from server behavior, like we were seeing where sometimes the citybureau.org server would return a 429 response for no real reason.

The betamax docs probably explain this better than I do. In particular, [this page](https://pypi.python.org/pypi/betamax#what-does-it-even-do) gives a decent overview.

Most Ruby project I've worked on have used something similar to this to emulate network activity but keep tests deterministic. It seems to be a decent solution when there aren't other ways to mock out requests.

jim:
I should also mention that I *did not* spend any time thinking about which geocoding service we should use. The nice thing about the geocoder library is that it abstracts these details away, enabling us to switch service by just changing a method call or two.

I'm thinking of using the Mapbox API for this project since I know a little about their other tools, but I am in no way opposed to using a different one.

eads:
The code itself looks clean and nice.

I kind of avoided Betamax a few years ago (I think around the time it came out) because I found it a bit buggy and the interface too cute by half. But it seems pretty clearly the best solution for the job at this point, my curmudgeonly aesthetics aside. 

As far as geocoding APIs, Google's is the best and it ain't even close. We'd have to pay for Mapbox, and while their service overall is splendid, it's definitely not a good value proposition for just geocoding. We used the Mapbox geocoder at NPR and had to give up and switch back to Google. 

Still, I'm not clear that we're allowed to store addresses and lat/lngs from Google's geocoder under their terms of service. I did some research and found conflicting answers. I'm going to keep looking into this.

There's also a 2,500 per day limit on geocoding via Google's free offering. That _should_ be fine but something to keep in mind for the current architecture of the thing. Leaving geocoding as a separate pipeline is wise, but if we start bumping up against limits and don't have resources (either in form of donated API access or cash money) to continue, we'll probably want to skip previously geocoded items, which means putting the geocoding more directly into the data storage pipeline(s).

eads:
It's still not clear to me whether storing geocoded addresses would violate the Google terms of service. On the one hand, addresses and lat/lngs are both facts (which you can't copyright) and mostly derived from public sources. On the other hand, Google seems pretty clear you can't store content from the maps API long term, so the question is whether lat/lng and address count as content. I think we're probably safe but want to be extra sure. 

However, this provision is still around:

screen shot 2017-09-05 at 2 55 42 pm

That's a problem for us unless our output involves a Google map, which I wouldn't want to require (worth asking @diaholliday but I can already guess his answer). 

I've reached out to a contact at Google to see what they say. For now, I think we have to assume we'll need something else.

eads:
Other options I think could be viable:

* [Geocodio](https://geocod.io/): Not [currently supported](https://github.com/DenisCarriere/geocoder/issues/127) by Python Geocoder. But we could fix that. They have a 2,500/day free plan. They also add political boundaries. There is a [dedicated Python library](https://github.com/bennylope/pygeocodio), at least.
* MapBox: Caveats from my experience apply. However, could be fine for what we need to do and a company that's continually improving their product. Their free version allows for 50k geocoder requests per month or about 1,666 per day.
* Texas A&M geoservices: This thing is lovely and also more than a little clunky. The main downside of API is you have to split natural language addresses into street/city/state components. The other main downside is you have to sign up as a partner and continually re-up your allotment.

There's many others. I'm leaning towards Mapbox, though. I think it's accurate enough for our purposes even if not for geocoding a really messy batch of open data and I know folks over there if we do run into issues. Geocodio looks pretty damn compelling as well, minus the Python geocoder support.

eads:
Added #95; will implement mapbox for now and merge.

eads:
Tried Mapbox again, it choked on the basic test case :( :( ... "room 201A" and "city hall" messed it up terribly and even with the precise address, it comes up with a lat/lng that puts you in the middle of the street in front of city hall. Will try geocodio.

eads:
Also tried Geocodio, Mapquest, and Bing. Mapquest was the most accurate/robust for a random assortment of addresses I entered, though Bing and Mapquest were both close -- but none as good as Google at properly geolocating city hall, specifically. Bing and Geocodio did well with the address and poorly with "city hall" and vice versa for Mapquest. 

The weirdest part is you put the address for city hall (121 n lasalle) into Bing's web ui and it nails it. You put the same address (after bing itself has corrected it) into their geocoder and you wind up with a lat/lng across the river.

r-wei:
What do we need geocoding for? Addresses will be enough for the documenters, right?

eads:
@r-wei there was discussion of mapping, which requires geocoding to get lat/lngs. If it was just addresses, we could just skip this step. 

It should be pretty handy -- a Documenter using the central calendar could essentially tap the calendar item to get routing information if they're driving to an event.

But that's not really realistic if the coordinates take them several blocks away :/

I'm curious to see what Google says and, if they offer to let us use their data beyond what's laid out in their terms of service, to see if we want to take advantage of that.

r-wei:
Ah ok. Yes, I remember there was interest in mapping also to see where in the city most meetings were taking place, or to create other visualizations from this data (maybe during the hackathon).

If the central calendar is a google calendar, we could put the addresses into the location, and google's apps would take care of mapping.. just throwing out ideas.

eads:
The addresses-in-location-and-geocode-on-client-side has been running through my head as a potential option. Potentially has the benefit of using whatever mapping tool you like.

But you couldn't really do an interactive map with all events on the client side.

jim:
I was thinking about adding a cache to the geocoder (probably Redis) so we
would only ever lookup the coordinates for any address once. I wasn't sure
of the deployment plan, though, and didn't want to add additional external
dependencies until that was sorted out.

One possible need for having coordinates in the database is to enable a
geographic search for events within a certain distance of a location.

eads:
For the amount of data we need to cache, the file system might also be fine.

easherma:
Mapzen Search is also a great option for geocoding that I personally have extensive experience:

https://mapzen.com/products/search/

It integrates a couple different sources of address data as well as venues from OSM. With a focus on the Chicago/Illinois region, you could easily zone in on addresses as well as places (such as 'Chicago City Hall').

easherma:
@eads I'm going to try and set some time aside to look at this tomorrow and offer my two cents
Would you prefer comments here or in #95. Also I assume you'd prefer a seperate pull request from me if I have a contrib for geocoding, or should I try to comment on the code as is? I want to help effectively and not create a mess!

eads:
Ha, it's all a mess @easherma. Probably easiest just to talk about this in slack when you're available. I'm around all day.

easherma:
stubbing out some thoughts here as I'm getting more familiar with the codebase

- it does not appear that the scrapers all have consistent location of potential street address data. 
- in my opinion, a decent pipeline would be:

1. write out the scraped results as json or csv and make the sure the data is standardized
2. Prep a list of addresses and drop duplicates
3. geocode them, output the results and join/merge with scraped output create a final, appended output
4. we can do this all with flat files, could prototype with pandas and then write it out in more vanilla Python if we prefer.
5. we don't really need a relational DB IMO, especially at this stage. overall the data is small (looks like under 5 megs), and we're not planning on changing it much if at all once it is scrapped and structured. we could even parse them into markdown files and render a static site with little effort.

jim:
@eads @easherma @r-wei Were any decisions made regarding the direction to take here? I'm happy to  own this PR and get it ready for merging unless someone else is attached to it!

eads:
@jim hey! I got it in there, using MapZen's geocoder. Some cursory tests make me a little more confident in it the other non-Google geocoders. Of course, that could also be entirely wrong... 

In all events, it's pretty much exactly what you had already written, but with mapzen. I also removed the one test until we can figure out how to get it to run on travis.

I'd love to sync up with you on this and pass off additional tasks. See #143 and #145.

PR #86, opened 2017-08-30
Tinkering with tests

nickdevlin:


eads:
Am I understanding correctly this is a scraper for the public building commission with a failing test? (The failing test appears trivial).

Issue #77, opened 2017-08-25
create FAQ


eads:
finish #54 then move on to this

jim:
We have a [FAQ page](https://city-bureau.gitbooks.io/documenters-event-aggregator/faq.html), so I'm closing this.

Issue #49, opened 2017-08-15
Schedules in PDF format


jim:
Based on a discussion, we need to document the following process for developers in the documentation:

- If a data source is difficult to scrape (data is in a PDF, image, etc.), notate that on the sources spreadsheet and don't attempt to write a scraper for it.

Tasks to close this issue:

- [ ] Add documentation on how to handle difficult to scrape sources.

-----

*Original issue content*

Not sure if we want to get into trying to parse PDFs, but the [Chicago Housing Authority's meetings](http://www.thecha.org/about/board-of-commissioners/board-information-and-meetings/) are posted as an image and PDF.

Is it worth trying to automate the import of this information? Or is is better to just have a list of sources that a human needs to enter into the system once per year?

jim:
Looks like the Illinois Department of Corrections Advisory Board also posts most of their info inside PDFs. Here's the notice for their next meeting: https://www.illinois.gov/idoc/aboutus/advisoryboard/Pages/default.aspx

jim:
The [Cook County Commission on Human Rights](https://www.cookcountyil.gov/agency/commission-human-rights-0) also uses PDFs.

jim:
Same with [Housing Authority of Cook County](http://thehacc.org/about/): http://thehacc.org/wp-content/uploads/2017/08/2017-Scheduled-Board-Meetings.pdf

eads:
cc @diaholliday

diaholliday:
@jim we talked about this a moment ago but I wanted to post here as well in case anyone has the same question:

The Documenters program will eventually employ a Documenters Manager that will coordinate and assign meetings to Documenters—the manager will also be tasked with manually inputting meeting ties and locations that can't be scraped.

If you come across a PDF or other means of listing events that can't be scraped please note that and the form of documentation (PDF, etc) on the public meeting excel sheet. I'll eventually go through the list and create a list for the Documenters Manager to check on a recurring basis and add manually to the calendar.

jim:
@diaholliday Cool. I think the TODO here is to add that to the project docs, and I'll close this ticket when that has been done.

r-wei:
I added a column in the spreadsheet for this

PR #87, opened 2017-08-30
Add FAQ file.

jim:
This gives us a place to put frequently asked questions. Addresses #77 and completes #49.


eads:
I made some edits to the verbiage and should be ready to go with this.

PR #89, opened 2017-08-30
82 SPIDER Illinois Labor Relations Board

joshuarrrr:
This is a completed spider and test for the [Illinois Labor Relation Board events page](https://www.illinois.gov/ilrb/meetings/Pages/default.aspx). It's a simple page that just lists the next upcoming meeting for each of the three boards.

There's little structure on the page to work with, so the scraper may be a bit fragile.

Questions:

- How do we handle undefined end_times? Set to None or estimate a duration
- I could use some more guidance on classification. It seems like this page only lists confirmed meetings, but most of the other scrapers default to tentative
- Is there a taxonomy of meeting types? I set these to "board meeting", but I can change it to None

### DIFF
--- a/documenters_aggregator/spiders/ilrb.py
+++ b/documenters_aggregator/spiders/ilrb.py
@@ -1,0 +1,147 @@
+# -*- coding: utf-8 -*-
+"""
+All spiders should yield data shaped according to the Open Civic Data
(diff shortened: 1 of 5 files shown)


joshuarrrr:
Pulled from wrong branch

Issue #82, opened 2017-08-29
SPIDER: Illinois Labor Relations Board


joshuarrrr:

joshuarrrr:
Done via #90

PR #90, opened 2017-08-30
82 SPIDER Illinois Labor Relations Board

joshuarrrr:
This is a completed spider and test (#82) for the [Illinois Labor Relation Board events page](https://www.illinois.gov/ilrb/meetings/Pages/default.aspx).  It's a simple page that just lists the next upcoming meeting for each of the three boards.

There's little structure on the page to work with, so the scraper may be a bit fragile.

Questions:

- How do we handle undefined end_times? Set to None or estimate a duration
- I could use some more guidance on classification. It seems like this page only lists confirmed meetings, but most of the other scrapers default to tentative
- Is there a taxonomy of meeting types? I set these to "board meeting", but I can change it to None

### DIFF
--- a/documenters_aggregator/spiders/ilrb.py
+++ b/documenters_aggregator/spiders/ilrb.py
@@ -5,14 +5,14 @@ specification (http://docs.opencivicdata.org/en/latest/data/event.html).
 """
 import scrapy
 
-import re
-from datetime import datetime, timedelta
+from datetime import datetime
 from pytz import timezone
+from slugify import slugify
 
 
 class IlrbSpider(scrapy.Spider):
     name = 'ilrb'
-    allowed_domains = ['https://www.illinois.gov']
+    allowed_domains = ['www.illinois.gov']
     start_urls = ['https://www.illinois.gov/ilrb/meetings/Pages/default.aspx']
 
     """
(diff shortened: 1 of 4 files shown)


documenters_aggregator/spiders/ilrb.py
```diff
@@ -0,0 +1,147 @@
...
+                continue
+            name = self._parse_name(item)
+            yield {
+                '_type': 'event',
+                'id': self._generate_id(self._format_date(start_datetime), name),
+                'name': name,
+                'description': self._parse_description(item),
+                'classification': self._parse_classification(item),
+                'start_time': self._format_date(start_datetime),
+                'end_time': self._calc_end(start_datetime, hours=1),
```
eads: I'm going to leave this PR open until we discuss this evening (or here today), but I feel pretty strongly this should be `None` or some other indication that we don't know. We shouldn't speculate about such things and instead just should be accurate.

documenters_aggregator/spiders/ilrb.py
```diff
@@ -0,0 +1,147 @@
...
+import scrapy
+
+import re
+from datetime import datetime, timedelta
+from pytz import timezone
+
+
+class IlrbSpider(scrapy.Spider):
+    name = 'ilrb'
+    allowed_domains = ['https://www.illinois.gov']
```
eads: should be just `www.illinois.gov`

documenters_aggregator/spiders/ilrb.py
```diff
@@ -0,0 +1,147 @@
...
+            }
+
+    def _generate_id(self, start_time, name):
+        """
+        We use the start time to generate an ID since there is no publically
+        exposed meeting ID.
+        """
+
+        date = start_time.split('T')[0]
+        dashified = re.sub(r'[^a-z]+', '-', name.lower())
```
eads: Not a blocker, but it's possible that [python-slugify](https://github.com/un33k/python-slugify) is a little safer option.

eads:
One last thing: `./tests/test_ilrb.py:29:29: E711 comparison to None should be 'if cond is None:'`

I think that's template cruft and I'll just fix it myself :)

Thanks! One more down.

PR #91, opened 2017-08-30
Template updates

joshuarrrr:
I ran into some errors and failing tests that can be fixed in the templates.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -24,7 +24,7 @@ def genspider(ctx, name, domain, start_urls=None):
     Example:
     invoke genspider testspider http://www.citybureau.org -s=http://citybureau.org/articles,http://citybureau.org/staff
     """
-    spider_filename = _gen_spider(name, domain)
+    spider_filename = _gen_spider(name, domain, start_urls)
     print('Created {0}'.format(spider_filename))
 
     test_filename = _gen_tests(name, domain)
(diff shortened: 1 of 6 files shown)


r-wei:
Thanks for doing this! These bugs have been, well, bugging me. :P 

I'd like to change line 14 in templates/spider.tmpl to use the start_urls argument from genspider (if present), and the domain (without the leading http://) otherwise. And we should also change tests/test_tasks.py so that the tests pass. @joshuarrrr, do you want to take a crack at those? If not, would you be ok with me contributing to this branch?

eads:
+1 to fixing tests in test_tasks. I wonder how to test to make sure future tweaks to the templates don't break syntax tests etc.

joshuarrrr:
@r-wei I'll change line 14 and fix the tests.

joshuarrrr:
@r-wei I fixed the failing tests and added support for start_urls.

But could you take a look? I'm don't have that much experience using Jinja to template out other Python files, and I imagine there's probably a cleaner way to do the logic.

I also just used the initial start_urls string input; it might be cleaner to use the parsed list instead.

eads:
I'll review.

r-wei:
Thanks @eads. There's some inconsistency in the code on the expected format of start_urls. I was expecting each url to start with http, and thus using `,http` to split the urls (in case there were non-escaped commas). Joshua has been removing the http's (because it's cleaner for the template?). I don't have strong opinions on which way to go.

Issue #84, opened 2017-08-29
Spider: Cook County Electoral Board


r-wei:
http://www.cookcountyclerk.com/countyboard/boardmeetings/Pages/default.aspx

Appears not to be on legistar

PR #92, opened 2017-09-01
SPIDER: cook county elections board - working spider

r-wei:
resolves 84

### DIFF
--- a/documenters_aggregator/spiders/cceb.py
+++ b/documenters_aggregator/spiders/cceb.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+"""
+All spiders should yield data shaped according to the Open Civic Data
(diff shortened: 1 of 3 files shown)


documenters_aggregator/spiders/cceb.py
```diff
@@ -0,0 +1,126 @@
...
+"""
+import scrapy
+
+from datetime import datetime
+from pytz import timezone
+
+
+class CcebSpider(scrapy.Spider):
+    name = 'cceb'
+    allowed_domains = ['http://cookcountyclerk.com/elections/electoralboard/Pages/default.aspx']
```
eads: this should just be `cookcountyclerk.com`...

perhaps we should clarify this point in the docs / inline comments in the genspider templates?

r-wei:
closed #84

PR #100, opened 2017-09-07
Add files via upload

shinda:
cookcountyil.gov/calendar spider has a few issues that I might need assistance with...

1.The spider seems to only scrape data from events for a single month.

2. Although there isn't much css / html markup in the output generated by the spider, there is
    enough of it to be noticeable thus making it a challenge to read the results.

3. There are events on the site for Board, Committee, and Sub-Committee meetings that do not
     have end times and I'm not sure about how to handle those since they could either last for a few 
     minutes or a couple of hours. 

4. I have commented out the make_date statements since I'm not yet completely confident in how 
     it's supposed to work and also because the spider scraped some event times from the HTML / 
     CSS formating. 

I'd love some feedback on these and any other issues!

### DIFF
--- a/cookcountyil.py
+++ b/cookcountyil.py
@@ -1,0 +1,177 @@
+## -*- coding: utf-8 -*-
+
+"""
+All spiders should yield data shaped according to the Open Civic Data
+specification (http://docs.opencivicdata.org/en/latest/data/event.html).
+"""
+
+import scrapy
+import re
+import urllib3
+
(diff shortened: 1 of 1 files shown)


jim:
This data will be pulled from the Cook County feed. Thanks for the submission!

shinda:
Hi Jim, I just now realized that I could have allowed edits to the code.  Is more work needed for it?

PR #105, opened 2017-09-13
Joshuarrrr's template updates

eads:
🚨 @r-wei @joshuarrrr 🚨 -- bunch of changes here. This could supersede #91 if yall on board with the changes. I opened a new PR because I was struggling to push to the one @joshuarrrr opened and I'm a caveman and didn't want to figure out what I was doing wrong.

The big change here is that the `invoke genspider` signature has changed. I'm proposing it be:

```
invoke genspider <NAME> <START_URLS>
```


eads:
The under the hood stuff is a tweak to the domain logic (domains are now is implicitly generated from start_url list) and to the templates, which now use the jinja2 `{{ var|join(DELIM) }}` filter to spit out the domain and url lists. Plus the inevitable changes to tests.

Issue #94, opened 2017-09-06
PROPOSAL: descriptive event name as spider object property


eads:
It would be nice to capture the official names of events. The acronyms are lovely for the internal spider names but we'll need and want the official name down the road, and this feels like a useful place to stash it.

eads:
Closed in #106

PR #106, opened 2017-09-13
add `long_name` property to all spiders

eads:
the acronyms are lovely and yet confusing! refs #94 

there's some stuff to figure out here, like how genspider should work with a long name, which also makes me think this should wait for #105


eads:
So I'm thinking one of two options for `invoke genspider`:

Required option:

```
invoke genspider cpb "Chicago Police Board" http://...
```

Optional option:

```
invoke genspider cpd http:// -t "Chicago Police Board"
```

where the long name is optional and replaced by "cpb" or maybe "CPB" when generated.

What do yall think?

joshuarrrr:
I would lean toward the required option.

eads:
Unless someone objects, I think @joshuarrrr's instinct is good and we should just require it.

It isn't something you have to do a lot, and capturing that information up front is good to get out of the way at that point.

r-wei:
Sounds good to me

PR #108, opened 2017-09-18
107 spider cit

dcldmartin:
This one just had dates - no meeting times - and no additional info. Not sure how useful it is....


eads:
Looks like there's a bunch of imports that need to be removed before tests pass. 

@diaholliday what do you think we should do about the missing times?

diaholliday:
Can we have them all say 10a? Looks like the Land Bank always meets at the same time. You can see the times and minutes when you click further into the calendar notice.

dcldmartin:
I think you're looking at the wrong one :) The infrastructure trust doesn't
have any times on their meetings as far as I can see on the site:

http://chicagoinfrastructure.org/public-records/scheduled-meetings/

diaholliday:
Ah yes, you are correct. I was looking at the Land Bank. But it looks like the Trust does also meet (mostly) at 10a, though it's not consistent from meeting to meeting: http://chicagoinfrastructure.org/press/blog/

dcldmartin:
Interesting.... And scraping the blog might be a little tough. It's a
little less consistently structured. I could just point to the blog page in
an additional field to check for time/location.

Though, if the date on the blog post is to be believed, the notice for
rescheduling the last meeting was given after the meeting happened. And
notice was given of the previous meeting only a couple days in advance. If
the point of this info is to plan ahead this may not be the best source.

r-wei:
Something strange is going on with this pull request. The travis ci check is failing because of a failed test in the cpb spider, but the cpb.py and test_cpb.py are not in the 107-spider-cit branch... @eads , can you rebuild travis?

eads:
Man, being sick really caused me to miss a lot of stuff. This all looks good.

PR #112, opened 2017-09-20
add dockerfile

easherma:
should be a simple add, made a simple dockerfile, ran tests on python 3.4, 3.5, 3.6


eads:
Woot! Also the first step to deploying on Amazon Elastic Cluster!

eads:
Er EC2 Container Services. So many acronyms and names.

PR #114, opened 2017-09-20
WIP ccc spider

adebruine:
Having trouble parsing the list of links to event pages/updating tests accordingly.


documenters_aggregator/spiders/ccc.py
```diff
@@ -0,0 +1,139 @@
...
+import re
+
+from datetime import datetime
+from pytz import timezone
+
+
+class CccSpider(scrapy.Spider):
+    name = 'ccc'
+    allowed_domains = ['http://www.ccc.edu/departments/Pages/Board-of-Trustees.aspx']
+    start_urls = ['http://http://www.ccc.edu/departments/Pages/Board-of-Trustees.aspx']
```
r-wei: an extra http:// here

r-wei: Maybe try using this site for start_urls instead? http://www.ccc.edu/events/Pages/default.aspx?dept=Board%20of%20Trustees

documenters_aggregator/spiders/ccc.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        for link in response.css('a.Eventslink::attr(href)').extract():
+            next_url = "http://www.ccc.edu" + link
+            return scrapy.Request(next_url, callback=self.parse_event_page())
+
+    def parse_event_page(self, response):
+            yield {
+                '_type': 'event',
+                'id': self._parse_id(),
+                'name': self._parse_name(),
```
r-wei: Try `self._parse_name(response)`. The _parse_name function probably needs the response to know what to parse.

documenters_aggregator/spiders/ccc.py
```diff
@@ -0,0 +1,139 @@
...
+            },
+        }
+
+    def _parse_all_day(self, item):
+        """
+        Parse or generate all-day status. Defaults to false.
+        """
+        return False
+
+    def _parse_name(self, item):
```
r-wei: This is fine, but for consistency/ease of understanding, it might be better to do `def _parse_name(self, response)`. Same for _parse_description, _parse_start, etc.

tests/test_ccc.py
```diff
@@ -0,0 +1,68 @@
...
+#def test_tests():
+#    print('Please write some tests for this spider or at least disable this one.')
+#    assert False
+
+
+"""
+Uncomment below
+"""
+
+test_response = file_response('files/ccc.html')
```
r-wei: What file_response does is take a file (ccc.html) that we've saved locally and create a Response object from it, so that we can test out the spider's parse method without needing an Internet connection. (The website might change, but we don't want that to break our tests.) Since your parse method goes to a different website for each event, we can just test the _parse_event_page function on a single event's site. You can do something like this:
- Save one of the event pages in tests/files/  (for example, save this site: http://www.ccc.edu/events/Pages/Committee-on-Finance-and-Administrative-Services-Sept.-2017.aspx as tests/files/ccc_event.html)
- Code: 
```python
test_response = file_response('files/ccc_event.html')
spider = CccSpider()
item = spider._parse_event_page(test_response)
# write your tests as usual using item instead of parsed_items[0]
```

r-wei:
Hey @adebruine , I just gave a bunch of suggestions and comments. Feel free to post here if you're having trouble with a specific line of code or error messages. :)

Issue #110, opened 2017-09-19
Fix spiders not conforming to events schema


r-wei:
Since we've been figuring out the [events schema](https://github.com/City-Bureau/documenters-aggregator/blob/master/docs/event_schema.md) while spiders are being made, some of the earlier spiders don't conform to the events schema. At the hackathon, I plan to go through a bunch of my old spiders and bring them up to date to the events schema.

joshuarrrr:
Sounds good, I was thinking about this too.

r-wei:
validation pipeline is done / a first version of the validation pipeline has been merged in

Issue #75, opened 2017-08-23
Event format validator


jim:
I think it would be nice to have an event validator as a part of the pipeline in order to ensure that any data that gets downstream has all of the required fields in the correct formats.

r-wei:
Line 76 in pull request #70 might be the airtable-python place for it..?

r-wei:
deployed

PR #123, opened 2017-10-01
validation pipeline

r-wei:
related to #75 #110 

Here's a basic validation pipeline to check if a batch of items (e.g., all the items scraped from one spider) conform to the events schema. Example usage is at the bottom of ValidationPipeline.py. 

If events don't conform to the schema, we log the id and the non-conforming field in a logfile (examples: `tests/files/log_cchhs.txt` and `tests/files/log_cchhs_location.txt`). I'm thinking that a person could use the validation pipeline to go through spiders and fix them, so the logs are written in a way to help you easily see patterns of errors.

This checks for: required fields, required values (like _type must be equal to 'event'), required types, string formatting for datetimes.

This doesn't yet check for logical errors like:
- start time <= end time
- status = passed if an only if start time > now
- ids are unique


jim:
This is great! I was thinking about building something like this myself. I think this is an important piece of functionality to help ensure that the data we push downstream satisfies certain requirements.

I'd like to propose that we create a `log` directory (or perhaps a `tmp/log` directory in the root of the project and write any log files there. This will make them easy to add to `.gitignore` and keep them out of git. What do you think?

From a code standpoint, if we change the `check_required_fields`, etc. methods to validate a single item (instead of a batch), it would move all of the iteration up into the `validate_batch` method. I think this would make the validation methods themselves a bit simpler and easier to follow.

r-wei:
@jim sounds great! we can work on this tomorrow. :)

r-wei:
Note to self / to do: 

- log messages instead of printing them
- rename the noncompliance attributes to something more understandable
- include sources field in validation

Issue #146, opened 2017-10-14
SPIDER: Animal Care and Control Commission (Advisory Board)


palewire:

palewire:
Closed by #153

PR #153, opened 2017-10-14
SPIDER: Animal Care and Control Commission

palewire:
For #146

### DIFF
--- a/documenters_aggregator/spiders/accc.py
+++ b/documenters_aggregator/spiders/accc.py
@@ -1,0 +1,120 @@
+# -*- coding: utf-8 -*-
+"""
+All spiders should yield data shaped according to the Open Civic Data
+specification (http://docs.opencivicdata.org/en/latest/data/event.html).
+"""
+import scrapy
+from datetime import datetime
+from dateutil.parser import parse as dateparse
+
+
+class AcccSpider(scrapy.Spider):
+    name = 'accc'
+    long_name = 'Animal Care and Control Commission'
+    allowed_domains = ['www.cityofchicago.org']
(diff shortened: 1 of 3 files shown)


documenters_aggregator/spiders/accc.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+"""
+All spiders should yield data shaped according to the Open Civic Data
+specification (http://docs.opencivicdata.org/en/latest/data/event.html).
+"""
+import scrapy
+from datetime import datetime
```
eads: Pyflakes says this is imported but unused.

tests/test_accc.py
```diff
@@ -0,0 +1,21 @@
+import pytest
```
eads: Imported by unused, per pyflakes.

documenters_aggregator/spiders/accc.py
```diff
@@ -0,0 +1,120 @@
...
+                'all_day': self._parse_all_day(text),
+                'status': self._parse_status(text),
+                'location': self._parse_location(text),
+            }
+
+    def _parse_id(self, item):
+        """
+        Calulate ID. ID must be unique within the data source being scraped.
+        """
+        return None
```
eads: We need an ID for the event. It just has to be unique within the universe of data the spider is collecting, so you could just slugify name and date/time as long as they never have more than one meeting at the same time. Or see if there's an ID in the DOM.

palewire:
I think the changes you've requested have been pushed.

eads:
kitten cop

PR #156, opened 2017-10-15
WIP 0061 spider cityclerk

bonfirefan:
Until [palewire:patch-2](https://github.com/City-Bureau/documenters-aggregator/pull/148#issuecomment-336669782) gets fully merged, I won't be able to test this, but placing it here so it can at least get feedback.


tests/test_ccc.py
```diff
@@ -0,0 +1,67 @@
...
+from tests.utils import file_response
+from documenters_aggregator.spiders.ccc import CccSpider
+
+def test_tests():
+    print('Please write some tests for this spider or at least disable this one.')
+    assert False
+
+
+"""
+Uncomment below
```
eads: You'll need to merge latest master which the blockers from yesterday and then write a few tests for this.

documenters_aggregator/spiders/ccc.py
```diff
@@ -0,0 +1,69 @@
...
+        return scrapy.Request(next_url, callback=self.parse, dont_filter=True)
+
+    def _parse_location(self, item):
+        """
+        Parse or generate location. Url, latitutde and longitude are all
+        optional and may be more trouble than they're worth to collect.
+        """
+
+        e_pg = requests.get(ocd_url + item['id'])
+        return e_pg.content['location']
```
r-wei: We'll geocode the latitude and longitude, so you don't have to worry about it.

I got an error on lines 68-9 running ```scrapy crawl ccc```, and I think this fixes it:
```
e_pg = requests.get(self.ocd_url + item['id'])  # ocd_url isn't a global variable, but you can access it with self
e_dict = json.loads(e_pg.text)  # transform the response object into a dictionary
return e_dict['location']
```

bonfirefan:
Okay, this was likely annoying with the multiple commits, but it looks like invoke runtests completely fails for me, as in it gets hung up on the first call to _getconftestmodules and sent me down a gdal installation rabbithole.

eads:
Wild. I can check the tests on my machine, but perhaps we should chat about the nature of these issues -- I can't tell if I made a mistake in my rush to get deployment and geocoding nailed down or if there's something more fundamental.

bonfirefan:
No, the issue was on my end - I was using a conda virtualenvironment, and I suspected that something may not have been complete with package installations. Using virtualenv solves the problem, so I'm able to test now.

bonfirefan:
It looks like an issue with another spider is causing tests to fail. Unable to figure out if there is an issue on my end, otherwise.

bonfirefan:
Hey all, I think this might be okay to merge, but like I said, the tests are failing on another issue. Otherwise I'm not sure what else to change here.

PR #148, opened 2017-10-14
Add pipelines import to fix scrapy bug

palewire:
The scrapy command is currently failing because the settings are failing to import the logging pipeline from a nested file. I've corrected that problem by introducing an import to the file's parent module that will line up with the path provided in the settings.

Here's the bug I saw that sent me down this route

```python
$ scrapy crawl accc
no pupa_settings on path, using defaults
2017-10-14 13:47:18 [scrapy.utils.log] INFO: Scrapy 1.4.0 started (bot: documenters_aggregator)
2017-10-14 13:47:18 [scrapy.utils.log] INFO: Overridden settings: {'NEWSPIDER_MODULE': 'documenters_aggregator.spiders', 'CLOSESPIDER_ERRORCOUNT': 5, 'COOKIES_ENABLED': False, 'COMMANDS_MODULE': 'documenters_aggregator.commands', 'USER_AGENT': 'Documenters Aggregator [development mode]. Learn more and say hello at https://city-bureau.gitbooks.io/documenters-event-aggregator/', 'BOT_NAME': 'documenters_aggregator', 'SPIDER_MODULES': ['documenters_aggregator.spiders']}
2017-10-14 13:47:18 [scrapy.middleware] INFO: Enabled extensions:
['scrapy.extensions.corestats.CoreStats',
 'scrapy.extensions.telnet.TelnetConsole',
 'scrapy.extensions.memusage.MemoryUsage',
 'scrapy.extensions.logstats.LogStats']
2017-10-14 13:47:18 [scrapy.middleware] INFO: Enabled downloader middlewares:
['scrapy.downloadermiddlewares.httpauth.HttpAuthMiddleware',
 'scrapy.downloadermiddlewares.downloadtimeout.DownloadTimeoutMiddleware',
 'scrapy.downloadermiddlewares.defaultheaders.DefaultHeadersMiddleware',
 'scrapy.downloadermiddlewares.useragent.UserAgentMiddleware',
 'documenters_aggregator.middlewares.DocumentersAggregatorRobotsTxtMiddleware',
 'scrapy.downloadermiddlewares.retry.RetryMiddleware',
 'scrapy.downloadermiddlewares.redirect.MetaRefreshMiddleware',
 'scrapy.downloadermiddlewares.httpcompression.HttpCompressionMiddleware',
 'scrapy.downloadermiddlewares.redirect.RedirectMiddleware',
 'scrapy.downloadermiddlewares.httpproxy.HttpProxyMiddleware',
[...]

PR #157, opened 2017-10-15
Spider: PBCC

### LINKED ISSUE #71: SPIDER: Public Building Commission of Chicago

pjsier:
Closes #71. This is ready for review, but I wanted to flag a couple things I wasn't sure about because  they seemed different than most of the other scrapers.

* I used the few PBCC event types for `classification`, but I could also see those getting moved to `name` alone
* The location information is all over the place in terms of structure and text content, so I left it off since it has the URL included
* I used the `all_day` field events at Daley Plaza without a start time, but not sure if those should just be listed as starting at midnight
* Most events had no description, but the advertisements for bids had full pages ([one example](http://www.pbcchicago.com/content/working/opening_display.asp?BID_ID=502)). Instead of trying to pick which part of the page to include I left `description` blank across the board
* Locally this is dependent on the changes in PR #148, but that could just be my environment

### DIFF
--- a/documenters_aggregator/spiders/pbcc.py
+++ b/documenters_aggregator/spiders/pbcc.py
@@ -115,18 +115,29 @@ class PbccSpider(scrapy.Spider):
         Parse or generate location. Url, latitutde and longitude are all
         optional and may be more trouble than they're worth to collect.
 
-        TODO: Location details are inconsistent across event types, addresses
-        are mixed up with names and neither are reliably in the same place.
-        Could make manual exceptions for board meetings, or go to sub-page
+        All board meetings and special board meetings are required to be
+        held at the same location, so hard-coding that here with Mapzen
+        Search coordinates included. Otherwise location information is too
+        inconsistent for a reliable format.
         """
-        return {
-            'url': None,
-            'name': None,
-            'coordinates': {
-                'latitude': None,
-                'longitude': None,
-            },
-        }
+        if 'Board Meeting' in self._parse_classification(item):
+            return {
+                'url': 'https://thedaleycenter.com',
+                'name': 'Second Floor Board Room, Richard J. Daley Center, 50 W. Washington Street',
+                'coordinates': {
+                    'latitude': 41.884089,
+                    'longitude': -87.630191,
+                },
+            }
+        else:
+            return {
+                'url': None,
+                'name': None,
+                'coordinates': {
+                    'latitude': None,
+                    'longitude': None,
+                },
+            }
 
     def _parse_all_day(self, item):
         """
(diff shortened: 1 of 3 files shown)


documenters_aggregator/spiders/pbcc.py
```diff
@@ -0,0 +1,198 @@
...
+        event_link = item.css('a::attr(href)').extract_first()
+        if '?eID=' in event_link:
+            id_suffix = event_link.split('?eID=')[-1]
+        elif '?BID_ID=' in event_link:
+            id_suffix = event_link.split('?BID_ID=')[-1]
+        else:
+            raise ValueError('ID is required and not present in eID or BID_ID params')
+        return 'PBCC{}'.format(id_suffix)
+
+    def _parse_classification(self, item):
```
eads: I like using the PBCC event types for the classification. Agency's with a reasonable classification scheme mean we can filter easily on a combination of agency name and classification! 👍

documenters_aggregator/spiders/pbcc.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def _parse_description(self, item):
+        """
+        Parse or generate event name.
+
+        Other than the page-long info on AFBs, there isn't much of a description for
+        events outside of the classification and name, so returning None. Could be
+        implemented later for AFBs
+        """
+        return None
```
eads: If there is a [page like this](http://www.pbcchicago.com/content/working/opening_display.asp?BID_ID=502) it seems like it could be nice to stash in description.

documenters_aggregator/spiders/pbcc.py
```diff
@@ -0,0 +1,198 @@
...
+        return 'tentative'
+
+    def _parse_location(self, item):
+        """
+        Parse or generate location. Url, latitutde and longitude are all
+        optional and may be more trouble than they're worth to collect.
+
+        TODO: Location details are inconsistent across event types, addresses
+        are mixed up with names and neither are reliably in the same place.
+        Could make manual exceptions for board meetings, or go to sub-page
```
eads: I'd like to see an exception for board meetings. If there's a way to capture at least an address for board meetings, that would be good.

pjsier:
Added those changes in the last commit. The board meeting location is usually listed at the detail page, but according to [one of the earlier events](http://www.pbcchicago.com/content/about/calendar_detail.asp?tID=2&eID=51) it's required to be at that location in their by-laws so it seems like a safe bet to hard-code it

eads:
This does get at an interesting issue -- the hardcoded location feels like the sort of thing someone should check on a regular basis in case the by-laws _do_ change. That's perhaps beyond the scope of what this slice of the larger project can do, but it's still a significant issue. I think it's entirely reasonable to hard code and then figure it out though :)

PR #154, opened 2017-10-14
Fixed typo in spider template

palewire:

PR #158, opened 2017-10-15
Fix test template failures from #154

pjsier:
Looks like all Travis tests are failing because the typo fixed in #154 was still in the test template. This should fix it


eads:
Excellent, thanks for this. I was sitting down to do this and here this is :)

PR #163, opened 2017-10-18
Use Flake8

bergren2:
Was seeing an issue with our linters parsing a `src` directory that I think arose from these changes: 1fd9c7a. This PR is to add [Flake8](http://flake8.pycqa.org/), which has PEP 8 and pyflakes linters built in, while also allowing us to ignore directories we don't need to lint.


eads:
Good business!

PR #172, opened 2017-11-07
Fix PBCC tests

pjsier:
Fixes the error with the PBCC scraper tests pointed out in #171. The starting date is set in the scraper with `datetime.now()`, which works when it's run normally, but caused an error for the sample data generated in October because it tried to create November 31 based on the HTML file. Manually setting it to that time now so the tests pass


jim:
Thanks for fixing this!

PR #176, opened 2017-11-10
Add CTA board spider

pjsier:
Adding a spider for the Chicago Transit Authority board. It's mostly straightforward, but there's one part I have a question on.

On the [events page](http://www.transitchicago.com/agendaminute/default.aspx?F_OrderBy=MeetingDate+DESC), one type of meeting's start time is consistently marked "Immediately following Cmt Strategic Planning & Svs". I'm just setting the date on those and leaving the time at midnight, but would it be better to add a time an hour or so after the previous meeting?

### DIFF
--- a/documenters_aggregator/spiders/chi_transit.py
+++ b/documenters_aggregator/spiders/chi_transit.py
@@ -1,0 +1,144 @@
+# -*- coding: utf-8 -*-
+"""
+All spiders should yield data shaped according to the Open Civic Data
+specification (http://docs.opencivicdata.org/en/latest/data/event.html).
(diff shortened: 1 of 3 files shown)


r-wei:
I'd say put the same time for the finance+audit meeting as the strategic planning one. It looks like the same room is used for other meetings right after, so these two probably go together.

pjsier:
@r-wei works for me! Just updated that, and also changed the start URL so that it shows all events by default

jim:
@pjsier This looks good to me! I'll get this merged in shortly.

PR #182, opened 2017-11-17
Standardize ids

jim:
This pull request changes the way we generate events ids to a new scheme that is more human- and machine-readable. (We should always treat ids as opaque, and I thought about hashing the result to enforce this, but that seemed overbearing at this point).

I want to get this locked down so that we can start using the system. We can easily make a change to any of the other values for an event later on, but getting the identifiers right needs to be done now as changing their format later will be extremely painful.

Most of the spiders are using their own heuristic to come up with an event identifier (usually using the date and sometimes the event name), as there is often not a publicly-visible ID for each event.

This set of patches will (when complete) update all spiders to use IDs in the following format:

`<spider-name>/<YYYYMMddhhmm>/<id-or-x>/<underscored-event-name>`

This will give us IDs such as:

`chi_buildings/201710161230/2176/daley_plaza_italian_exhibit`

when the event has a useful id exposed, and like this when it does not:

`chi_buildings/201710161230/x/daley_plaza_italian_exhibit`

This scheme achieves identifiers that are just as unique as the prior scheme, but with a regular format and only using a-z, 0-9, /, and _.

### DIFF
--- a/deploy/prod_settings.py.example
+++ b/deploy/prod_settings.py.example
@@ -30,12 +30,12 @@ COOKIES_ENABLED = False
 # One of:
 # * documenters_aggregator.pipelines.DocumentersAggregatorLoggingPipeline,
 # * documenters_aggregator.pipelines.DocumentersAggregatorSQLAlchemyPipeline,
-# * documenters_aggregator.pipelines.DocumentersAggregatorAirtablePipeline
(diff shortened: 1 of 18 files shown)


r-wei:
This looks good to me. I can also run it on my laptop tonight to double check.

Issue #180, opened 2017-11-13
Add community area to events


diaholliday:
Could set up a way to list the community area for each meeting in Airtable? I currently have an Airtable column for it but assume it's connected to the geocoding plan for the meetings.

eads:
Whoops, didn't mean to close. Was just gonna say getting political geographies _might_ be an argument for the Texas A&M geocoder.

easherma:
Mapzen has community areas. Other political boundaries I can't remember off the top of my head, but once we get lat/lng we can make another call to other APIs that would give us boundaries we'd like.

jim:
We might want to have `state`, `city`, `area`, etc. fields within the `location` since we probably get most of that from the initial geocode. I'm not sure how many levels we should support, but I think looking in to the [MapZen Places API](https://mapzen.com/data/places/) is a good place to start.

diaholliday:
@r-wei can we add this to the kanban board also?

r-wei:
deployed yesterday using mapzen

Issue #173, opened 2017-11-10
Geocoder pipeline needs to cache results


jim:
This is if we have trouble hitting the MapZen API rate limit again.

r-wei:
deployed yesterday

PR #194, opened 2017-11-27
WIP: cache and use geocoder results

r-wei:
I created a table in airtable called 'GeocodeDatabase' to cache the lat/longs from mapzen. The geocoder pipeline first checks that table. Only if it doesn't find anything, it makes a mapzen request and records it in GeocodeDatabase. 

Done writing, but haven't tested it yet.

addresses #173


r-wei:
resolves #180

r-wei:
may resolve #197

PR #200, opened 2017-12-13
WIP: include a validation task when travis runs

r-wei:
@eads @jim, I'd like to include a validation task on Travis that does the following:
- checks if there are any new or modified spiders in the pull request
- runs those spiders, runs the scraped items through a GithubValidationPipeline, and saves them
- if <90% of the newly scraped items from a spider passes validation, then throw an error on Travis

Does this seem like a good idea? I think it would be nice to have some validation show up on Travis when new pull requests are made. Then again, running the spiders during CI adds some uncertainty...

Here is some code that almost works. Some problems:
- Calling `scrapy crawl` inside of an invoke task throws an import error: no module called documenters_aggregator
- `invoke -l` now gives a bunch of scrapy output...

### DIFF
--- a/documenters_aggregator/pipelines/__init__.py
+++ b/documenters_aggregator/pipelines/__init__.py
@@ -3,6 +3,7 @@ from .geocoder import GeocoderPipeline
 from .LoggingPipeline import DocumentersAggregatorLoggingPipeline
 from .sqlalchemy import DocumentersAggregatorSQLAlchemyPipeline
 from .ValidationPipeline import ValidationPipeline
+from .github import GithubValidationPipeline
 
 
 __all__ = (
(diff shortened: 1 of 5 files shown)


pjsier:
Ahh didn't realize `ls` throws an error if nothing matches the pattern. Changing the last line to this should fix it:

```bash
if [ -e scripts/*.json ]; then ls scripts/*.json | xargs -I{} sh -c 'invoke validate-spider $(basename "${1%%.*}")' -- {}; fi
```

r-wei:
Things left to do:
- remove pandas dependency (and numpy, etc)
- add tests for TravisPipeline

r-wei:
Hey @pjsier, I merged your tests into this PR and added a few more. If things look good to you, we can merge this in to master.

pjsier:
@r-wei looks good to me! Not sure what the `ecs` error is about, but I'm guessing it has to do with an older version we merged in

Issue #215, opened 2018-01-06
Formatting issue with start/end times


diaholliday:
I've been playing around with this a bit but can't seem to figure out where the real problem lies. There's something going on between the scraped start/end times and the formula start/end times. It's possible to manually input start/end times correctly but the formula reads them wrong in the conversion
![screen shot 2018-01-06 at 11 34 14 am](https://user-images.githubusercontent.com/4017450/34642246-8eee5d5a-f2d5-11e7-83cb-30851c507c0e.png)

Long story short, the formula throws off meetings that are input manually and while the scraped meetings do seem (mostly) correct, the formula conversion renders them less useful if new meetings can't be correctly added by hand. It's generally difficult to know what the correct time is if they aren't all standardized through the same system.

Not sure what the answer here is but it;s very important that the times are correct and reliable (for obvious reasons)

diaholliday:
I went ahead and added 6 hours to the start time on my manual inputs to get the correct formula time and here's what I have now
![screen shot 2018-01-06 at 11 54 14 am](https://user-images.githubusercontent.com/4017450/34642428-7719ea34-f2d8-11e7-8aa3-3ac17d2c0471.png)

It looks mostly correct except some of the formula times are still an hour off. Does the timezone formula conversion account for daylight savings time?

r-wei:
resolved by #223

diaholliday:
@jim @r-wei just wanted to check in on this closed issue. I see a "meetings test" tab in Airtable that seems to have correct start times and start time formula columns, but I don't think this has been fixed in the main meetings tab (currently we have to add 6 hours in the start time columns for it to match the Start time formula columns that are visible to Documenters on the assignment form). Can we confirm the times now match and can be applied to the main meetings tab? 

One of the side issues here is the ability to properly use the Manual Input Form for meetings without having to add 6 hours to the formula: https://airtable.com/shrAUMvvmRKcOEYmu 

Meetings test view:
![screen shot 2018-03-18 at 8 44 57 pm](https://user-images.githubusercontent.com/4017450/37574301-e65993ac-2aed-11e8-954e-468b2752210b.png)

Main Meetings view: 
![screen shot 2018-03-18 at 8 50 19 pm](https://user-images.githubusercontent.com/4017450/37574321-027f033c-2aee-11e8-8357-2e0d3c2473d8.png)

Issue #202, opened 2017-12-20
chi_infra spider is broken


r-wei:
Here is the error from ```scrapy crawl chi_infra```

```
[scrapy.core.scraper] ERROR: Spider error processing <GET http://chicagoinfrastructure.org/public-records/scheduled-meetings/> (referer: None)
Traceback (most recent call last):
  File "/Users/aweia/.virtualenvs/documenters-aggregator/lib/python3.6/site-packages/scrapy/utils/defer.py", line 102, in iter_errback
    yield next(it)
  File "/Users/aweia/.virtualenvs/documenters-aggregator/lib/python3.6/site-packages/scrapy/spidermiddlewares/offsite.py", line 29, in process_spider_output
    for x in result:
  File "/Users/aweia/.virtualenvs/documenters-aggregator/lib/python3.6/site-packages/scrapy/spidermiddlewares/referer.py", line 339, in <genexpr>
    return (_set_referer(r) for r in result or ())
  File "/Users/aweia/.virtualenvs/documenters-aggregator/lib/python3.6/site-packages/scrapy/spidermiddlewares/urllength.py", line 37, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "/Users/aweia/.virtualenvs/documenters-aggregator/lib/python3.6/site-packages/scrapy/spidermiddlewares/depth.py", line 58, in <genexpr>
    return (r for r in result or () if _filter(r))
  File "/Users/aweia/projects/documenters-aggregator/documenters_aggregator/spiders/chi_infra.py", line 30, in parse
    self.year = int(year_match.group(1))
AttributeError: 'NoneType' object has no attribute 'group'
```

Working on this issue would familiarize you with our spider infrastructure and working with scrapy.

r-wei:
resolved by #223

Issue #221, opened 2018-01-11
City Council Spider Times are Incorrect


diaholliday:
Not sure if this applies other spiders or not but the City Council start times are off by several hours. For example, I know the committee on education starts at 2:30p tomorrow, not 8:30p. Is there a way to correct this and, second priority, any others that may be off?
![screen shot 2018-01-11 at 1 13 48 pm](https://user-images.githubusercontent.com/4017450/34842583-8c1f8b6c-f6d1-11e7-8073-9bd863c5c500.png)

easherma:
Sounds like it's probably a UTC time error given the hour interval. Should be an easy fix, I might be able to address this today and fix it on the scraper end. As for the data already in Airtable, I don't have access at the moment so I can't do much there.

diaholliday:
I'll defer to @r-wei on the push to Airtable, she has access (the accounts cost $$ so I'm trying to keep the number of accounts down until we have a better sense of who needs access and/or if a limited number of people can just push all necessary changes there)

easherma:
The spider appears to be outputting start times as I'd expect ( i.e. 2018-01-11T16:00:00-06:00 is 10:00 AM in Chicago which is right for https://chicago.legistar.com/MeetingDetail.aspx?ID=584299&GUID=C817555E-7ABB-4FAF-B79E-5CD1C4630278&Options=info&Search= ), so the issue may be in the way it's sent to Airtable, or the way Airtable is interpreting it. 

I'd wager the last one is probably the problem.

easherma:
It sounds like Airtable itself may be defaulting to a timezone, maybe there
is a project or account setting with a timezone selected?

r-wei:
I can fix this. I think 2018-01-11T16:00:00-06:00 is 4pm Chicago time, so it's the spider not Airtable. The start times for the ward nights are showing up correctly in Airtable.

easherma:
Isn't 16:00 - 6 10 AM? I checked the spider output locally but this seemed
to line up to me.

r-wei:
```
>>> import datetime
>>> import pytz
>>> datetime.datetime.now()
datetime.datetime(2018, 1, 12, 16, 49, 48, 404872)
>>> chicago_tz = pytz.timezone('America/Chicago')
>>> chicago_tz.localize(datetime.datetime.now()).isoformat()
'2018-01-12T16:50:46.670077-06:00'
```

[...]

PR #223, opened 2018-01-12
0221 fix citycouncil times

r-wei:
Resolving issue #0221


r-wei:
resolves #202. should resolve #215

PR #228, opened 2018-01-25
Initial stab at geocoding process, separating some concerns

easherma:
I haven't been able to follow up much on this since I took the first stab. 

I was working towards separating some of the process into individual Pipelines, would would work in this order:

* clean/standardize location data
* check known locations
* geocode blank/unknown locations
* validate and add known locations


jim:
See also #249.

easherma:
http://geocoder.readthedocs.io/ docs for  the library fyi

easherma:
To avoid confusion, I closed this since the other PR was merged.

PR #245, opened 2018-02-13
Prototype using datetime objects instead of ISO8601 formatted strings.

jim:
I've tested this against Airtable, and it correctly pushes dates and displays them in the correct time in the UI without needing the "force UTC" setting:

screen shot 2018-02-12 at 6 51 52 pm

- [ ] Update template to include calls to `_naive_datetime_to_tz` *including* the `_source_tz` argument:

    ```python
    self._naive_datetime_to_tz(date_time, "America/Chicago")
    ```


bonfirefan:
Seconded, this seems to make everything else much simpler. I'll add a print friendly date formatter to the local exporter. 
.

PR #247, opened 2018-02-13
Cleanup deploy dir

jim:
I had to adjust a setting in Codebuild to make these changes, so deploys will not work until this is merged 😄 

It should apply cleanly; I am opening this for visibility into the changes I am making.


jim:
This will resolve #241.

r-wei:
Even `deploy/prod_settings.py` is no longer needed on AWS?

jim:
@r-wei We do still use `deploy/prod_settings.py` in production, so I only deleted the `prod_settings.py.example` since `prod_settings.py` doesn't contain any secrets and I don't know why anyone would need an example for it anymore.

PR #248, opened 2018-02-13
Auto deploy lambda scheduler

jim:
When this is merged, the script that chooses which spider to run will be updated on every deploy.

It is currently setup to run every 30 minutes. When we have more than 48 spiders, we'll need to reassess this to avoid spiders from being skipped.

### DIFF
--- a/deploy/aws_setup.py
+++ b/deploy/aws_setup.py
@@ -117,9 +117,13 @@ def get_definitions(family_name):
 
 
 def update_lambda_function():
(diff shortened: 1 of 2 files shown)


jim:
@r-wei The list of spiders is now generated as a part of the script, so we won't need to manually update the list anymore.

jim:
@r-wei LMK if you think this is ready to merge 😄

r-wei:
Looks good to me :)

Issue #181, opened 2017-11-17
Add missing tests for Chi_animalSpider and fix broken spider


jim:
[There are only a few tests currently](https://github.com/City-Bureau/documenters-aggregator/blob/master/tests/test_chi_animal.py). We should have better coverage for this spider, something more along the lines of [this](https://github.com/City-Bureau/documenters-aggregator/blob/master/tests/test_chi_pubhealth.py).

r-wei:
@milti want to give this a try? There are 2 things that need to be done:
1.  The spider errors out when you try to run it. Running `scrapy crawl chi_animal` gives the following message:
```  
File "/documenters-aggregator/documenters_aggregator/spiders/chi_animal.py", line 41, in parse
    start_time, start_time_str = self._parse_start(text)
  File "documenters-aggregator/documenters_aggregator/spiders/chi_animal.py", line 115, in _parse_start
    naive_date = dateparse(item)
  File ".virtualenvs/documenters-aggregator/lib/python3.6/site-packages/dateutil/parser.py", line 1182, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File ".virtualenvs/documenters-aggregator/lib/python3.6/site-packages/dateutil/parser.py", line 559, in parse
    raise ValueError("Unknown string format")
```

2. We should have more tests for this spider. See Jim's comment above. It looks like the [test file we're currently using](https://github.com/City-Bureau/city-scrapers/blob/master/tests/files/chi_animal.html) is from 2017, so we should replace it with the 2018 version. The new version can be downloaded using [curl](http://osxdaily.com/2014/02/13/download-with-curl/).

milti:
the page added an event that doesnt follow the pattern of the other entries:

2018 Meeting Dates:

    March 15, 2018 - 8:30 a.m.
    **May 17, 2018 - 6:30 p.m. - (Evening)**
    July 19, 2018 - 8:30 a.m.
    September 20, 2018 - 8:30 a.m.
    November 15, 2018 - 8:30 a.m.

discussing this with rebecca, we decided in getting the dateparse function to split the line after the am/pm designation and parse the LHS of the split string and dump or contain the RHS in another object.

milti:
adding the current URL to use as a test file has now broken the test: 

=================================== FAILURES ===================================
___________________________________ test_len ___________________________________
    def test_len():
>       assert len(parsed_items) == 3
E       AssertionError: assert 5 == 3
[...]

PR #251, opened 2018-02-16
Localdev pipe

bonfirefan:
Addreses: #251. Initial pipeline for local csv outputs - creating a pull request for review. Whenever a spider is crawled, a csv file should be created in the documenters_aggregator/local_outputs folder with the agency name and timestamp for testing. Currently, this works for the following spiders: 

- chi_animal
- chi_city_college
- chi_citycouncil
- chi_infa
- chi_library
- chi_parks
- chi_policeboard
- chi_school_actions
- chi_schools
- cook_board
- cook_county
- cook_hospitals
- cook_pubhealth
- il_labor
- il_pubhealth
- metra_board
- regionaltransit
- ward46

The pipeline is outputting blank files for the following spiders: 
- chi_buildings
- chi_police
- chi_pubhealth
- chi_transit
- cook_electoral
- cook_landbank
- ward_night

I'll be continuing to look into whether things are breaking due to the pipeline (conversion errors for CSV possibly) or due to the spider parser. For example, an index error will throw due to #181 when running the chi_animal spider. 

There's also a deprecation warning for the dispatcher, referenced [here](https://github.com/scrapy/scrapy/issues/1762). I'll see if there's a way to implement their recommendations without adding PyDispatcher as a new dependency.


r-wei:
Hey @bonfirefan, looks good. I merged another PR into master that caused a conflict on your branch, so I resolved it.

A couple of thoughts:
- I like the writing to csv. It's super readable when you open it in excel, sheets, etc. 
- I don't think we need latitude and longitude since these will be empty... unless we also want the geocoder to be enabled for everyone?
- I think the start and end times could be even more readable. Maybe "mm/dd/yyyy hh:mm:ss" where hh is between 0 and 23. <<This is a format that google sheets understands as a date time.
- On blank files: My guess is that the spider errors out on the first event (or there are no events). I didn't run all of them, but chi_city_college also returned a blank file for me, while chi_police worked.
- We should add some tests for this pipeline before merging in. We could do something similar to what Pat and I did for the [travis validation pipeline](https://github.com/City-Bureau/city-scrapers/blob/master/tests/test_travis_pipeline.py). Basically, add a json file with some test items, and then check that the pipeline processes them as expected.

easherma:
Just a troubleshooting thought: if the spider writes a blank csv, does it
also write blank json?

Re: lat/lng, IMHO it makes sense to have them there. Either the spider will
run through the geocoder before output in which case those columns will be
populated, or they can be geocoded at a later point and those columns can
be updated.

PR #253, opened 2018-02-18
Rebuild PBCC spider for updated site

### LINKED ISSUE #252: PBCC (chi_buildings) spider needs to be redone
The Public Building Commission of Chicago updated its website (http://www.pbcchicago.com), so the current spider won't work. The good news is now they have all of their calendar events set up on a JSON endpoint, so we can pull all the meetings for the year with a URL like this: http://www.pbcchicago.com/wp-admin/admin-ajax.php?action=eventorganiser-fullcal&start=2018-01-01&end=2018-12-31&timeformat=g:i+a&users_events=0

pjsier:
Closes #252. A more detailed description and location information aren't included in the JSON endpoint, so the spider scrapes those pages if they aren't board meetings (which generally don't have any more info)

### DIFF
--- a/tests/test_chi_buildings.py
+++ b/tests/test_chi_buildings.py
@@ -1,86 +1,84 @@
 import pytest
+import scrapy
 
 from datetime import datetime
 from tests.utils import file_response
 from documenters_aggregator.spiders.chi_buildings import Chi_buildingsSpider
 
 
-test_response = file_response('files/chi_buildings.html')
+test_json_response = file_response('files/chi_buildings.json')
+test_event_response = file_response('files/chi_buildings.html')
 spider = Chi_buildingsSpider()
 # Setting spider date to time test files were generated
-spider.calendar_date = datetime(2017, 10, 15)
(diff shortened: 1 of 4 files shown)


tests/test_chi_buildings.py
```diff
@@ -1,86 +1,85 @@
...
-def test_classification():
-    assert parsed_items[0]['classification'] == 'Daley Plaza'
+def test_end_time():
+    assert parsed_items[0]['end_time'] == '2018-01-04T14:00:00-06:00'
 
 
-def test_afb_classification():
-    assert parsed_items[3]['classification'] == 'Advertisement for Bids'
+def test_id():
+    print(parsed_items[0]['id'])
```
r-wei: remove this line

r-wei:
Awesome! The travis validation also ran :D  There's an extra print statement that I marked inline (not sure if you'll get notified of it..), line 64 of test_chi_buildings.py . Otherwise, looks good.

tests/test_chi_buildings.py
pjsier: Ahh whoops, thanks for catching that!

PR #261, opened 2018-03-08
Repair Chicago Dept of Public Health scraper

### LINKED ISSUE #258: chi_pubhealth is out of date
The url is for 2017: https://www.cityofchicago.org/city/en/depts/cdph/supp_info/boh/2017-board-of-health.html

We need to find the 2018 events page and scrape that instead.

hancush:
Use the 2018 link, adapt `xpaths` to new page, simplify date parsing. 

A couple of Qs:

- I'm skipping events for which a date can't be parsed. Judging by [the placeholder links](https://www.cityofchicago.org/city/en/depts/cdph/supp_info/boh/2018-board-of-health-meetings.html), this should work through this year. However, is this the right thing to do? For example, if they misspell February next year, this spider would silently pass over that event. 
- I think this is right from the text of #258 ("scrape that instead"), but just to confirm: We no longer need to scrape events from 2017?

Closes #258

### DIFF
--- a/documenters_aggregator/spiders/chi_pubhealth.py
+++ b/documenters_aggregator/spiders/chi_pubhealth.py
@@ -43,7 +43,7 @@ class Chi_pubhealthSpider(Spider):
                 date = datetime.strptime(date_text, '%B %d')
 
             except ValueError:
-                pass
+                continue
 
             else:
                 naive_start_time = datetime(year, date.month, date.day, 9)
--- a/documenters_aggregator/spiders/chi_pubhealth.py
+++ b/documenters_aggregator/spiders/chi_pubhealth.py
@@ -16,7 +16,24 @@ class Chi_pubhealthSpider(Spider):
     name = 'chi_pubhealth'
     long_name = 'Chicago Department of Public Health'
     allowed_domains = ['www.cityofchicago.org']
-    start_urls = ['https://www.cityofchicago.org/city/en/depts/cdph/supp_info/boh/2018-board-of-health-meetings.html']
+
+    @property
+    def start_urls(self):
+        """
+        DPH generally uses a standard URL format, but sometimes deviates from
+        the pattern. This property inserts the current year into the standard
+        format, as well as known variants, in hopes DPH sticks to one of their
+        conventions and this scraper does not need to be updated annually.
+        """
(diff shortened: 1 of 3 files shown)


r-wei:
Hey @hancush,

Responses to your Qs:
- Yeah, silent "failures" are something we haven't entirely worked through. Some ideas: (a) You could move `datetime.strptime(...)` outside of the try-except so a spelling error would bubble up. (b) You could just take the first 3 letters of the month, assuming people are less likely to misspell those.
- Yes, the 2017 events are already in City Bureau's database.

Our travisCI runs new/modified spiders on pull requests as a data validation check. I'm looking at the travis log, and it's not getting any items from chi_pubhealth. When you run `scrapy crawl chi_pubhealth`, are you getting items? Depending on your `documenters_aggregator/settings.py`, the items may be writing to the `local_outputs/` folder.

The travisCI data validation is new, so if you are getting items from `scrapy crawl chi_pubhealth`, I can take a look this weekend to see if there's a problem on the data validation end.

hancush:
Hi @bonfirefan @r-wei – Thank you so much for your feedback! I'll be afk at a conference the next couple of days, but I'll look into generalizability ([a perennial problem](https://github.com/opencivicdata/scrapers-us-municipal/issues/161), it seems) and check on the output later this weekend.

r-wei:
@hancush :+1: 

I ran `scrapy crawl chi_pubhealth`, and saw that it's only scraping 2 items (the Jan and Feb ones). The reason our data validation isn't giving any items is because it doesn't validate events in the past.

hancush:
@r-wei @bonfirefan Believe this is ready for another look.

r-wei:
hm, looks like the spider is still only scraping the Jan and Feb events. I think this is because it's only looking for \<a> tags. We want to scrape events in the future, even if there is less information about them.

r-wei:
Thanks @hancush, last thing: The data validation is because the the scraped data doesn't follow the events schema. More details [here](https://travis-ci.org/City-Bureau/city-scrapers/builds/353794876#L828), but let me know if that error message is too terse :)

hancush:
Thanks for your help & patience, @r-wei – looks like tests/validation are passing now! (P.S. Sorry this took so long, something came up on my end. Should be quicker moving forward.)

r-wei:
no worries, thanks @hancush !

Issue #265, opened 2018-03-13
Scraper status at-a-glance


diaholliday:
Could we come up with a way to see which scrapers are broken/working at-a-glance? A real-time status would be ideal but I could also see this being a daily check that displays the status of each scraper currently in operation (i.e. working, broken, needs attention, etc). 

This is the sort of tool that would be ideal for import to the platform down the line—if there's a way to make a working model in the meantime, that'd be ideal. Currently, it's difficult for the Documenters Manager to see if a meeting is not being imported, because the related scraper doesn't show up as broken (it just doesn't show up in the queue at all).

pjsier:
I can take a stab at this, and I'm summarizing a conversation with @diaholliday on Slack here. The current idea is to list each scraper in the README next to an SVG icon indicating its current status, similar to the Travis CI status badge.

The icons can be stored on S3 with filenames based on the scraper name, and we can update each scraper icon automatically after they run. We'll need to find the most straightforward way of detecting whether a scraper failed or needs attention (potentially through the ECS task state, or through added exception handling on Scrapy) and call a Lambda function to update the icon either way.

diaholliday:
This all sounds great! My only request for now is that the display names for each scraper be as close as possible to the real department name (or easy to read abbreviations). Some of the scraper filenames are closer to shorthand—it'd be great if these display names were easy to discern at a glance.

pjsier:
Definitely! I didn't clarify in the issue, but I was thinking it would look something like this. The scraper name would be mostly for the filename

### Scraper Status

| Scraper | Status |
|-------------------------------------------|------|
| Chicago Parks District                    | ![Build Status](https://travis-ci.org/City-Bureau/city-scrapers.svg?branch=master) |
| Chicago Public Schools Board of Education | ![Build Status](https://travis-ci.org/City-Bureau/city-scrapers.svg?branch=master) |

diaholliday:
Yesss, this is great.

jim:
This is up and running!

PR #266, opened 2018-03-14
Create icon dashboard for scraper status

pjsier:
Working on #265. Adds a CloudFormation template that creates the Lambda function and S3 bucket to store icons for each scraper's status. I tested it on my AWS account, and everything runs as expected from a test event (when the event-specific parts are commented out). We should be able to add a trigger from an ECS task state change to get it running, but I'm not sure where in the event object I can pull the scraper name

@jim let me know what you think, and if you have a sense of where we can get the scraper info

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,6 +118,8 @@ src/
 deploy/*.sh
 deploy/*.py
 
+!deploy/scrapers_status.py
+
 # OS files
 .DS_Store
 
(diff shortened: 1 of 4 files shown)


pjsier:
I pulled the scraper name from the task definition after taking a look at the `aws_setup.py` script, so this should be ready for review. There are some details on the last step (connecting the ECS events to Lambda) in [this AWS blog post](https://aws.amazon.com/blogs/compute/monitor-cluster-state-with-amazon-ecs-event-stream/). The relevant part is in "Send event stream events to your Lambda function", and you can ignore the rest

easherma:
This might be better as a different issue, but would someone like a
dashboard be useful to implement?

https://github.com/DormyMo/SpiderKeeper/blob/master/README.md

jim:
@pjsier I'm going to merge this and give deploying it a shot!

PR #269, opened 2018-03-19
update description in chi_school spider

r-wei:


bonfirefan:
This was being used as an example, but with @r-wei 's blessing, we'll close it.

diaholliday:
Has the description also been updated? if so we can take it off the other list as well.

bonfirefan:
Yes, it was updated in the pull request that @ckwms63 made! #281

PR #271, opened 2018-03-19
249 geocode branch

easherma:
Couple things here:

* I added my Pipfile, in case we'd like to use pipenv to manage dependencies as a more elegant solution to the requirements.txt issue. Feel free to ignore it, as it is slightly off-topic. 

* I could use feedback about how the test is implemented. As a developer, it was also confusing to follow existing documentation and code and see clearly how to best incorporate a test into the existing flow. We need to make this easier and more explicit if we want volunteer coders to contribute to functionality as well as adding spiders. 

* As is, the geocoder is pretty dumb, and it doesn't return any metadata about how 'good' of a result it gets back. The test data, luckily or unluckily in this case, is a perfect example of the kind of edge case we are likely to run into. I have some ideas about how to handle this, which also involve the other parts of the pipeline. Either way, this is going to be a pretty extensive conversation that we should focus on soon.


easherma:
This passes the test I wrote (no surprise there), but I'm not convinced its being applied in the pipelines correctly. What are all the pipelines that run in production? Should I be listing them all in order to do local tests? My operating understanding of the pipelines is that they pass each scraped item through in order.

easherma:
not writing/overwriting the location URL atm

easherma:
ok this breaks because I added lat/lng to the csv output, which we can roll back or address ( I'll stop at this point because I feel like I'm starting to risk stepping on some toes)

Issue #140, opened 2017-10-12
save raw data somewhere


eads:

jim:
By "raw data", do you mean the HTML we fetch from the scraped sites? Or the processed items that are sent through the pipeline?

eads:
the processed items -- I'm just thinking it would be good to save the json from each scrape on s3 since it's cheap as hell

jim:
I guess I'm curious what the use case here is. Do we want to have an easy way to load the already scraped data onto a new machine? Or is this more of a historic record of what we've scraped in case we ever want to go back and look at what happened, say, for troubleshooting?

PR #272, opened 2018-03-19
Json feed export

easherma:
addresses #140


easherma:
this should really go to a new branch but I don't think I have permission to do that

PR #274, opened 2018-03-20
WIP: update Park District Board of Commissioners description

diaholliday:
New general description text applies to all chi_parks meetings. I may need to add a line about the board specifically down the line.


bonfirefan:
Addresses #275

PR #277, opened 2018-03-25
Reverse status table order

pjsier:
Changes the layout of the status table for #265. @diaholliday let me know if this is what you were thinking!

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,41 +55,41 @@ B. Research sources for public meetings. Answer questions like: Are we scraping 
 
 ### Chicago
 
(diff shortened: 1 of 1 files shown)


diaholliday:
Yep, that's it, thanks! My thinking on this is basically that the priority info we'll need from the dashboard is whether or not the scraper is working, then which one it is. It's a relatively small change but seeing the status first going from right to left seems like the most intuitive way to arrange it.

Is the next step to merge it? I'll leave that to you. I think I need another quick in-person lesson in GitHub before I start merging things.

PR #280, opened 2018-03-27
Add no-cache to SVG icon headers

pjsier:
Should fix some issues with #265. Sets 'Cache-Control' header for the SVGs on S3


pjsier:
It looks like this is a caching thing, since when I look at some example URLs it works https://s3.amazonaws.com/city-scrapers-status/cook_hospitals.svg

`ward_night` isn't working though, so not sure if the task definition has a different name?

jim:
@pat I checked and there isn't a file in the bucket for the ward night scraper, so it probably is having a problem. The Dashboard is already working!

Issue #275, opened 2018-03-20
Spiders in need of new descriptions (descriptions included)


diaholliday:
This would be a great first issue for non-coders—see Rebecca's video tutorial for instructions: https://www.youtube.com/watch?v=m_MjzgvVZ28

**[Note: When creating your pull request with an updated description, please tag with the Label "Description" and reference Issue #275 in your pull request]**

The following spiders need description edits. I've included new description text for each spider below:

**Cook County Government - JAC Council Meeting (JAC):** (@natashamathur is working)
The Cook County Justice Advisory Council (JAC) is charged with the coordination and implementation of the Cook County President’s criminal and juvenile justice reform efforts and public safety policy development. 

(Also change display name to Cook County Government - Justice Advisory Council (JAC)^):

**Cook County Board of Commissioners**
County Commissioners are elected officials who oversee county activities and work to ensure that citizen concerns are met, federal and state requirements are fulfilled, and county operations run smoothly. The Cook County Board of Commissioners is the governing board and legislative body of the county. It is comprised of 17 Commissioners, each serving a four-year term and is elected from single member districts. Each district represents approximately 300,000 residents.

**Chicago City Council Committee on Housing and Real Estate**
The City Council Committee on Housing and Real Estate hears proposed changes to the City of Chicago’s housing policies. Additionally, the committee hears real estate transactions between the City and other government entities, corporations and institutions.

**Public Building Commission of Chicago**
The Public Building Commission of Chicago oversees and helps ensure quality facilities. Since its inception, the PBC has aimed to enhance education, safety and recreation across Chicago by building or renovating hundreds of schools, city colleges, libraries, parks, fire houses, police stations and other facilities

**Chicago Transit Authority** (Replace current text with the following)
The governing arm of the CTA is the Chicago Transit Board. The Board consists of seven members, with four appointed by the Mayor of Chicago and three appointed by the Governor of Illinois. The Mayor's appointees are subject to the approval of the Governor and the Chicago City Council; the Governor's appointees are subject to the approval of the Mayor and the Illinois State Senate. The Board's separate committees include Human Resources, Strategic Planning, Capital Construction Oversight, and the Finance, Audit & Budget Committee.

**Cook County Land Bank Authority**
[...]

PR #281, opened 2018-03-31
Update CPS desc

ckwms63:


bonfirefan:
This looks good! Thanks @ckwms63

diaholliday:
@bonfirefan this issue was closed but has it been merged? i'm looking at upcoming CPS board meetings in Airtable and they all still say N/A, not sure why.

bonfirefan:
@diaholliday that's odd, it seems to populate with descriptions in my local outputs when I crawl `chi_schools`. The tests should also prevent the spider from crawling without returning the description. Is it not making it to Airtable somehow? Maybe there are no new meetings since last week when we merged this in? @pjsier sees that the spider hasn't run in deployment since April 5th according to the dashboard, so maybe that's the answer.

diaholliday:
Yeah i was thinking it could be because the spider hasn't run since the 5th. Is there a way to make it run now to test it?

bonfirefan:
Addresses #275 
@diaholliday I think that @jim proposed disabling the geocoder pipeline, which means we'll at least be able to populate new events. Not sure what the schedule is on that

jim:
I disabled the geocoding pipeline last night, so events should be showing up again.

PR #288, opened 2018-04-03
WIP 257 spider chi mwrd

easherma:
This PR is for adding the Metropolitan Water Reclamation District spider!

water

A couple notes: 
- since this was on legistar, I used the previous legistar spider code as a guide along with the contrib docs. I tried to leave comments where the code varied, but I'm open to feedback or suggestions. 
- The spider runs successfully, but I have not written tests yet. The tests for the other legistar spiders have a text file instead of an html file, but it wasn't clear to me in the code or documentation where these files came from or were generated. Should I follow the testing pattern for other legistar scrapers? If so, I could use some clarity on that and how/when to generate the test text file(perhaps by running the scraper and writing the results out?), which I can then add to the docs. 
- Some meetings entered have a date but no time. I handled this in my spider, and opted to include these (MWRD dosen't appear to use legistar in a very verbose way, so a blank time may be an all day meeting ,a time that isn't set yet, or something with an arbitrary start such as the 'Study Session' listed on the [calendar](https://mwrd.legistar.com/Calendar.aspx). I'm not sure if we have a defined standard/process for handling ambiguous times and I wasn't able to find an answer in the existing documentation, but again I'm happy to add to the docs once that is clarified.


jim:
Hey Eric,

Thanks for the PR and for asking for clarification on these topics. I'll do my best to answer your questions below. 

> since this was on legistar, I used the previous legistar spider code as a guide along with the contrib docs. I tried to leave comments where the code varied, but I'm open to feedback or suggestions.

I can do a more thorough review of the code when this is farther along, but my initial feedback is to please limit the changes in this PR to just what is needed for the spider. Right now there is a `Pipfile` and another script in the changes that seem unrelated to this work.

> The spider runs successfully, but I have not written tests yet. The tests for the other legistar spiders have a text file instead of an html file, but it wasn't clear to me in the code or documentation where these files came from or were generated. Should I follow the testing pattern for other legistar scrapers? If so, I could use some clarity on that and how/when to generate the test text file(perhaps by running the scraper and writing the results out?), which I can then add to the docs.

For the tests, please follow whatever testing conventions we have in place. For the Legistar spiders, this means storing the data that comes back from the API in a text file and then reading it back in during test execution. (Side note: I'm not totally sure why the [cook_board.txt](https://github.com/City-Bureau/city-scrapers/blob/master/tests/files/cook_board.txt) is a txt file instead of a JSON file; if you'd like, so feel free to use a JSON file for your spider. We can convert the other one later on.)

I'm not sure how the previous legistar fixtures were created. It's possible that @bonfirefan knows as I think she worked on at least one of them, but feel free to be creative about getting that fixture data in place. One idea would be to write a short Python script that uses the same [legistar library](https://github.com/opencivicdata/python-legistar-scraper) to dump the events to the screen (where they could be redirected to a file if needed). We could then reuse that script when setting up future spiders.

> Some meetings entered have a date but no time. I handled this in my spider, and opted to include these (MWRD dosen't appear to use legistar in a very verbose way, so a blank time may be an all day meeting, a time that isn't set yet, or something with an arbitrary start such as the 'Study Session' listed on the calendar. I'm not sure if we have a defined standard/process for handling ambiguous times and I wasn't able to find an answer in the existing documentation, but again I'm happy to add to the docs once that is clarified.

I don't think we have an established practice for handling these situations yet. To keep things simple, let's just ignore events that don't have a valid date and time and write a log statement saying as much. In the future we might have a way to track events with partial data, but I don't think we want to tackle that right now.

Adding this policy to the docs would be great and hopefully help out other devs. Documentation changes that are directly related to code changes are fine to include in the same PR.

Pipfile
```diff
@@ -0,0 +1,90 @@
+[[source]]
```
jim: Please remove this file from the PR.

documenters_aggregator/pipelines/feedWriter.py
```diff
@@ -0,0 +1,41 @@
+import json
```
jim: This file appears to be unrelated to the spider.

easherma:
>For the tests, please follow whatever testing conventions we have in place. For the Legistar spiders, this means storing the data that comes back from the API in a text file and then reading it back in during test execution. (Side note: I'm not totally sure why the cook_board.txt is a txt file instead of a JSON file; if you'd like, so feel free to use a JSON file for your spider. We can convert the other one later on.)

>I'm not sure how the previous legistar fixtures were created. It's possible that @bonfirefan knows as I think she worked on at least one of them, but feel free to be creative about getting that fixture data in place. One idea would be to write a short Python script that uses the same legistar library to dump the events to the screen (where they could be redirected to a file if needed). We could then reuse that script when setting up future spiders.

@jim I've gone back and forth on this aspect and I feel pretty stuck without some further dialogue. In liew of that, can you clarify if by 'API', you meant the city-scraper API or the legistar API?

The legistar scraper doesn't have documentation, and I've started playing around with it but haven't immediately found a stage in it's process where output resembles the output I see for the other text fixtures.

I generated a fixture using our scrapy project itself, but I'm not sure it really makes sense to write tests on that, as I believe I'd be feeding in already scraped and formatted data back into the spider to test, and that doesn't seem to be the objectives of the tests. 

I can really go off in a creative direction here, but I can't help thinking that I might just be missing something more simple that someone who already made these fixtures would know. I'd have to think that if there was a more convoluted process used for making the fixtures for chi_parks or chi_board it would have been documented. Conversely, the chi_citycouncil spider is constructed quite differently for the either two, although its test data seems to more closely match the fixtures I'd generate with my own spider. 
TL;DR I've tried a bunch of different things, and can keep trying. Do you think this process would benefit from some input from someone who wrote one of the other legistar tests and/or generated those fixtures?

jim:
Hey Eric,

> @jim I've gone back and forth on this aspect and I feel pretty stuck without some further dialogue. In lieu of that, can you clarify if by 'API', you meant the city-scraper API or the legistar API?

Sorry if I wasn't clear before. I was referring to the Legistar API generally and not being super specific about a particular component.

> The legistar scraper doesn't have documentation, and I've started playing around with it but haven't immediately found a stage in it's process where output resembles the output I see for the other text fixtures.

Thanks for doing this research. It's too bad it didn't provide an answer, but a worthwhile endeavor regardless.

> I generated a fixture using our scrapy project itself, but I'm not sure it really makes sense to write tests on that, as I believe I'd be feeding in already scraped and formatted data back into the spider to test, and that doesn't seem to be the objectives of the tests.

I completely agree.

> I can really go off in a creative direction here, but I can't help thinking that I might just be missing something more simple that someone who already made these fixtures would know. I'd have to think that if there was a more convoluted process used for making the fixtures for chi_parks or chi_board it would have been documented. Conversely, the chi_citycouncil spider is constructed quite differently for the either two, although its test data seems to more closely match the fixtures I'd generate with my own spider.
> 
> TL;DR I've tried a bunch of different things, and can keep trying. Do you think this process would benefit from some input from someone who wrote one of the other legistar tests and/or generated those fixtures?

I think the reality is that we may never know how the other fixtures were created. Instead of trying to figure that out, let's try to forge ahead and come up with a solution that we are happy with. Maybe we'll even apply it to the other legistar spiders if all goes well.

 How about you try this:

1. Make a small stand-alone Python script that does what the `_make_legistar_call` method in your spider does, but instead of returning the resulting data, it saves that data into a file. Use the file format of your choice for this.

2. Then, in your test, read in the file created in step 1 and and pass it to the `_parse_events` method of the spider instance. Then you should be able to test that the spider returns the expected output.

3. We should probably keep this script created during step 1 in the project for future use, and perhaps we make it into an invoke task. But that can wait until last, and we can talk about that more when you get to it.

Can you give this a try and let us know how it goes? Thanks for working through this despite our lack of documentation.

easherma:
@jim thanks for clarifying. That sounds like a good plan; I'll give it a go as soon as I have a block of time to set aside. I suspect I'll end up submitting a PR for documentation on the legistar scraper as well by the time this gets resolved. :-D

bonfirefan:
Completed with #359

easherma:
@bonfirefan can we document what steps were used to generate the test file? I was working on tests and a script that generates the test file. I can push what I have but the structure of the test file dosen’t quite match.

cherdeman:
Hi @easherma,  I used the legistar events calls in the scraper to get a generator object with the events, along the lines of:
```
import datetime as datetime

les = LegistarEventsScraper(jurisdiction=None, datadir=None)
les.EVENTSPAGE = 'https://mwrd.legistar.com/Calendar.aspx'
les.BASE_URL = 'https://mwrd.legistar.com'
if not since:
    since = datetime.today().year
events = les.events(since=since)
```
Then used a list comprehension to pull the events into a list. I used the json dump method to write out the file, which can't write out generator objects. The second element of each tuple item pulled from events  is a generator, so I replaced it with a 0 to circumvent the error. Note that we want to maintain the tuple (rather than just pulling out the 0 element) so that the parsing method in the spider still works correctly. The result is something along the lines of:
```
import json

event_lst = [(x[0],0) for x in events]
with open('test.json', 'w') as f:
    json.dump(event_lst, f)
```
Note that I ran this in an interpreter and did keyboard interrupt while pulling from the generator into the list since we didn't need absolutely all events in the legistar for this purpose. Certainly an argument for rerunning the file in a manner that can be exactly reproduced more easily but I think it adequately serves the purpose at this time.

Hope this helps!

easherma:
Claire,

Sounds good! I've got something very close to that in the script I was
working on, so good to hear we're on the same page there!

Issue #62, opened 2017-08-16
invoke fails on windows


bonfirefan:
`invoke runtests`
`You indicated pty=True, but your platform doesn't support the 'pty' module!`

Unfortunately pseudo terminal handling is not available for Windows, and the pty=true flag prevents you from invoking runtests. Most people are probably not doing the frustrating thing of trying to run things in Windows like I am, but just case, I know one solution mentioned was to put in an os check.

eads:
I looked around a bit and you should be able to do something like:

```
import sys
# ...
pty = sys.platform != 'win32'
```

eads:
I'm also wondering whether or not to file something against `invoke` ... I feel like it should warn and skip the pseudo terminal rather than bailing out, but I can see the case for how it works now.

easherma:
I ran into similar/other issues with using windows directly. It may or may not be worth it to stub out an issue regarding Windows support/issues?

@bonfirefan were you able to get up and running? Alternative, if you're familiar with using Docker I recently added a Dockerfile that you could use to run on Windows instead

bonfirefan:
Hey there! I would love to test out the dockerfile - I ended up switching completely to Linux, though!

easherma:
Will do! Thanks, @bonfirefan. I too have switched totally over to Linux. 😎

easherma:
@bonfirefan, btw, when i started using pipenv to manage dependencies this problem completely went away, even on windows!

PR #290, opened 2018-04-04
Creating a prerequisites file

bonfirefan:
Adding some additional resources for a first-time development environment setup, with @ckwms63 's help.

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -1,0 +1,28 @@
+---
+title: Prerequisites
+---
+
+<h1 class="hidden">Prerequisites</h1>
+
+This document covers some of the issues associated with first-time development environment setup.
+
+## Github
+
+### Create a Github account 
+If you do not have an account already, go to [Github.com](https://github.com/) and sign up for an account.
+
+### Installing Github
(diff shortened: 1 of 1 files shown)


bonfirefan:
@r-wei I think I'll merge this in EOD if we don't see any mis-informations in the file as it would mainly be fore first-time coders who need a bit a catchup reference

r-wei:
@bonfirefan looks good to me, thanks for doing this :) I'm cool with iterating on this as Caroline suggested. Maybe also add a link from the README to this prerequisites page?

r-wei:
Another suggestion from @diaholliday -- maybe call this something like "additional resources" or "background materials"? so "prerequisites" doesn't suggest to people that it's something they have to be an expert at before joining

bonfirefan:
OK, I'll merge this in and we'll continue to edit this! What @easherma [mentioned with pipenv](https://github.com/City-Bureau/city-scrapers/issues/62#issuecomment-379005737) (which would be greaaaaat!) will hopefully change the Windows warning. Emmanuel said he'd give some feedback, too.

PR #291, opened 2018-04-04
updates doc links in  code and template to point to GH pages

easherma:
I noticed the spider template was pointing to the gitbooks version of the docs, which appear to be broken. It looks like we switched to GH pages for docs, so I updated all the URLS to be current.


bonfirefan:
Yeah those old gitbooks links are dead. Unless @eads wants to revive them or @jim thinks that gitpages isn't a permanent home either and we should link directly to the in-project doc files, this looks good to me.

Issue #276, opened 2018-03-20
Separating CPD beat meetings and labeling each district by neighborhood


diaholliday:
Is there a way to separate Chicago Police Department beat meetings from the more general CPD meetings in the Chicago_police spider? Or, alternatively, can we tag all beat meetings in the chicago_police spider with the following description (i.e. just the beat meetings, not the other meetings being scraped from the same  site https://home.chicagopolice.org/get-involved-with-caps/all-community-event-calendars): "CPD Beat meetings, held on all 279 police beats in the City, provide a regular opportunity for police officers, residents, and other community stakeholders to exchange information, identify and prioritize problems, and begin developing solutions to those problems." 

To take this one step further, is it possible or feasible to sort out each CPD district or create a separate spider for each CPD district so that we can tag each with its corresponding neighborhood coverage areas? That would be the ideal version of this spider.

easherma:
These are great ideas! I’ve been using the cpd spider for my local tests, so I could take a look and think through this.

diaholliday:
Sounds good, thanks! Let me know if I can be helpful in any way.

easherma:
@diaholliday If I can assume this ask is to better differentiate CPD meetings types on the admin side, I think a better approach than writing new spiders for each would be to do some additional sorting/tagging/labeling on the backend(after the data is scraped). This could still be done programatically before you'd have to touch the data.

I'm  not sure how flexible Airtable is in this regard, but it's something that would be totally feasible using some flat files, a Django app, or in any relational database.

diaholliday:
Yep, it's more for the admin side for now. Additional labeling/tagging does seem like a better solution than creating new spiders. In that case, this can hold until the platform is built.

r-wei:
resolved by #294

PR #294, opened 2018-04-05
Updated description for chi_police

wildisle:
I think this is good to go!


r-wei:
Addresses #276

bonfirefan:
Also addreses #275

r-wei:
These changes update the chi_police description and classify events as either "Beat Meeting District ##" or "District Advisory Committee". If there are no objections, I'll merge this in EOD.

PR #295, opened 2018-04-06
Scraper status deployment

jim:
This patch updates the deployment scripts so that the latest version of the scraper status lambda function is deployed automatically when code is pushed to production.

cc @pjsier


jim:
@pjsier We run a build on Codebuild on every push to master, and if it passes, `deploy/aws_setup.py` will run. That includes updating both lambda functions.

I'm hoping we don't actually need the cache clearing script and that the `cache-control` header we added fixes that. If it doesn't, we can add the cache clearing as an additional build step.

pjsier:
Got it, I think I was looking for that in Travis. It looks like we've still got some caching issues, even with `no-cache`, but they've at least gotten better. In that case this looks good to go!

PR #297, opened 2018-04-09
Update long name and descrip of chi_city_college spider.

novellac:
Based on issue #275


novellac:
I think this is good to go - note that there were no tests for the description, so only the spider has changed.

r-wei:
Ah, thanks for catching that, @novellac. Could you add a test for the new description on this pull request? Then we'll merge it in.

novellac:
Yes I'll try to add a test today.

novellac:
Chi city colleges tests seem to have passed. @r-wei is there anything more I need to do?

r-wei:
Looks good to me, thanks!

PR #300, opened 2018-04-10
Rename documenters_aggregator to city_scrapers.

jim:
This patch (almost) completes our transition from "documenters aggregator" to "city scrapers". (The exception is a little bit of AWS stuff that I don't want to mess with right now.)

Everything _should_ still work, but this is a lot of changes, so please glance through this and see if I missed anything.

The rationale behind this change is that we don't have to explain to new folks (or existing folks) why we have a strangely named folder that most of the code lives in 😄 .


jim:
(I'm looking into the failing tests. Looks like I touched some whitespace I shouldn't have 😄).

r-wei:
Hey @jim , I'm still seeing a couple instances of the old name. Updated the gitignore, but not sure about these in the deploy folder:
- ./deploy/aws_setup.py:PROJECT_SLUG = 'documenters_aggregator'
- ./deploy/scrapers_status.py:PROJECT_SLUG = 'documenters_aggregator'
- ./deploy/scrapers_status.py:    # Pull scraper name from ARN in documenters_aggregator-{SCRAPER}
- ./deploy/scraperScheduler/index.js:        taskDefinition: `documenters_aggregator-${task}`,

jim:
@r-wei Thanks for taking a look. I think those occurrences of the old name are OK as they are all in the AWS deployment stuff (names for ECS tasks, docker images, etc.) I want to update those, too, but was hoping to do that as a separate task.

The tests failing, of course, is not OK and I'll need to address that. 😄

jim:
I will refrain from posting additional 😄 on this PR. Things are getting out of control.

PR #304, opened 2018-04-11
Fix getting scraper name

### LINKED ISSUE #301: Scraper status Lambda function not updating on error
Some spiders have been failing, but the status Lambda didn't update their icons at all. I'm guessing this is because the status isn't listed as 'STOPPED' (https://github.com/City-Bureau/city-scrapers/blob/master/deploy/scrapers_status.py#L45), but there are also some other things it would throw errors for that should show up in the logs

pjsier:
Closes #301. The code for getting the scraper name was pulling the version number instead. This fixes it, and it's currently running in the Lambda. It also looks like GitHub is handling the `no-cache` header the way it should, so the dashboard should work as intended now. This also uses `timedelta` to get the time in CST rather than importing an external library to keep the setup simple

### DIFF
--- a/deploy/city-scrapers-status/scrapers_status.py
+++ b/deploy/city-scrapers-status/scrapers_status.py
@@ -1,6 +1,6 @@
 import os
 import boto3
-from datetime import datetime
+from datetime import datetime, timedelta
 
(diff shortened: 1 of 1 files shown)


jim:
Woot! 🏆

PR #308, opened 2018-04-13
Update description in regionaltransit (RTA) spider

o-stovicek:
From issue #275 
The code around the description was a bit different from the other spider I edited, so hopefully this works.

### DIFF
--- a/documenters_aggregator/spiders/regionaltransit.py
+++ b/documenters_aggregator/spiders/regionaltransit.py
@@ -48,10 +48,10 @@ class RegionaltransitSpider(Spider):
         event standard <http://docs.opencivicdata.org/en/latest/data/event.html>`_.
         """
 
-        description = ("The RTA Board of Directors is a 16-member group of professionals"
-                      "governing the activities and initiatives of the RTA. The RTA is "
-                      "charged with financial oversight, funding, and regional transit "
-                       "planning for the region’s transit operators: the Chicago Transit
+        description = ("The RTA Board of Directors is a 16-member group of professionals "
+                       "governing the activities and initiatives of the RTA. The RTA is "
+                       "charged with financial oversight, funding, and regional transit "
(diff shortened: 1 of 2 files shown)


documenters_aggregator/spiders/regionaltransit.py
```diff
@@ -17,7 +17,12 @@ class RegionaltransitSpider(Spider):
     domain_root = 'http://www.rtachicago.org'
 
     def parse_iframe(self, response):
-        description = response.request.meta['description']
+        description = ("The RTA Board of Directors is a 16-member group of professionals"
+                      "governing the activities and initiatives of the RTA. The RTA is "
+                      "charged with financial oversight, funding, and regional transit "
+                       "planning for the region’s transit operators: the Chicago Transit
```
easherma: looks like you're just missing a closing quote at the end of this line!

tests/test_regionaltransit.py
```diff
@@ -16,7 +16,12 @@ def test_name():
 
 
 def test_description():
-    assert parsed_items[0]['description'] == 'Test Description'
+    assert 'Test Description' == ("The RTA Board of Directors is a 16-member group of professionals"
+                        "governing the activities and initiatives of the RTA. The RTA is "
+                        "charged with financial oversight, funding, and regional transit "
+                        "planning for the region’s transit operators: the Chicago Transit
```
easherma: it looks like a closing quote is missing here as well. Also, Python is sensitive to [indentation](https://stackoverflow.com/questions/13884499/what-is-python-whitespace-and-how-does-it-work).  I'd recommend formatting these quotes to look more like the description in your spider, including the way the quotes are indented (which is essential) and the spacing (which just helps readability)

o-stovicek:
Passing checks now, so it would be great if someone could take a look at this!

bonfirefan:
@o-stovicek this looks good, merging it in! Thanks for the reviews, @easherma

Issue #296, opened 2018-04-07
Detroit Board of Education (DPSCD)


jmcbroom:
New scraper for the DPSCD Board of Education from http://detroitk12.org/board/meetings/

vincecima:
@jmcbroom if no one else has claimed this one, I would love to take a shot.

vincecima:
@jmcbroom I just saw the spreadsheet mentioned in (https://github.com/City-Bureau/city-scrapers/issues/292). Happy to take a shot at something else, let me know either way.

jmcbroom:
@vincecima please, feel free! I can grab a different one

PR #317, opened 2018-04-22
0296 spider det schools

vincecima:
Spider for https://github.com/City-Bureau/city-scrapers/issues/296. Some inline comments below.

Thanks!

### DIFF
--- a/documenters_aggregator/spiders/det_schools.py
+++ b/documenters_aggregator/spiders/det_schools.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 import scrapy
+import re
+from datetime import datetime
 from documenters_aggregator.spider import Spider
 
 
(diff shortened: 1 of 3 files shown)


documenters_aggregator/spiders/det_schools.py
```diff
@@ -0,0 +1,152 @@
...
+                'start_time': self._parse_start(item[2]),
+                'end_time': self._parse_end(item[2]),
+                'timezone': 'America/Detroit',
+                'status': self._parse_status(item),
+                'all_day': self._parse_all_day(item),
+                'location': self._parse_location(item[3]),
+                'sources': self._parse_sources(item),
+            }
+
+            data['id'] = self._generate_id(data, data['start_time'])
```
vincecima: Not sure if this is a common Python idiom, but the idea of calculating this value twice (locally and in the super class) seemed sort of awkward. Changing the initial `id` to something like `event_id` or changing the 2nd `id` to `internal_id` might make it clearer?

documenters_aggregator/spiders/det_schools.py
```diff
@@ -0,0 +1,152 @@
...
+        Get next page. You must add logic to `next_url` and
+        return a scrapy request.
+        """
+        next_url = None  # What is next URL?
+        return scrapy.Request(next_url, callback=self.parse)
+
+    def _parse_id(self, item):
+        parsed_url = urllib.parse.urlparse(item)
+        parsed_query = urllib.parse.parse_qs(parsed_url.query)
+        return parsed_query['eid'][0]
```
vincecima: I am using the Google Cal event ID as the unique ID string. They look like `MmlhcTcyNW50Nm43dWZqN3BpOWwzYmF1ZzRfMjAxODA1MTVUMTMwMDAwWiA4bmpua21kbzgxcDdyZGw0MjA4dDI2MmM2b0Bn`. Any problems with the length?

vincecima:
Not sure about that failure, is there somewhere where we have to add a `det` as a prefix? `invoke runtests` does work on my laptop.

vincecima:
Ok, setup a `virtualenv` locally and still can't recreate the issue.

bonfirefan:
Hey @vincecima I couldn't replicate the issue locally either, but I saw that it has to do with the folder name transition of `documenters_aggregator` to `city_scrapers`. I merged in the changes from the master branch into your branch and moved your scraper to the `city_scrapers` folder to match what it is in deployment. Let me know if these changes look good to you!

vincecima:
Thanks @bonfirefan, looks like that resolved it. I will fix up the last validation error and we should be good to go.

documenters_aggregator/spiders/det_schools.py
jim: Hmm, interesting. That is a really long ID. It's ugly but shouldn't break anything... 🤞

documenters_aggregator/spiders/det_schools.py
jim: I think that pattern might have started with some code I wrote, so highly unlikely that it's a Python idiom. I think using a different name like `event_id` would be nice, but let's hold off on that change for now as all of the spiders would need to updated.

documenters_aggregator/spiders/det_schools.py
```diff
@@ -0,0 +1,152 @@
+# -*- coding: utf-8 -*-
+import scrapy
+import re
+import urllib.parse
+from datetime import datetime
+from documenters_aggregator.spider import Spider
```
jim: I think this should be `from city_scrapers.spider import Spider` now.

bonfirefan:
Exciting! I think this would be our first Detroit spider. @diaholliday and @jim do we want this going to a separate airtable? Or is that already set up?

diaholliday:
Hm. Seems like a new Airtable base in the same account as the Chicago meetings Airtable might work best. Here's a link to a clean base (now called "Detroit meetings"): https://airtable.com/tbltbkAxLsKjcHcjI/viwr3NhWcwVI1pr1t

@jim let me know if you see any issues with this approach. I figure it's mostly temporary for now anyway. When the platform is launched it'll contains both sets of meetings.

vincecima:
Sounds good, I will keep an eye on the other PR so I can better understand the data pipeline.

PR #322, opened 2018-04-26
Airtable for detroit

r-wei:
This code creates a new function in the Airtable pipeline that sets the Airtable keys based on geography (Detroit or Chicago). It's a temporary solution so that the Detroit spiders write to a different table until the Platform is built (see PR #317 ). 

@jim , @pjsier , what do you think? I haven't actually changed the AWS environment variables. Wanted to see if this code was ok first.

@diaholliday , if this goes through, then the Detroit spiders will write to this table: https://airtable.com/tbltbkAxLsKjcHcjI/viwr3NhWcwVI1pr1t  And maybe we can even make a more public view for it: https://guide.airtable.com/public-web-database/  (temporary of course pending the Platform)


pjsier:
This should work, but we'll need to update `aws_setup.py` with the new environment variable names in the [`ENV_VARS` list](https://github.com/City-Bureau/city-scrapers/blob/master/deploy/aws_setup.py#L25), remove and re-register all of the current Batch job definitions, and maybe change the names to start with `CITY_SCRAPERS`.

These are all pretty small changes though, so this approach looks good overall!

Issue #324, opened 2018-04-29
cook_pubhealth errors when start_time is None


r-wei:
`scrapy crawl cook_pubhealth` gives the following TypeError:

```
2018-04-29 09:39:47 [scrapy.core.scraper] ERROR: Error processing {'location': {'url': None, 'coordinates': {'latitude': None, 'longitude': None}, 'address': None, 'name': None}, 'id': 'cook_pubhealth/None/57/new_rrp_law_goes_into_effect', 'description': 'The U.S. Environmental Protection Agency’s (USEPA) new rule, the Renovation, Repair and Painting (or RRP) law effective April 22, 2010. The new rule requires use of lead-safe work practices and other actions aimed at preventing lead poisoning. Contractors performing renovation, repair and painting projects that disturb lead-based paint in homes, child-care facilities, and schools built before 1978 must be certified by the USEPA and must follow specific work practices to prevent lead contamination.When: Thu Apr 22, 2010', 'timezone': 'America/Chicago', 'status': 'tentative', '_type': 'event', 'all_day': False, 'sources': [{'url': 'http://www.cookcountypublichealth.org/events-view/57', 'note': ''}], 'name': 'New RRP Law Goes into Effect', 'end_time': None, 'start_time': None, 'classification': 'New RRP Law Goes into Effect'}
Traceback (most recent call last):
  File "~/.virtualenvs/documenters-aggregator/lib/python3.4/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "~/repos/city-scrapers/city_scrapers/pipelines/localExporter.py", line 52, in process_item
    new_item['start_time'] = datetime.datetime.strftime(new_item['start_time'], '%Y-%m-%d %H:%M')
TypeError: descriptor 'strftime' requires a 'datetime.date' object but received a 'NoneType'
```

start_time is a required field, so we should first check why the spider is returning None sometimes and write code to try to get those start times. A benefit of this error is that we are alerted when a required field is missing.

r-wei:
@cherdeman is working

r-wei:
resolved by #325

PR #325, opened 2018-05-01
WIP: Fixed pipeline start time error

cherdeman:
Added additional try/except to localExport to address issues arising from start times with None values


r-wei:
resolves #324

PR #326, opened 2018-05-01
Description for the City Council

ckwms63:


r-wei:
looks good! :) addresses #275

PR #329, opened 2018-05-01
Create windows-remote-setup.md

bonfirefan:
This option allows users on any environment to code on a Cloud-based IDE for free.


bonfirefan:
@pjsier See if you can check this over! @jim @r-wei we'd likely want this as our alternative for Windows users, as it doesn't require additional installations or partitioning.

docs/windows-remote-setup.md
```diff
@@ -0,0 +1,50 @@
...
+You'll get the option of choose your python version. If Python 3.6 is selected, press ENTER, or type 2 to select 3.6.
+
+9. Check that `python3` points to Python 3.6: 
+```
+python3 -V
+```
+10. One of the project dependencies is the `Inflector` python package. It will fail unless we set local encoding settings:
+```
+sudo apt-get install language-pack-en
+export LC_ALL=en_GB.UTF-8
```
jim: I think these environment variables would need to be set in `.bash_profile` or some other dotfile so they get loaded every time?

bonfirefan: I was wondering that. They were installed as a workaround for an error encountered during installation for `Inflector`: `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 375: ordinal not in range(128)` 
The dependencies in requirements.txt installed fine, but it's probably a good idea to set these in `.bashrc`

pjsier:
@bonfirefan I needed to add `sudo apt-get install -y python3.6-dev` for Twisted to install without an error about missing `Python.h`, but otherwise it works! I ran `invoke runtests` and everything passed.

We might be able to simplify this by putting all of these steps in a bash script so that people can just clone the repo and run that. Also, would it make sense to leave the default `python3` as-is, since we're setting up a virtual environment anyway? It would be one less step, and I've heard there can be complications with overriding `python3` on Ubuntu (although it probably doesn't matter that much in an online IDE)

bonfirefan:
Okay, great! Oh no - `python3.6-dev` was definitely an installation I needed as well, but left it out in these instructions!

If you want to come up with a modified version of [this](https://gist.github.com/bonfirefan/e9ebe1dec5bb86a89a30204558e8106c)  that adds the `-y` flags, @jim was thinking the same thing.

easherma:
Nice work! This is super groovy. :)

bonfirefan:
Thanks @pjsier for setting that up! I've placed that within the Codeanywhere instructions. I've added a note on Windows within the installation guide, which links to this file. It also now links out to the Docker instructions to avoid confusion. I've also numbered `storing-data.md` to `04_storing-data.md` because we've been missing our 4th reference file for a while - it seems to make sense, but let me know if that's the wrong order.

pjsier:
@bonfirefan looks good! Only thing that might be worth changing is the name of the virtualenv. `cityscraperenv` isn't in the `.gitignore` file, so if someone uses that and runs `git add .` they might add it to a commit by mistake. We could move it to another directory or change the name to one of the ones listed in the `.gitignore` like `venv` or `ENV`?

bonfirefan:
@pjsier could we specify that they create the virtualenv outside of their Github project folder?

jim:
@pjsier @bonfirefan Did either of you have success with connecting CodeAnywhere to GitHub? I've tried it twice and each time I ended up with an empty project that wasn't configured to push code to GitHub without setting up credentials in some way.

bonfirefan:
Hey @jim, same thing happened for me. I just cloned the repo into the development environment instead of trying to connect within their application. Should we specify that?

jim:
@bonfirefan I think it's OK for folks to do it manually if needed. We should make sure that it's a smooth transition from this guide to the other one we link to. I think we can merge this in and then make tweaks as needed going forward. It would be cool to give it a try on Monday night if any new folks come without a dev environment setup.

bonfirefan:
@jim Once everything is installed, you treat the IDE like you would your own computer and run a normal `git clone` command - we'll see how well that translates as they get to Next Steps in Step 3. I'll merge this in and agree that we'll likely continue to edit it here and there.

Issue #328, opened 2018-05-01
Revise Contribution guide to increase visibility into work


jim:
* _New_ Developers should open a `[WIP]` pull request when they start working on an issue so that we can see what is being worked on.
* _Update_ Mention that issues without activity for 30 days may be reassigned or closed (just like we do with PRs).

This came out of a convo with @diaholliday.

cc @r-wei @bonfirefan

r-wei:
resolved by #349

Issue #318, opened 2018-04-23
New Spider Request: CPS Community Action Council


diaholliday:
Link to the site is here: http://cps.edu/FACE/Pages/CAC.aspx

Looks like a pretty straightforward spider with one main exception—meetings aren't listed individually, they're listed on a recurring schedule:
![screen shot 2018-04-23 at 6 36 31 pm](https://user-images.githubusercontent.com/4017450/39158392-67f0f952-4725-11e8-8b41-695fd1f99cd4.png)

brettvanderwerff:
Can you assign this to me?. I am an open source newbie, but have some scraping experience with my own pet projects. I'm still looking at your code base to figure out how things run, but would like the chance to try this if you are willing to give it to me. Thank you.

diaholliday:
Hi @brettvanderwerff! It's all yours. I'm pretty sure no one has started it yet. Are you able to create pull requests? Let me know if you need a certain level of access that you don't already have. @jim @bonfirefan or I can help you get set up.

brettvanderwerff:
Do you wan't me to put in the pull requests as soon as I set up the new branch to let people know it is a work in progress?

diaholliday:
I'll wait for @jim or @bonfirefan to let us know what the next best step is. Don't want to tell you the wrong thing.

diaholliday:
Hey @brettvanderwerff I just double-checked procedure on this. This conversation is sufficient as a claim on the spider. No need to start a PR at this point. This spider is all yours!

brettvanderwerff:
Ok thanks I'm starting out on this, I'm new to scrapy, so I'm just getting up to speed on the documentation it might take me a little bit to get up and running. I'll put in a WIP pull request once I have some worthwhile code.

bonfirefan:
@brettvanderweff, my apologies on the delay here! You’ve totally got it, all you need to do to claim an issue is to comment! And yes, whenever you feel at a comfortable point, please do make a pull request for review. We of course won’t merge it in until completion and you’ll be able to continue to make changes.

brettvanderwerff:
I just want to keep the lines of communication open on this and let you know that I am still engaged in this and plugging along. There are a lot of firsts in this one for me.

r-wei:
@brettvanderwerff , thanks for the update. Feel free also to ask questions in slack or show up to an in-person coding session: https://www.citybureau.org/city-scrapers-get-involved (scroll to the bottom for the calendar)

[...]

PR #330, opened 2018-05-02
WIP: 0318 spider cps community action council

brettvanderwerff:
This is a rough start to the #318  CPS community action council spider. I made the pull request because of issue #328, which seemed to look for new contributors to open pull requests early for development visibility. I am pretty new so a lot of this is in the very, very early stages and not really operational.


jim:
Hi @brettvanderwerff! Thanks for opening this PR. This way we know that you're working on this scraper and can try to provide whatever support you need. Please let us know on this PR or in Slack if you have any questions and we'll do our best to answer them!

r-wei:
Hey Brett, thanks for reaching out. Here are some responses to the problems:

1. I’m not seeing any strange encodings in the scraped data or errors when I run this spider. Maybe you’ve already fixed this? You can also use the [strip()](https://docs.python.org/3/library/stdtypes.html#str.strip) method to remove whitespace on the ends of strings.

2. The dates look correct to me. Since the events are recurring, would you be able to write code to generate all the events through the end of 2018? So for the Austin one, this would be: May 8, June 12, July 10, etc--all the way through December. I know that the [ward night spider](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/ward_night.py) does this, and maybe @jim has hints on how to do easily. But if you feel it's too much work, we can merge this in, and open another issue to refactor this spider to get all the recurrences.

2 and 3. Humboldt Park — East: Let’s leave this event off for now. I’m not seeing these events listed in Humboldt Park East’s website either, so we can have someone reach out to Carl Hurdlik as a non-coding task.

4. It looks good. The comments are helpful. :)

brettvanderwerff:
I get the weird encoding when I view the locally outputted data I see things like "Humboldt Park â€“ East" instead of "Humboldt Park East", maybe it is just my environment though?  If you don;t see it I'm not going to worry about it. 

I will try to generate all the events through the end of 2018 and will look at  the ward night spider for a better idea of what you are talking about. 

I'll try to write in a mechanism to skip Humboldt Park — East.

I should be able to get tho these in the next few days I hope. 

Thanks for the review.

brettvanderwerff:
This spider is basically done with one exception:

I addressed most of @r-wei's comments, the last thing is to write code to generate all the events through the end of 2018. Could someone shine more detail on what the expected output format for this would be?

I tried looking at the output of the wardnight spider that was suggested as an example  by running "scrapy crawl ward_night" in the terminal, which give me a csv like it should  but unfortunately the csv is empty so I can't see the output you are shooting for.  Please let me know thanks.

r-wei:
Hey @brettvanderwerff , generating all the events means creating a bunch of scraped items (dictionaries) with identical data except for the start_time. For some context, we're feeding the scraped items into a database where each row is a different event. Does that make sense?

Thanks for the tip about the ward night spider. I'll take a look into that.

brettvanderwerff:
@r-wei Thanks for the response, I think I can see how that would be done without changing the spider too much. I'll start work on that.

r-wei:
looks good, thanks @brettvanderwerff

PR #336, opened 2018-05-07
Test Update for Chi_animalSpider #181

cherdeman:


r-wei:
looks good! resolves #181

PR #338, opened 2018-05-08
Update description in cook_landbank spider

o-stovicek:
From issue #275


o-stovicek:
Can someone help me figure out what's going wrong here? The test appears to be failing based on asserting that the description... is the identical description

o-stovicek:
Looks like that worked!

bonfirefan:
Hey @o-stovicek, simplistically the difference is between equality and identity. Generally when checking if two values are equal we use `==` and if we are checking the identity of object like `None` we use `is`. Definitely not an obvious distinction on first pass.

Issue #192, opened 2017-11-27
new spider request: cook county housing authority


r-wei:
Up for grabs. See [docs on contributing a spider](https://city-bureau.gitbooks.io/documenters-event-aggregator/contribute-a-spider.html). Site to scrape: http://thehacc.org/events/

r-wei:
Non-coding tasks:
- [ ] should these events be update manually or by a spider?
- [ ] if manual, add these events to airtable

emmanuellyautomated:
currently working on this one!

diaholliday:
Hey @emmanuellyautomated, are you still working on this spider? Would love to add it to the set. Just wanted to check in while it's on my mind. We can re-assign it if you'd rather hand it off.

diaholliday:
hey @emmanuellyautomated just wanted to check in on this, are you still working on this spider or should we put back out to the group? No worries if you have to let it go—I'll plan to go ahead and re-assign it if I don't hear back from you by May 4.

diaholliday:
@emmanuellyautomated I appreciate you work and time on this but I need to go ahead and re-assign this spider so we can start sending Documenters out to these meetings on a more regular basis.

r-wei:
@cherdeman is working

r-wei:
resolved by #339

diaholliday:
Can we add this to the Scraper Status section too? @pjsier

PR #339, opened 2018-05-08
Cook County Housing Authority spider #192

cherdeman:


r-wei:
Yep, you can use [curl](http://www.compciv.org/recipes/cli/downloading-with-curl/), and save the file in the [city-scrapers/test/files folder](https://github.com/City-Bureau/city-scrapers/tree/master/tests/files).

cherdeman:
Followup, presumably I'll need to aggregate the html of each individual event page into one file?

r-wei:
@cherdeman , you could just download one of the individual event pages using curl and test the [_parse_event function](https://github.com/City-Bureau/city-scrapers/pull/339/files#diff-f4f4e6d144d4b9009ffee4f32c8b3c2bR40). The [test for cook_pubhealth](https://github.com/City-Bureau/city-scrapers/blob/master/tests/test_cook_pubhealth.py) is an example of testing against a single event page.

cherdeman:
@r-wei Got it! I added tests and the static file, let me know if you need any updates.

r-wei:
Hey @cherdeman , can you make sure all the functions (individual tests) in test_cook_housingauthority.py and test_chi_library.py have unique names? For example, if you have multiple tests called `test_allday`, pytest only runs one of them. Other than that, looks good!

cherdeman:
Ah my bad @r-wei! I think I caught all of them this time, let me know if you notice any more.

r-wei:
resolves #340 #192

PR #343, opened 2018-05-10
Add scraper status SVGs to README

pjsier:
Adds the 3 new scrapers to the status table in the README (the community action council one hasn't finished building, but the link is right). We should probably add a line about this to contributing documentation too


pjsier:
Will merge this tomorrow if no one has any issues, and added the most recent scraper

PR #347, opened 2018-05-14
WIP: Report to Sentry if a Scraper fails

vincecima:


vincecima:
@jim yes, will give this a shot tomorrow and then I will check in via Slack to see if I am going down the right road.

PR #349, opened 2018-05-14
Add to FAQ

r-wei:
I added a question and answer for how to run a spider that requires a Google Sheets API key.


r-wei:
Hey @jim , what do you think about this PR to address #328 ?

bonfirefan:
Can we merge these in? This seems to address #328 well:

- [x] _New_ Developers should open a [WIP] pull request when they start working on an issue so that we can see what is being worked on.

- [x] _Update_ Mention that issues without activity for 30 days may be reassigned or closed (just like we do with PRs).

Issue #257, opened 2018-02-26
Spider Needed: Metropolitan Water Reclamation District of Greater Chicago


diaholliday:
This one looks pretty straightforward and uses the same system as Chicago City Council: https://mwrd.legistar.com/Calendar.aspx

nickjordan:
Looking into this.

diaholliday:
Thanks @nickjordan! Any updates so far?

diaholliday:
Btw here's some background on why we're interested in the MWRD meetings: https://www.bettergov.org/news/shaw-time-for-a-watchdog-at-metropolitan-water-district?rf=head&eType=EmailBlastContent&eId=e6dc14d0-782b-4e2b-993e-67ec1f88f1d3

eads:
Waaay back when, I met a guy running for MWRD president and he made no bones that it's a way into the political machine.

easherma:
True story. A little way back when, when I was more involved in local
politics, I got offered a 'chance' to run for one of those positions.

diaholliday:
Anyone interested in building this scraper if @nickjordan can't? I'd love to start sending Docs to them going forward. @nickjordan still all yours if you've already started it and/or want to keep working on it.

easherma:
@diaholliday I can grab this one, it's been some time and we haven't heard from @nickjordan , and there is a soft spot in my heart for the MWRD. :-)

diaholliday:
Woot! Thanks, I'm v curious about those meetings. Can't wait to send Documenters to them.

easherma:
MWRD uses Legistar, relevant issues #33, #74

easherma:
[PR opened](https://github.com/City-Bureau/city-scrapers/pull/288)

diaholliday:
hey @easherma can you update me on the status of this PR? I'm going through the April milestone list to see if I can help move anything forward. No worries, if you're still working on it!

easherma:
The spider works but tests aren't written yet.

bonfirefan:
Completed with #359

PR #359, opened 2018-05-18
renamed spider and test files, address PR  #288 and issue #257

cherdeman:
I may have done the branching wrong on this but hopefully the pull works!


bonfirefan:
Hey @cherdeman, thanks for writing out the tests and creating the test file! [Travis does an additional check ](https://github.com/City-Bureau/city-scrapers/blob/master/travis/validate_spiders.sh) on new spiders[ to validate their output](https://github.com/City-Bureau/city-scrapers/blob/master/tasks.py#L161), and in the process calls the spider with `scrapy crawl`. It looks like the crawler itself is not in a working state and breaks when parsing the name of the item. When you run `scrapy crawl chi_water --loglevel=ERROR` you should be able to replicate the issue.

bonfirefan:
So, it looks we're returning a tuple from legistar - the first item is the dictionary we want, the second one a Legistar class object. If you set `item = item[0]` in the events `for` loop, it should hopefully resolve the issue.

cherdeman:
Thanks @bonfirefan! updated the test file to match and I think it's good to go.

bonfirefan:
This looks ready to merge in! Pinging @diaholliday  - the only thing I think we should clarify is that the spider currently skips Water Reclamation events with no meeting time. When you [look at the events](https://mwrd.legistar.com/Calendar.aspx) you can see that those sessions are usually "Study Sessions" presenting different committee processes (Budget, Procurement) or findings/discussion (Ethics - having an inspector general, Engineer - managing organic material). These are usually scheduled "after the regular board meeting" and seem to vary in actual start time from 12pm-1:30pm. Given that the Board meeting re-convenes as the Committee of the Whole (listed as two separate events) on the same day, then re-convenes for these study sessions, would we be ok including these study sessions with a placeholder time of the board time + 15 minutes, or no time at all?

bonfirefan:
Actually, for these Study Sessions, it looks as if the date is not being captured at all, though it is being converted and the event itself gets scraped. @cherdeman Could we keep the start date and set the time to midnight if there is no time?

cherdeman:
@bonfirefan I think that should do it, let me know if there are still remaining issues with the time.

PR #369, opened 2018-05-25
Update for new page layout

### LINKED ISSUE #353: IL Labor Relations Board spider isn't finding any items

mkrump:
- Update to reflect new page layout.
- Fixes #353

### DIFF
--- a/city_scrapers/spiders/il_labor.py
+++ b/city_scrapers/spiders/il_labor.py
@@ -32,7 +32,7 @@ class Il_laborSpider(Spider):
         """
 
         # There's not a lot of structure on this page, so this selector is pretty fragile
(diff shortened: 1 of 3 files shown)


mkrump:
For [State Panel Meeting ](https://www2.illinois.gov/ilrb/meetings/pages/default.aspx) there is now two locations. I wasn't sure how that should be handled. For now just concatenated with an 'Or'.

pjsier:
It should be fine there, but since the locations are pretty far apart and it's a live stream it almost seems like it should return two events? Could be something to think through for the event schema

r-wei:
@pjsier , ooh, I like that. Let's discuss it on the events schema doc

PR #374, opened 2018-06-04
New events schema

r-wei:
The new events schema is in this PR (docs/06_event_schema.md) as agreed upon by the City Scrapers committee.

Other changes:
- new spider.generate_id and spider.generate_status helper functions that agree with the new events schema
- updated chi_police spider and its test to the new events schema as an example
- disabled `test_id`'s that now fail with the new spider.generate_id function
- temporary fix for bug unrelated to events schema: a chi_schools_community_action_council test fails because the result is dependent on the month in which the spider is run
- temporary fix for bug unrelated to events schema: cook_housingauthority spider error (Json TypeError)
- updated travis validation pipeline and its tests with the new events schema
- disabled travis validation for all the spiders except chi_police as a temporary fix so that this PR will pass (the next PR #375 will re-enable travis validation for all the spiders)

If there are no major objections, I'll merge this in around 5pm, so this infrastructure is in place for tonight's in-person session. Sorry about the short notice! Feel free to make changes directly to this branch @pjsier @jim @bonfirefan

### DIFF
--- a/city_scrapers/pipelines/TravisValidation.py
+++ b/city_scrapers/pipelines/TravisValidation.py
@@ -1,5 +1,5 @@
 import re
-from datetime import datetime
+from datetime import date, time
 from pytz import timezone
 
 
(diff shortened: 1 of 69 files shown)


pjsier:
This looks great! I'm not entirely sure about using a list for locations though. I think I mentioned it, but it seems like multiple instances of an event at the same time but different places should be treated as two separate events.

If it's a livestream (which we've seen) then it doesn't seem important to treat that as a separate location, and if it is actually the same topic and time but in different locations that feels like it could be a separate assignment.

Also, one small thing, but if `start` and `end` are dictionaries, do the keys need to include `start_` and `end_`? We should be able to do something like

```python
{
  "start": {
    "date": DATE,
    "time": TIME,
    "notes": NOTES 
  }
}
```

r-wei:
@pjsier sounds good, I made those changes: 
- only one location (location is a dict not a list, some instructions in the events schema about multiple locations)
- removed prefixes from the start and end dictionaries
- added a classification field to differentiate between board and various committee meetings (Darryl asked for this)

Looks like the build is passing [here](https://travis-ci.org/City-Bureau/city-scrapers/pull_requests), but github is not showing it on this page...

Issue #380, opened 2018-06-05
Update spiders to new events schema


r-wei:
We just released a [new events schema](https://github.com/City-Bureau/city-scrapers/blob/master/docs/06_event_schema.md)! 

Changes:
- New `documents` field for agendas and meetings notes
- More useful and consistent `classification` and `status` fields
- Changes to the types of the `start` and `end `fields to allow for looser requirements if the start or end times are missing on the website
- `timezone` field moved to a spider-level attribute

This means we have to update all of the spiders to the new schema. We kicked it off at our June 4 in person session. Instead of opening a different issue for each spider, we'll organize who is doing what on this [sign up sheet](https://docs.google.com/spreadsheets/d/1wkSl3XkhEQhWYh3Dq-G8SS0uNXThqQ8CM0LJhe5i2FM/edit#gid=0)--please sign up there. 

For a guided walkthrough of how to update a spider (as well as useful helper functions), check out [these instructions](https://docs.google.com/document/d/1DOfx4NFbSWGwhSI8_XkN__V5dVc_VVR8oevOVoXbxgI/edit).

r-wei:
hey @laurieskelly @myersjustinc I see your usernames on the spreadsheet. are you still working on updating the chi_animal spider? https://docs.google.com/spreadsheets/d/1wkSl3XkhEQhWYh3Dq-G8SS0uNXThqQ8CM0LJhe5i2FM/edit#gid=0  

Feel free to [join our slack channel](https://github.com/City-Bureau/city-scrapers#ready-to-code-with-us) if you want to chat about the project in real time.

myersjustinc:
@r-wei Got my bearings again, and my [380-spider-chi_animal](https://github.com/myersjustinc/city-scrapers/tree/380-spider-chi_animal) branch already was pretty close to having that new schema in place. Did some rebasing and other cleanup in there, and all the tests pass. I still need to look at why I don't see anything written to disk when I `scrapy crawl chi_animal`, even though the logging messages all look good. Hoping to figure that out relatively soon and have a PR to you in the next day or two.

r-wei:
@myersjustinc thanks! You might try checking which item pipelines are turned on in your settings: https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/settings/base.py

As a baseline, I turn off all my pipelines and check that something like `scrapy crawl chi_animal -o chi-animal.json´ writes the meetings to chi-animal.json

PR #383, opened 2018-06-09
WIP: 380- update chi_transit

thenoelman:


city_scrapers/spiders/chi_transit.py
```diff
@@ -13,10 +13,12 @@
 
 class ChiTransitSpider(Spider):
     name = 'chi_transit'
-    long_name = 'Chicago Transit Authority'
+    agency_id = 'Chicago Transit Authority'
     allowed_domains = ['www.transitchicago.com']
     base_url = 'http://www.transitchicago.com'
-    start_urls = ['https://www.transitchicago.com/board/notices-agendas-minutes/']
+    timezone = 'America/Chicago'
+    START_URL = ['https://www.transitchicago.com/board/notices-agendas-minutes/']
```
bonfirefan: Is there a reason you have changed the start_url to a nested list? This could be what is throwing Travis off.

thenoelman: I'm sure I made lots of mistakes!  I was attempting to steal code from update #378 , and I was following along with the videos on the website.  I'm on slack if you have the time to walk me through what needs to be fixed.

bonfirefan:
Hey @thenoelman, this is looking good! It looks like a few things still need to be updated to the new events schema: start_time, end_time. The status seems to also have changed to "tenative". Instructions for the new schema are [here](https://github.com/City-Bureau/city-scrapers/issues/380)

thenoelman:
actually, there were quite a few bugs on both ends.  I'm finally debugging locally.  I'm confident that what I've got coded in my spider is correct, but I'm struggling to get my test case up and running.  I have a question- is the date supposed to be formatted as "date(YYYY, mm, dd)" and the time as "time(HH24, mm)"?  That's how it was in the example.  I fear that I may have taken things too literally, because now the test case is complaining about my date/time formatting.  Here are screenshots of my spider output (see date and time highlighted), and what my test case is complaining about.  Note- I'm just trying to get the test case up and running.  I haven't actually built tests for the changes yet.  It's complaining about the datetime from I think line 158 where I reformatted it.  Should I not have done so?

bonfirefan:
Hey @thenoelman, I'm trying to replicate your errors but don't seem to see the updated date functions. Could you commit your changes and we can see what's going on? It looks like the issue is in the date parsing within the spider itself. Somehow a newline ('\n') has made its way into the datestring fed into the date parser, which is not matching the datestring format that it's expecting.

thenoelman:
@bonfirefan I pushed my chi_transit.py changes.  Take a look and let me know how we can best update the test case.  Anyway, I think the only thing that needs updating at this point is the test_case.

bonfirefan:
@thenoelman hey there! Wanted to check in and see how things are going with this issue.

PR #391, opened 2018-06-13
Update wayne_cow schema for #380

mkrump:
- update schema
- replace static description
- scrape documents

### DIFF
--- a/city_scrapers/spiders/wayne_cow.py
+++ b/city_scrapers/spiders/wayne_cow.py
@@ -1,12 +1,16 @@
 # -*- coding: utf-8 -*-
-import scrapy
 from datetime import datetime
+from urllib.parse import urljoin
+
 from city_scrapers.spider import Spider
 
 
 class Wayne_cowSpider(Spider):
     name = 'wayne_cow'
+    agency_id = 'Detroit Committee of the Whole'
+    timezone = 'America/Detroit'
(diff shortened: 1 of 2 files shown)


city_scrapers/spiders/wayne_cow.py
```diff
@@ -1,12 +1,17 @@
...
 from datetime import datetime
+from urllib.parse import urljoin
+
 from city_scrapers.spider import Spider
 
 
 class Wayne_cowSpider(Spider):
     name = 'wayne_cow'
+    # TODO add agency_id
+    agency_id = '????'
```
mkrump: Wasn't sure what `agency_id` should be. Should be ready once that is updated.

r-wei: @diaholliday , what should be the official name for this agency? https://www.waynecounty.com/elected/commission/committee-of-the-whole.aspx

r-wei: @novellac , do you know the official name of this agency? https://www.waynecounty.com/elected/commission/committee-of-the-whole.aspx

diaholliday: Sorry, @r-wei, didn't realize that was your question (and was generally tied up in Detroit). The name is the Committee of the Whole (AKA COW). We could call it committee_whole unless it needs to be one word.

PR #393, opened 2018-06-14
Update regionaltransit schema for #380

mkrump:
- update schema
- change name to il_regional_transit
- replace static description with info at bottom of the page
- use classification field to differentiate board, finance committee and citizens advisory council


city_scrapers/spiders/il_regional_transit.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @staticmethod
+    def _parse_classification(item):
+        name = item.get('name', '')
+        if 'Committee' in name:
+            return 'Committee'
+        if 'Board' in name:
+            return 'Board'
+        if 'Commission' in name:
+            return 'Commission'
```
r-wei: Hey @mkrump , could you add a case here for "citizens advisory"? It looks like they don't have any meetings posted right now, but it's called a variety of things: "citizens advisory council", "citizens advisory board", "citizens advisory committee". In any case, we would want to return "Citizens Advisory Council" to differentiate these meetings from one run by RTA board members. Otherwise, looks good! Thanks for knocking out all these spiders!

PR #396, opened 2018-06-18
Update cook_landbank for #380

vincecima:
Adding this PR and updated Google Sheet to reflect I am working on this.

Question/To Do Log:
1. Agenda section's structure changes, making a best effort to extract all the text. Would we rather extract the raw HTML?
1. If no to HTML, will strip off Agenda leading text in description.
1. Update test(s)
1. Implement `documents` field
1. Restore `long_name` field


r-wei:
hey @vincecima , since we're treating agendas as documents, you can just return a url to the agenda

vincecima:
@r-wei thanks for the feedback. The issue in the spreadsheet mentioned that we wanted to replace the static description with `event specific description`. Would that be best served by the agenda's contents or just a repeat of the event name (`CCLBA Board of Directors Meeting`)?

r-wei:
@vincecima for the event specific description, we could scrape the information under "Event Details" and stop at the agenda. Something like:
"The Land Transactions Committee will convene on Friday, June 8th at the hour of 10:00 AM at the location of 69 W. Washington St., 22nd Floor, Conference Room ‘B”, Chicago, Illinois, to consider the following:\n\n\nAGENDA\n(TBD or url)"

vincecima:
Ok, @r-wei I think this is all set for review. Thanks!

vincecima:
@r-wei ok, I think my changes are ready to be reviewed. Getting that `module not found error`. Made sure to merge in master and its still happening. Looks like a couple of the other PRs are having the same issue. Will take another shot at replicating the issue locally later. Thanks!

r-wei:
Thanks @vincecima ! I may not find time to look into this this weekend, but it might be related to PR #406. [This file](https://github.com/City-Bureau/city-scrapers/blob/master/travis/debug_validations.sh) is meant to simulate the travis validation pipeline locally.

@pjsier , could you take a quick look to see if something from that PR could be causing this?

pjsier:
@r-wei yep, this build ran while the bug was still in place. I restarted the build so the module problem goes away, but it looks like there are still a few small issues

vincecima:
Thanks for the help and for pointing me at that simulator script. Will get this fixed up and ready this weekend.

vincecima:
Ok, everything fixed up and I corrected the start/end field structures. Thanks!

PR #399, opened 2018-06-19
Update city college spider

jim:
This updates the Chicago City College spider for 2018 and a new website.


jim:
Re: #380 🎉

r-wei:
Just a small of schema thing:
- we're using 'event_description' instead of 'description' now

PR #401, opened 2018-06-19
380 metra board

ab1470:


r-wei:
new schema looks good! Can we make 2 changes:
- change the spider's name to il_metra_board
- instead of "transit" in the classification field, say either "board" or "citizens advisory"

city_scrapers/spiders/il_metra_board.py
```diff
@@ -54,7 +55,14 @@ def _parse_classification(self, item):
...
         """
-        return 'transit'
+        full_name = item.css('td[headers=Name]::text').extract_first()
+
+        if "Board Meeting" in full_name:
+            return "Board Meeting"
+        elif "Citizens Advisory Board Meeting" in full_name:
+            return "Citizens Advisory"
+        else:
+            return ''
```
r-wei: @ab1470 this is the right idea :) However, since "Board Meeting" is a substring of "Citizens Advisory Board Meeting", line 63 will never run and no meetings will be classified as "Citizens Advisory".

I also see that there is a safety committee meeting posted for July. Let's classify that one as a "Commitee Meeting". 

Almost there! :)

r-wei:
@ab1470 , I left a comment inline with the code. Posting about it here just in case you don't see it.

ab1470:
Ahhh, sorry 😕. I should have caught that. 

Unfortunately I can’t make it tonight, but I will fix it tomorrow and submit a new PR.

PR #403, opened 2018-06-19
\#380 Updating il_labor spider

noahkconley:


r-wei:
Hey @noahkconley , just a couple of things:
- looks like we're not getting the address for the 11am meeting. maybe the site was recently updated? looks different than what I remember on Monday
- the "description" field is now called "event_description"

noahkconley:
@r-wei I've addressed the comments above

PR #404, opened 2018-06-19
Setup pipenv

jim:
TODO:
- [x] Figure out how to get Legistar to be installable with pipenv.


Pipfile
```diff
@@ -0,0 +1,26 @@
...
+airtable-python-wrapper = "*"
+"boto3" = "*"
+inflector = "*"
+invoke = "*"
+python-dateutil = "*"
+requests = "*"
+"jinja2" = "*"
+usaddress = "*"
+pytz = "*"
+
```
easherma: legistar = {git = "https://github.com/opencivicdata/python-legistar-scraper", editable = true, ref = "eb93f5b9f00529ac6a69c62aa542b0c2357000d0"}

easherma:
@jim  I have been using pipenv for awhile with this project, this is from my Pipfile:

`legistar = {git = "https://github.com/opencivicdata/python-legistar-scraper", editable = true, ref = "75a93ec42ab186a7ea4927be390a8bbb4148eaaf"}
`
Note: the above ref was for an earlier branch that had removed pupa as a dependency, I believe that it eventually was merged to master, so a newer ref might be preferred. I left a comment on the Pipfile with the latest ref but didn't test it. 

Hope that helps! :)

jim:
@easherma Thanks for the pointer. That ended up working out just fine.

PR #407, opened 2018-06-19
WIP: update cook_hospitals schema

mkrump:
- update schema
- use classification field to differentiate board and various committee meetings
- remove static description
- scrape documents

### DIFF
--- a/city_scrapers/spiders/cook_hospitals.py
+++ b/city_scrapers/spiders/cook_hospitals.py
@@ -7,6 +7,9 @@ from city_scrapers.spider import Spider
 
 class Cook_hospitalsSpider(Spider):
     name = 'cook_hospitals'
+    agency_id = 'Cook County Health and Hospitals System'
+    timezone = 'America/Chicago'
+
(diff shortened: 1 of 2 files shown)


city_scrapers/spiders/cook_hospitals.py
```diff
@@ -21,118 +24,80 @@ def parse(self, response):
...
+    @staticmethod
+    def _parse_name(item):
         """
         Get event name from item's text
         """
         return item.xpath('text()').extract_first().strip()
 
-    def _parse_description(self, subitem):
+    @staticmethod
+    def _parse_description(subitem):
```
mkrump: The signup spreadsheet said to remove the static description, but I didn't immediately see what should replace it on http://www.cookcountyhhs.org/about-cchhs/governance/board-committee-meetings/

r-wei: Looks like there's not much, so you could just return an empty string for the event description. We want to remove the static descriptions because they are actually descriptions of the host agencies and not of the events.

PR #413, opened 2018-07-01
Fix broken link to event schema documentation.

aeggenberger:


aeggenberger:
It looks like the test failed because it was run after the end of June 2018. I'm not sure whether or how to fix it. This pull request is a small change to the documentation.

Issue #370, opened 2018-05-26
Chi_buildings spider revision


diaholliday:
This spider is working fine but it's pulling a majority of unrelated meetings. Can we revise it so that it only displays board meetings from the Public Building Commission?

r-wei:
This spider scrapes a JSON endpoint (something like [this](http://www.pbcchicago.com/wp-admin/admin-ajax.php?action=eventorganiser-fullcal&start=20180528)), so making this change should be straightforward. I think filtering on the `className` field should do the trick.

cherdeman:
Hi @diaholliday and @r-wei  I can grab this!

r-wei:
Great :) I'll sign you up here: https://docs.google.com/spreadsheets/d/1wkSl3XkhEQhWYh3Dq-G8SS0uNXThqQ8CM0LJhe5i2FM/edit#gid=0  --there are some additional notes

r-wei:
resolved by #416

PR #416, opened 2018-07-03
Chi Buildings Revision #370

cherdeman:
This is updated to pull only board meetings, audit committee meetings, and admin opp committee meetings. I saw the note about parsing descriptions but it appears that that functionality is built in, could you give me more context on that? @r-wei @diaholliday


r-wei:
wow, there is a lot of logic about parsing descriptions already there... but it looks like the current descriptions are mostly null, so I think it's ok if we change it. Instead, let's put the text that we get from item.get('description').

cherdeman:
Got it, I think I pulled that through here. LMK if there's anything else.

r-wei:
looks good thanks!

PR #417, opened 2018-07-03
update CAC spider to the new schema

r-wei:
- removed static descriptions
- added neighborhood websites to sources 
- made sure the tests don't depend on the month that the spider is run

@jim or @bonfirefan , can you double check that the tests for this spider don't depend on the month that it's run?


pjsier:
It's working for me, and it doesn't look like the tests should break depending on the month. Merging now

PR #421, opened 2018-07-05
Update templates to reflect new schema

mkrump:
Notice the templating code was a little out of sync with new schema. Updated templates / tests to match the updated schema. Changes below:

- Add 'status' test to tests
- Add 'documents' to test / spider
- Remove 'timezone' from template methods / tests
- Update 'location' test template to match new schema
- generate_id only takes data now
- Change 'start_time' -> 'start'
- Change 'end_time' -> 'end'
- 'yield data' in spider template is out of scope of loop (just need to indent)


templates/spider.tmpl
```diff
@@ -21,22 +22,21 @@ class {{ classname }}(Spider):
 
             data = {
                 '_type': 'event',
-                'id': self._parse_id(item),
                 'name': self._parse_name(item),
                 'description': self._parse_description(item),
```
r-wei: we should have 'event_description' instead of 'description'

templates/spider.tmpl
```diff
@@ -6,6 +6,7 @@ from city_scrapers.spider import Spider
 class {{ classname }}(Spider):
     name = '{{ name }}'
     long_name = '{{ long_name }}'
```
r-wei: agency_id instead of long_name

templates/spider.tmpl
```diff
@@ -21,22 +22,21 @@ class {{ classname }}(Spider):
...
-                'id': self._parse_id(item),
                 'name': self._parse_name(item),
                 'description': self._parse_description(item),
                 'classification': self._parse_classification(item),
-                'start_time': self._parse_start(item),
-                'end_time': self._parse_end(item),
-                'timezone': self._parse_timezone(item),
+                'start': self._parse_start(item),
+                'end': self._parse_end(item),
                 'status': self._parse_status(item),
```
r-wei: I think we'd also like to encourage people to use the _generate_status method. maybe move this below the _generate_id line and put `data['status'] = self._generate_status(data, text='')` ...

templates/spider.tmpl
```diff
@@ -108,13 +92,9 @@ class {{ classname }}(Spider):
...
             'address': '',
-            'coordinates': {
-                'latitude': '',
-                'longitude': '',
-            },
+            'name': '',
+            'neighborhood': '',
         }
 
     def _parse_status(self, item):
```
r-wei: we'd remove this function if using _generate_status

r-wei:
great, thanks!

PR #426, opened 2018-07-09
CS-419 wayne_health_human_services First pass

novellac:
Related to #426 
Feedback needed, especially for:
 - creating the ID - it keeps coming across as 000000, but I think it should be a date.
 - date parsing - output like in chi_animal, but not sure if this is still an ideal way. Better to split out time and date objects as in chi_school_actions
 - status - seems to default to "confirmed" no matter what I do. Is it possible (and worth it?) to add date logic for passed meetings?

### DIFF
--- a/city_scrapers/spiders/wayne_health_human_services.py
+++ b/city_scrapers/spiders/wayne_health_human_services.py
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
 
-# WE ARE BRINGING IN A MIXIN WHICH IMPORTS OTHER LIBRARIES.
+# THIS SPIDER USES A MIXIN FOR SHARED FUNCTIONALITY.
 # MIXINS ARE STORED IN /city-scrapers/city-scrapers/mixins
-# YOU CAN TAKE THE DEFINITIONS OUT OF THE MIXIN AND ADD THEM HERE IF THEY ARE
-# UNIQUE.
+# YOU CAN OVERRIDE THE MIXIN HERE BY CREATING YOUR OWN DEFINITION.
 
 from city_scrapers.spider import Spider
 from city_scrapers.mixins.wayne_commission import Wayne_commission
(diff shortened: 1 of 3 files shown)


novellac:
Note also that test coverage is not complete - ID and status aren't being checked.

city_scrapers/spiders/wayne_health_human_services.py
```diff
@@ -0,0 +1,95 @@
...
+        Parse start date and time.
+        """
+        # Dateparse can't always handle the inconsistent dates, so
+        # let's normalize them using scrapy's regular expressions.
+        month_str = item.xpath('.//td[2]/text()').re(r'[a-zA-Z]{3}')[0]
+        day_str = item.xpath('.//td[2]/text()').re(r'\d+')[0]
+        time_str = item.xpath('.//td[3]/text()').extract_first()
+        date_str = dateparse('{0} {1} {2} {3}'.format(month_str, day_str, self.yearStr, time_str))
+
+        return self._naive_datetime_to_tz(date_str, source_tz='America/Detroit')
```
mkrump: @novellac change this line to `return {'date': date_str.date(), 'time': date_str.time(), 'note': ''}` and I think that should fix the id / status issue. The schema / `start` object changed recently https://github.com/City-Bureau/city-scrapers/blob/master/docs/06_event_schema.md.

r-wei:
looks good to me :) @novellac , did you have anything else you wanted to add?

novellac:
This one is probably ok, but I'd like to leave the TODOs in there. It's part of group 13, and the other spiders have some logic for meeting status that will need adding.

novellac:
Oh, I take it back @r-wei I need to make sure it's getting a good ID first! Spoke/wrote too soon!

novellac:
Okey all the checks pass. I think this one is ready for review/primetime. :)

r-wei:
@novellac , looks good! The only thing is to remove the long_name attribute. We're using agency_id instead. Sorry about the confusion!

novellac:
I was getting some weird errors (I don't remember what they were when I removed long_name).

r-wei:
@novellac , could you try reproducing the errors and posting them?

novellac:
Reworked this spider to work with the mixin wayne_commission. CI checks will fail till the mixin is merged, I think.

pjsier:
Replaced by #459

Issue #418, opened 2018-07-03
New spiders needed


diaholliday:
**Chicago Department of Planning and Development** (5 commissions, all appear to be on separate pages):

- [x] Chicago Zoning Board of Appeals: https://www.cityofchicago.org/city/en/depts/dcd/supp_info/zoning_board_of_appeals.html
- [x] Chicago Plan Commission: https://www.cityofchicago.org/city/en/depts/dcd/supp_info/chicago_plan_commission.html
- [x] Commission on Chicago Landmarks:
https://www.cityofchicago.org/city/en/depts/dcd/supp_info/landmarks_commission.html
- [x] Community Development Commission:
https://www.cityofchicago.org/city/en/depts/dcd/supp_info/community_developmentcommission.html
- [x] Chicago Development Fund:
https://www.cityofchicago.org/city/en/depts/dcd/supp_info/chicago_developmentfund.html

- [ ] **Chicago Housing Authority**: http://www.thecha.org/about/meetings-agendas-and-resolutions/ 

- [ ] **Chicago Board of Election Commissioners**: http://www.chicagoelections.com/en/board-meetings.html

**Mayor's Pedestrian Advisory Councils** (2 councils on separate pages):
- [ ] Mayor's Pedestrian Advisory Council:
http://chicagocompletestreets.org/getinvolved/mayors-advisory-councils/
- [ ] Mayor's Bicycle Advisory Council:
http://chicagocompletestreets.org/getinvolved/mayors-advisory-councils/

If you do claim a spider, please make a note via comment below and on the Agency Table: https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=0

diaholliday:
@r-wei would you mind checking to see if the Chicago Housing Authority scraper is running? I see one meeting for it on the backend but I feel like there should be more.

diaholliday:
Just got a response from the HACC, they added the board meetings to the event calendar. They should show up now but I still need to confirm that the committees are included.

hancush:
hey @diaholliday, i don't have write access on the table, but i left a comment – i can take on the Mayor's Bicycle Advisory Council spider.

PR #427, opened 2018-07-09
Create Chicago Development Fund spider for #418

mkrump:


r-wei:
Just a couple things:
- we're not using the long_name attribute anymore
- the id's should be unique for each event. when I run the spider, I get 57 events and only 52 unique ids.

city_scrapers/spiders/chi_development_fund.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @staticmethod
+    def _parse_start(meeting):
+        # Not all dates on site a valid dates (e.g. Jan. 2011), so try to parse
+        # and return none if not possible
+        try:
+            dt, other_text = dateutil.parser.parse(meeting, fuzzy_with_tokens=True)
+            # based on Agenda mtg time seems to vary but time not stated mtg desc
+            name = [s.strip() for s in other_text if s.strip()]
+            return ' '.join(name), {'date': dt.date(), 'time': None, 'note': ''}
```
r-wei: For the 'note' field, can we put 'see agenda document for time'. It looks like the times are usually in those pdfs. otherwise, looks good.

PR #428, opened 2018-07-09
Add decorator to catch pipeline errors

vincecima:
Uses Python's `functools` library to create a decorator that will catch, report to Sentry and re-raise any exception that surfaces in the wrapped method.

Adding screenshots below to demonstrate what data shows in Sentry.

Please let me know if I should add the code to put the decorator on all the pipelines.

exception-pipeline-name

^ shown stack trace is from the correct location and shows the pipeline name.

exception-item

^ extra data includes all the inputs to the wrapped function (in this case the item being passed to the pipeline)

### DIFF
--- a/city_scrapers/pipelines/item.py
+++ b/city_scrapers/pipelines/item.py
@@ -1,9 +1,14 @@
+from city_scrapers.utils import report_error
+
+@report_error
 class CityScrapersItemPipeline(object):
     """
     Pipeline for doing any custom processing on individual
     items, i.e. assigning the long name to agency_name
     """
     def process_item(self, item, spider):
+        raise Exception("HELLO")
+        
         if hasattr(spider, 'long_name'):
             item['agency_name'] = spider.long_name
         else:
--- a/city_scrapers/pipelines/item.py
+++ b/city_scrapers/pipelines/item.py
@@ -1,14 +1,14 @@
 from city_scrapers.utils import report_error
 
-@report_error
 class CityScrapersItemPipeline(object):
     """
     Pipeline for doing any custom processing on individual
     items, i.e. assigning the long name to agency_name
     """
+    @report_error
     def process_item(self, item, spider):
         raise Exception("HELLO")
-        
+
         if hasattr(spider, 'long_name'):
             item['agency_name'] = spider.long_name
         else:
--- a/city_scrapers/pipelines/item.py
+++ b/city_scrapers/pipelines/item.py
@@ -7,8 +7,6 @@ class CityScrapersItemPipeline(object):
     """
     @report_error
     def process_item(self, item, spider):
-        raise Exception("HELLO")
-
         if hasattr(spider, 'long_name'):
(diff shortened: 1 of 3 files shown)


city_scrapers/pipelines/item.py
```diff
@@ -1,9 +1,14 @@
+from city_scrapers.utils import report_error
+
 class CityScrapersItemPipeline(object):
     """
     Pipeline for doing any custom processing on individual
     items, i.e. assigning the long name to agency_name
     """
+    @report_error
```
vincecima: This is here as an example

jim:
> Because this project uses scrapy_sentry, it doesn't appear there is anywhere else to grab the Sentry client object. How do we feel about using the util method that scrapy_sentry exposes?

I think it's OK. At least that way we are only configuring the Sentry client in one place.

> What context do we want logged along with exception?

I think we'd want:
* pipeline name (I assume we can get this from the stack trace?)
* the item that is being processed

That's enough to assist in troubleshooting.

> Are there any exception types we don't want caught?

Hmm, probably, but I don't know what they are. Let's catch them all for now and once we see how this behaves we can add a list of exceptions to ignore if needed.

> The actual reporting doesn't log anything in development mode, but I believe thats intended as we only configuration Sentry in production mode

Yeah, I think that is fine.

> I don't believe this method will work when wrapping spider's parse method because of an interaction with Scrapy's defer utilities. Is that an issue?

Since the `scrapy_sentry` catches errors in the spiders, that isn't the end of the world. I am curious, though, to learn more about this. I'm fairly new to Python and not all that familiar with how that part of Sentry operates. Can you point me to any resources you came across so I can read up on it for my own benefit?

city_scrapers/utils.py
```diff
@@ -15,3 +17,19 @@ def get_key(the_dict, location_string):
         return reduce(getitem, location_string.split('.'), the_dict) or ''
     except (KeyError, TypeError):
         return None
+
+def report_error(f):
+    @wraps(f)
+    def wrapper(*args, **kwds):
```
pjsier: This is a small thing, but the convention for keyword arguments is `**kwargs`, so we should update that here from `**kwds`

pjsier:
@vincecima This looks great! I tested it out locally with the production settings and it created an error report on Sentry (sorry to anyone who's going to get that email 😬). I left one small comment, but other than that if you could remove the test exception I think this should be ready to merge

pjsier:
@vincecima do you think you'll be able to make these last few changes? If not we can also push them to your branch so that we can still merge the PR

vincecima:
@pjsier yes, should have something in the next 36 hours

vincecima:
Just checking in, still working on it.

vincecima:
t

vincecima:
@pjsier updated the PR description, think this one is ready to rock.

pjsier:
This looks great, thanks! Squashing so that the git history is clear and merging now

PR #431, opened 2018-07-10
Reactivating CSV exporter and adding pydispatch directly

bonfirefan:
One-liner to get rid of the xlib/pydispatch deprecation warnings that may/may not complicate other exporter issues. PyDispatch is already in our requirements file.


bonfirefan:
Sadly, is not the solution to Issue #432

pjsier:
Closing since this was addressed in #434. Feel free to reopen if I'm missing something!

Issue #432, opened 2018-07-10
Successful crawls are no longer outputting new CSV files


novellac:
[432-Traceback-MissingCsvOutput.txt](https://github.com/City-Bureau/city-scrapers/files/2178499/432-Traceback-MissingCsvOutput.txt)

Issue: Successful crawls are no longer outputting new CSV files

Expected Result: When the command `scrapy crawl <spider name>` is run, a CSV file should be output to the folder `/city-scrapers/city-scrapers/local_outputs/`

Hints: May be a recently deprecated library. As per @bonfirefan and @jim, we may have updated our libraries at some point and need to add in the `pydispatch` library into our dependencies.

bonfirefan:
So, I tested adding in pydispatch and it doesn't seem to be the issue here.

bonfirefan:
@pjsier could this be related to the [s3 pipeline](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/pipelines/s3_item.py)?

pjsier:
@bonfirefan when I uncomment the CSV pipeline in `city_scrapers/settings/base.py` it works for me, otherwise I can't reproduce this. We might want to leave this on by default if it doesn't hurt anything for local development

bonfirefan:
@pjsier no wonder, that explains the local output portion. I think these are possibly two separate errors - it would be good to uncomment out the local exporter pipeline unless it was causing other errors?

Separate issue below. Here's what I'm seeing even after successfully outputting to csv after uncommenting the csv pipeline is output that matches @novellac's error output:

```
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable profile from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable config_file from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable credentials_file from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable data_path from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable profile from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable region from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable profile from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable ca_bundle from defaults.
2018-07-09 20:27:07 [botocore.session] DEBUG: Loading variable profile from defaults.
[...]

PR #434, opened 2018-07-10
Change S3 item pipeline import for logging

pjsier:
Fixes the logging errors in #432. Because the S3 item pipeline was getting imported from the `pipelines` `__init__` file, all Scrapy environments were trying to load `boto3` and logging errors when it failed to find AWS credentials set up locally.

I split out the import so that it's only included in the production environment, and now I don't see any `boto3` logs locally when running spiders in development (I saw logs for it before, but they didn't throw errors)


r-wei:
this removes the connect timeout error in #432 for me :)

bonfirefan:
This also removed the errors for me!

Issue #419, opened 2018-07-05
Wayne County Commission on Health and Human Services


novellac:
New spider for Wayne County Commission on Health and Human Services (wayne_health_human_services) from 
[https://www.waynecounty.com/elected/commission/health-human-services.aspx](https://www.waynecounty.com/elected/commission/health-human-services.aspx)

This spider is currently part of ticketing group 13 on the [Public Meetings Aggregator](https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297).

novellac:
Needs labels:
 - location: Detroit
 - architecture: spiders
 - new spider needed

pjsier:
Closed by #459

PR #436, opened 2018-07-10
435 wayne_public_services new spider

novellac:
This spider is same as spider #419 (the tests are slightly different, of course).


r-wei:
@novellac , if these spiders are the same, is there a way we can write one wayne_county spider class and then have each commission spider inherit from it? It looks like there will be a bunch of these, and it would be better for maintenance not to have so many copies of the same code.

novellac:
I'd like to try to run the new spider on the rest of the ticket group - I'm hoping you're right and I can use one wayne_county_commission spider class.

novellac:
This spider will fail CI checks until the mixins folder is committed, I think.

novellac:
Ready for review

pjsier:
Replaced by #459

PR #438, opened 2018-07-11
WIP - 437 wayne_audit first pass - additional logic for postponed meetings

novellac:
This spider is much the same as the spiders for #419 and #435, but has additional logic to cover manually entered "postponed" meetings. We may want to use this spider for #419 and #435, as well.


novellac:
Reworked this spider to work with the mixin wayne_commission. CI checks will fail till the mixin is merged, I think.

pjsier:
It looks like this was included in #441. Is this alright to close?

pjsier:
Closing since this was included in #441

Issue #439, opened 2018-07-13
Create "mixin" classes for agency spiders to inherit


novellac:
**Motivation:** In some cases, multiple agencies may be able to use the same spider code (albeit sometimes with minimal differences). Creating so many similar spiders may create unnecessary bloat, and maintaining so many similar spiders increases the chance of user error.

**Solution:** Allow agency spiders to pull from "mixin" classes.

**Proposed Actions:**
 - Create a "mixins" folder which would live at the same level as the current "spiders" folder.
 - Create mixin classes which pull in and override the spider baseclass.
 - Continue to create one spider for each agency, but allow these spiders to pull in the new Mixin class instead of the baseclass, if appropriate.

pjsier:
We wouldn't need to override the base class here, since the mixin would effectively just be swapping out certain functions, and we could even refer to methods that aren't defined on the mixin itself. 

So if we created this mixin:

```python
class ExampleMixin:
    def _parse_start(self, item):
        return self._parse_dates(item.date())
```

and had a spider inherit from both it and the base class spider

```python
class TestSpider(ExampleMixin, Spider):
    def parse(self, item):
        return {
            'start': self._parse_start(item),
            'end': self._parse_dates(item)
        }
    
    def _parse_dates(self, item):
        return item.date()
```

it would still work, even though `ExampleMixin` doesn't have a `_parse_dates` method defined on its own. This way, we can just implement the functions that are different for any spiders that share code, but the shared code is more transparent when you import multiple mixins at the same level rather than creating a new subclass each time. Django has a good example of this for their [generic editing views](https://github.com/django/django/blob/master/django/views/generic/edit.py) where some views are defined with a few mixins and no code in the body

pjsier:
Closed by #441

PR #441, opened 2018-07-15
WIP - 439 issue mixins classes

novellac:
Looking for feedback on the creation of mixins folder. Relates to issue #439


pjsier:
This looks great to me! Just ran it locally and it seems to work. There might be some linting errors, but otherwise is it ready?

novellac:
Hah yeah those E501 are the bane of my existence sometimes! Should I get those shorter, or is it ok?

pjsier:
Actually, I was wrong. The lint check is passing because we're ignoring E501, so this should be good to go

bonfirefan:
Tests out locally for me! Merging in.

Issue #444, opened 2018-07-16
Wayne Commission on Government Operations


novellac:
New spider for Wayne County Commission on Government Operations (wayne_government_operations) from
https://www.waynecounty.com/elected/commission/government-operations.aspx 

This spider is currently part of ticketing group 13 on the [Public Meetings Aggregator](https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297).

pjsier:
Closed by #459

PR #445, opened 2018-07-16
444 spider wayne government operations

novellac:
First pass at wayne_government_operations spider. Works with mixin wayne_commission. References issue #444


pjsier:
Replaced by  #459

PR #447, opened 2018-07-16
446 spider wayne public safety

novellac:
First pass at wayne_public_safety spider. Works with mixin wayne_commission. References issue #446


novellac:
Ready for review.

novellac:
Thanks for the note I removed them, can you give another look?

pjsier:
I just took a look at #445, and it looks like you might have branched each of the PR branches off of each other, since that one seems to have both government operations and economic development. Could you remove the economic development spider from that one as well? In the meantime I can take a look at this one.

Also, if it's alright with you, I'm gonna do a squash and merge on these so that each PR ends up as a single commit on the master branch. That way, it's easier to see the changes over time if someone needs to go back in the history

city_scrapers/spiders/wayne_public_safety.py
```diff
@@ -0,0 +1,64 @@
...
+    @staticmethod
+    def _parse_description(response):
+        """
+        Event description taken from static text at top of page.
+        """
+        desc_xpath = '//h2[contains(text(), "Public Safety, Judiciary")]/following-sibling::div/section/p/text()'
+        desc = response.xpath(desc_xpath).extract_first()
+        return desc
+
+    @staticmethod
```
pjsier: If the location is hard-coded I think it's fine to include it as a static dictionary without implementing the `_parse_location` method the same way you did with `end`.

pjsier:
Once the location dictionary is updated this should be good to merge, thanks!

pjsier:
I took a look at the other two PRs, and I think the code could be streamlined by making the static location a property on the mixin and then using both that and a `meeting_name` property in the `parse` method. That way, all spiders using the Wayne commission mixin have to do is add a `meeting_name` property and override the `_parse_description` method.

If that works for you I think the easiest way of handling it in the PRs is to merge the three Wayne commission branches into one branch and make the mixin update there, since there won't be many differences between them. Let me know what you think

pjsier:
Replaced by #459

PR #453, opened 2018-07-17
Create spider for Regional Transit Authority of Southeast Michigan

mkrump:


pjsier:
This looks good! Making a note here that I'm not sure this would be considered a Detroit agency, but that's not something that needs to be addressed in this PR

Issue #461, opened 2018-07-19
Refactor Wayne commission spiders


pjsier:
Refactor `wayne_cow`, `wayne_audit`, and `wayne_ways_means` to use the Wayne commission mixin similarly to the spiders in #459. @novellac is taking this one on

novellac:
@pjsier Refactored the relevant spiders and pushed to 457. Ready for review.

pjsier:
Closed by #462

PR #462, opened 2018-07-19
457 wayne commission aggregate

novellac:
Refactor wayne_audit, wayne_ways_means, and wayne_cow to use wayne_commission mixin, and not have long_name. References #461


pjsier:
Looks great, squashing and merging now

Issue #471, opened 2018-07-26
Detroit Land Bank Authority


novellac:
New spider for Detroit Land Bank Authority (det_land_bank) from
https://buildingdetroit.org/events/meetings

This spider is currently part of ticketing group 4 on the [Public Meetings Aggregator](https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297).

pjsier:
Closed by #473

PR #473, opened 2018-07-26
471 det_land_bank first pass

novellac:
Work in progress for det_land_bank
References #471


novellac:
Ok I think this is ready for review.

pjsier:
This looks good to me. One thing we haven't been super strict about that could be helpful here is using consistent camel-casing in class names, making this `DetLandBankSpider`. We'll need to fix that in existing spiders and our template, but it would be nice to have that in new spiders. Otherwise I think this is ready to merge

novellac:
I have just been creating the spiders as per https://github.com/City-Bureau/city-scrapers/blob/master/docs/03_contribute.md

novellac:
Renamed to spider and all spidery things to use camel case.

PR #470, opened 2018-07-25
WIP - 468 spider det city planning

novellac:
First pass at Detroit City Planning Commission

Issue #468, opened 2018-07-23
Detroit City Planning Commission


novellac:
New spider for Detroit City Planning Commission (det_city_planning) from
http://www.detroitmi.gov/Government/Boards/City-Planning-Commission-Information

This spider is currently part of ticketing group 1 on the [Public Meetings Aggregator](https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297).

novellac:
Needs labels:

 - location: Detroit
 - architecture: spiders
 - new spider needed

pjsier:
Closed by #480

PR #480, opened 2018-07-30
WIP - 468 Clean branch with det_city_planning spider

novellac:
This PR references issue #468. This PR should replace #470


pjsier:
This website is a bit odd since there are agendas for dates that aren't listed in the meeting section. Would it make sense to first pull dates from the agendas section and then add any dates from the meetings section that aren't included in there?

Also, since it gives 8:00pm as the approximate time, I think it's safe to include that in the `end` dict as long as the none is included still

tests/test_det_city_planning.py
```diff
@@ -0,0 +1,72 @@
...
+from datetime import date, time
+
+
+from tests.utils import file_response
+from city_scrapers.spiders.det_city_planning import DetCityPlanningSpider
+
+
+test_response = file_response('files/det_city_planning.html', url='https://www.detroitmi.gov/Government/Boards/City-Planning-Commission-Meetings')
+spider = DetCityPlanningSpider()
+parsed_items = [item for item in spider.parse(test_response) if isinstance(item, dict)]
```
mkrump: Here it might be a good idea to sort these based on date, similar to https://github.com/City-Bureau/city-scrapers/pull/481/files#diff-e8f00c59a44be8fe98cff4fc44aaef24R112. Since the spider the meetings go into a dictionary before iterating over them (i.e. can't guarantee order).

city_scrapers/spiders/det_city_planning.py
```diff
@@ -0,0 +1,99 @@
...
+        return meetings
+
+    def _create_documents(self, base_url, url):
+        """
+        Parse or generate documents.
+        """
+        url = urljoin(base_url, url) if url else ''
+        if url:
+            return [{'url': url, 'note': 'Agenda'}]
+        return []
```
mkrump: Coulde probably  refactor to 

```
def _create_documents(self, base_url, url):
    """
    Parse or generate documents.
    """
    if url:
        return [{'url': urljoin(base_url, url), 'note': 'Agenda'}]
    return []
```

novellac: Ok that makes sense thanks.

tests/test_det_city_planning.py
novellac: Done.

novellac:
@pjsier 
Re: the agendas section - I created two dictionaries (has/no agendas) and compared them for clarity.
Re: end times - I'm not sure I understand what you mean by using 8pm as well as using none for the end dict.

pjsier:
@novellac the dictionary approach makes sense to me, and that was actually a typo. I meant to say as long as the note is still included about it ending around 8pm. So it should be safe to set the end date and time at the same day at 8pm

novellac:
Ok, added end time, kept note.

PR #481, opened 2018-07-30
Create Brownfield Redevelopment Authority spider

mkrump:


pjsier:
This looks good, only feedback is that it should be `det_brownfield_redevelopment_authority`

PR #485, opened 2018-07-31
Update `chi_animal` to new event schema

myersjustinc:
Addresses #380.

Paired with @laurieskelly on this at SRCCON.

Moving the discussion here so we don't clutter up the main thread in #380.

### DIFF
--- a/city_scrapers/spiders/chi_animal.py
+++ b/city_scrapers/spiders/chi_animal.py
@@ -39,10 +39,12 @@ class Chi_animalSpider(Spider):
             data = {
                 '_type': 'event',
                 'name': self._parse_name(text),
-                'description': self._parse_description(text),
+                'event_description': self._parse_description(text),
                 'classification': self._parse_classification(text),
-                'start_time': self._parse_start(text),
-                'end_time': self._parse_end(text),
+                'start': {
+                    'date': self._parse_start(text),
+                },
+                'end': {},
                 'all_day': self._parse_all_day(text),
                 'timezone': 'America/Chicago',
                 'status': self._parse_status(text),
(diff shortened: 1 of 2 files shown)


myersjustinc:
@r-wei Following up on our conversation in #380. This now should line up with the new schema; of course, let me know if there's anything else you see that needs tweaking.

r-wei:
thanks @myersjustinc ! I'm usually able to reserve time on Friday mornings to look at PRs, but we're busy this week, so I'll take a look this weekend.

myersjustinc:
No problem. I'll be traveling for several days starting Saturday, so I'll be slower to respond than usual. Have a good weekend!

r-wei:
Hi @myersjustinc , looks good. 

Could you make these few changes:
- split the start datetime object into separate date and time objects (example [here](https://github.com/City-Bureau/city-scrapers/blob/master/docs/06_event_schema.md))
- add an end date and time that is 3 hours after the start time along with a "note" in the "end" dictionary that says "estimated 3 hours after the start time" (this is our convention for sites that don't list an end time) 
- move the timezone field from the scraped data to a class variable like [this](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_community_development.py#L13)
- rename the "long_name" class variable to "agency_id"
- from the [commisson's website](https://www.cityofchicago.org/city/en/depts/cacc/supp_info/public_notice.html), add urls for the "Commission Rules for Public Participation" and "Answers to Frequently Asked Question at the Commission Meetings" to the documents field

Let me know if you have any questions!

myersjustinc:
@r-wei Sure, that all sounds doable. I'll be on vacation til early next week, but I'll work on this when I get back.

r-wei:
looks good, thanks @myersjustinc !

Issue #484, opened 2018-07-31
Tests failing locally for a number of det/michigan tests due to ordering issues


bonfirefan:
Working with @mkrump on this - if anyone else can replicate this failure, this occurs when running tests locally. Occurs with `pytest -s tests` as well. Appears to be due to items being referenced from unordered dictionaries .

When the list of parsed items are ordered, this resolves the error. 
```
=================================== FAILURES ===================================
_______________________________ test_prev_start ________________________________

    def test_prev_start():
>       assert parsed_prev_items[0]['start'] == {
            'date': date(2017, 6, 28), 'time': None, 'note': ''
        }
E       AssertionError: assert {'date': date... 'time': None} == {'date': datet... 'time': None}
E         Omitting 2 identical items, use -vv to show
E         Differing items:
E         {'date': datetime.date(2016, 9, 14)} != {'date': datetime.date(2017, 6, 28)}
E         Use -v to get the full diff

tests/test_det_downtown_development_authority.py:132: AssertionError
_________________________________ test_prev_id _________________________________

    def test_prev_id():
>       assert parsed_prev_items[0]['id'] == \
               'det_downtown_development_authority/201706280000/x/board_of_directors'
E       AssertionError: assert 'det_downtown..._of_directors' == 'det_downtown_..._of_directors'
E         - det_downtown_development_authority/201609140000/x/board_of_directors
E         ?                                        ^^^^
E         + det_downtown_development_authority/201706280000/x/board_of_directors
E         ?                                       ++ ^^

tests/test_det_downtown_development_authority.py:144: AssertionError
_____________________________ test_prev_documents ______________________________

    def test_prev_documents():
>       assert parsed_prev_items[0]['documents'] == [
            {
                'url': 'http://www.degc.org/wp-content/uploads/2017-06-28-dda-special-board-meeting-materials.pdf',
                'note': 'meeting materials',
            },
            {
                'url': 'http://www.degc.org/wp-content/uploads/2017-06-28-Attachment-D.pdf',
                'note': 'attachment d',
            },
            {
                'url': 'http://www.degc.org/wp-content/uploads/2017-06-28-Attachment-K.pdf',
                'note': 'attachment k',
            },
            {
                'url': 'http://www.degc.org/wp-content/uploads/2017-06-28-DDA-Board-Meeting-Minutes.pdf',
                'note': 'minutes',
            },
        ]
[...]

PR #486, opened 2018-07-31
Sort test items so tests are consistent across environments

mkrump:
@bonfirefan I sorted the test cases for the 4 files that I would expect this issue to affect. Try this pr when you get a minute and see if this resolves the issue that you were having locally #484


mkrump:
Going to merge and assume that this fixes #484. Feel free to reopen the issue if it doesn't.

Issue #478, opened 2018-07-30
Belle Isle Advisory Committee


novellac:
New spider for Belle Isle Advisory Committee (mi_belle_isle) from https://www.michigan.gov/dnr/0,4570,7-350-79137_79763_79901---,00.html

This spider does not have a ticketing group (it was part of Group 2 but that needs to be broken up since different/better sources have been found for some of these spiders).

novellac:
Needs labels:

 - location: Detroit
 - architecture: spiders
 - new spider needed

pjsier:
Closed by #488

PR #488, opened 2018-07-31
478 spider mi belle isle

novellac:
References issue #478 
Ready for reviews!

### DIFF
--- a/city_scrapers/spiders/mi_belle_isle.py
+++ b/city_scrapers/spiders/mi_belle_isle.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+import re
+from datetime import datetime
+from dateutil.parser import parse as dateparse
+import scrapy
+from city_scrapers.spider import Spider
+
+
+class MiBelleIsleSpider(Spider):
+    name = 'mi_belle_isle'
+    agency_id = 'Belle Isle Advisory Committee'
+    timezone = 'America/Detroit'
+    allowed_domains = ['www.michigan.gov']
+    start_urls = ["https://www.michigan.gov/dnr/0,4570,7-350-79137_79763_79901---,00.html"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` a dict that follows the Event Schema
+        <https://city-bureau.github.io/city-scrapers/06_event_schema.html>.
+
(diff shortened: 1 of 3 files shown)


pjsier:
This looks great! Do you think you'd be able to capture the minutes in `_parse_documents` as well using something similar to `_match_documents`? It would be fine to ignore any meeting minutes that don't have a corresponding meeting in the schedule

city_scrapers/spiders/mi_belle_isle.py
```diff
@@ -0,0 +1,108 @@
...
+    def _match_documents(self, item, response):
+        """
+        Match up the documents with the date to which they belong
+        """
+        meeting_dates = self._parse_date_and_times(item)
+        meeting_agendas = self._parse_documents(response)
+
+        if meeting_dates[0] in meeting_agendas:
+            return [{'url': self.allowed_domains[0]+meeting_agendas[meeting_dates[0]], 'note': 'Agenda'}]
+        return []
```
mkrump: Probably doesn't matter, since not paging in this spider, but instead of `+`ing allowed domains with url could do what you did `det_city_planning` spider and use `urljoin` + `response.url`. Also, you could unpack the tuple on 103 to make it a bit clearer what `meeting_dates[0]` is. 

```
def _match_documents(self, item, response):
    """
    Match up the documents with the date to which they belong
    """
    meeting_date, *_ = self._parse_date_and_times(item)
    meeting_agendas = self._parse_documents(response)

    if meeting_date in meeting_agendas:
        agenda_url = meeting_agendas[meeting_date]
        return [{'url': urljoin(response.url, agenda_url), 'note': 'Agenda'}]
    return []
```

novellac:
@pjsier Regarding capturing meeting minutes, do you want me to add a minutes dict to the documents array (so eventually there would be two documents for each item)?

novellac:
Oh, ok! If possible, can this be added to the CB schema doc (or the contrib doc?)? I haven't been trying to collect minutes so far, I only thought we were to do agendas.

pjsier:
I think the agenda in the event schema is just an example, but the `documents` array is for any documents on the page. Sorry for the confusion! That could probably be made more explicit in the docs

novellac:
Minutes added, please let me know if this is closer to what you had in mind.

Issue #492, opened 2018-08-05
Detroit Public Library Commission


novellac:
New spider for Detroit Public Library Commission (det_library_commission) from https://detroitpubliclibrary.org/about/commission

This spider is part of ticket group 5.

novellac:
Needs labels:

 - location: Detroit
 - architecture: spiders
 - new spider needed

pjsier:
Closed by #493

PR #493, opened 2018-08-05
492 spider - det_library_commission

novellac:
First pass at Library Commission spider. References #492

### DIFF
--- a/city_scrapers/spiders/det_library_commission.py
+++ b/city_scrapers/spiders/det_library_commission.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+import urllib.parse
+from dateutil.parser import parse
+import scrapy
+from city_scrapers.spider import Spider
+
+
+class DetLibraryCommissionSpider(Spider):
+    name = 'det_library_commission'
+    agency_id = 'Detroit Public Library'
+    timezone = 'America/Detroit'
+    allowed_domains = ['detroitpubliclibrary.org']
+    start_urls = ['https://detroitpubliclibrary.org/about/commission']
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` a dict that follows the Event Schema
+        <https://city-bureau.github.io/city-scrapers/06_event_schema.html>.
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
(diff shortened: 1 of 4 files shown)


city_scrapers/spiders/det_library_commission.py
```diff
@@ -0,0 +1,96 @@
...
+        yield data
+
+    @staticmethod
+    def _parse_name(response):
+        name_xpath = '//header/h1/text()'
+        name = response.xpath(name_xpath).extract_first()
+        return name
+
+    @staticmethod
+    def _parse_classification(meeting_name):
```
pjsier: I think you can safely remove this and just always return Commission since the page limits the meetings to regular commission meetings already

city_scrapers/spiders/det_library_commission.py
```diff
@@ -0,0 +1,96 @@
...
+                'note': '',
+            },
+            'end': {
+                'date': date,
+                'time': end_time,
+                'note': '',
+            },
+            'all_day': False,
+            'location': location,
+            'documents': [],
```
pjsier: It would be great to pull the minutes here and create the old meetings from the (limited) details in the minutes if you think it's doable. It's not a problem if you don't think there's enough to go off of there though

city_scrapers/spiders/det_library_commission.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @staticmethod
+    def _parse_date_and_times(response):
+        date_text = response.xpath('//tr[1]/td/text()').extract_first()
+        time_text = response.xpath('//tr[2]/td/text()').extract_first()
+        start_time, end_time = time_text.split(" - ")
+        date_value = parse(date_text)
+        start_value = parse(start_time)
+        end_value = parse(end_time)
+
```
mkrump: If you wanted you could maybe combine time w/ date text and return the start_date_time, end_date_time. Not sure if that's actually cleaner though.

```
start_date_time = parse(date_text + '  ' + start_time)
end_date_time = parse(date_text + '  ' + end_time)
```

novellac: I kind of liked it better with the three variables, because I wanted to keep it simpler up in the _parse_item method.

Issue #496, opened 2018-08-06
Police and Fire Retirement System


novellac:
New spider for Detroit Police and Fire Retirement System (det_police_fire_retirement) from http://www.pfrsdetroit.org/Resources/Meetings

This spider is part of ticketing group 7 on the Public Meetings spreadsheet (https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297)

PR #497, opened 2018-08-06
496 spider det police fire retirement

novellac:
Ready for review. Draws on #489 (det_general_retirement_system). References #496.


novellac:
If all good, then ready for merge.

Issue #503, opened 2018-08-08
Detroit City Council


novellac:
New spider for Detroit City Council (det_city_council) from 
http://www.detroitmi.gov/Government/City-Council/City-Council-Sessions

These spiders need to be assigned their own ticketing group. They currently take up rows 21-28 on the spiders spreadsheet: https://docs.google.com/spreadsheets/d/1Uu9FmcQAy8AXhBR2bj84LbSdlfgMNNroa9P89rsDJSQ/edit#gid=1214169297

novellac:
Needs labels:

 - location: Detroit
 - architecture: spiders
 - new spider needed

pjsier:
Closed by #508

PR #508, opened 2018-08-13
Create Detroit City Council spider

mkrump:


novellac:
References issue #503

PR #512, opened 2018-09-06
Write feed output to Azure

pjsier:
Initial setup for writing feed output to Azure Blob Storage instead of S3


pjsier:
Merging since failures are unrelated

PR #518, opened 2018-09-15
Add Chicago Mayor's Bicycle Advisory Council spider 🚲

hancush:
Hello!

This PR adds a spider and tests for the Mayor's Bicycle Advisory Council spider (#418), scraped from [this link](http://chicagocompletestreets.org/getinvolved/mayors-advisory-councils/). 

I have a couple of questions:

- The default `_parse_status` returns "tentative" if there is no agenda. The meeting info available is pretty sparse, so that essentially becomes the default here, if the meeting hasn't passed. Since the information is so sparse, should I leave it as is, or override the base method to return confirmed?
- When I ran the tests locally, I got a failure for a different spider. Should I fix that in this PR?

```
========================================================== FAILURES ===========================================================
_________________________________________________________ test_status _________________________________________________________

    def test_status():
>       assert parsed_items[0]['status'] == CONFIRMED
E       AssertionError: assert 'passed' == 'confirmed'
E         - passed
E         + confirmed

tests/test_cook_landbank.py:59: AssertionError
```


hancush:
Ooh, I just realized there's [an archive link with agendas](http://chicagocompletestreets.org/getinvolved/mayors-advisory-councils/mbac-meeting-archives/)! I'll need to roll that in.

pjsier:
Thanks for this! Adding the agendas sounds great, but if that becomes an issue at all returning "tentative" should be fine since there isn't a lot of info.

The `cook_landbank` error looks like it's because I didn't add `freezegun` after updating that spider and its fixtures in #517, so it's generating the status based on the current date. You're welcome to fix it in this PR if you'd like, but this should be good to merge without it after the agenda update

hancush:
Thanks, @pjsier! I'll get on that this evening.

pjsier:
No problem, thanks for the updates @hancush! Scraping the archive page sense sounds good to me, and I think we've seen similar setups on other scrapers. Merging now

PR #527, opened 2018-10-10
Update documentation

pjsier:


bonfirefan:
This looks really good!! Wow. So everything is at https://cityscrapers.org/ now?

pjsier:
@bonfirefan yep! Still a work in progress, but we added a lot of context and kept the existing technical docs

Issue #540, opened 2018-10-24
Rewrite Detroit Wrecking Contractors Board scraper


pjsier:
The city of Detroit recently updated its website, so we'll need to rewrite this scraper. We may be able to generalize the scraper behavior similarly to the [Wayne Commission mixin](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/mixins/wayne_commission.py)

pjsier:
Closed by #542

PR #542, opened 2018-10-26
539 - Rewrite det_wrecking_examiners

novellac:
Looks as if city of detroit moved from Dot Net Nuke to Drupal!

This wasn't able to use wayne_mixin.
Mentions #540


pjsier:
Looks great!

Issue #578, opened 2018-11-13
Spider: Allegheny County Council


bonfirefan:
Allegheny County: This spider uses Legistar, meaning we will use the Legistar API. Please template it using the [cook_board spider](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/cook_board.py). 

spider name: alle_city
https://alleghenycounty.legistar.com/Calendar.aspx

Covers the following:

<ul class="rcbList" style="list-style:none;margin:0;padding:0;zoom:1;"><li class="rcbItem">Committee on Appointment Review</li><li class="rcbItem">Committee on Budget and Finance</li><li class="rcbItem">Committee on Economic Development &amp; Housing</li><li class="rcbItem">Committee on Government Reform</li><li class="rcbItem">Committee on Health &amp; Human Services</li><li class="rcbItem">Committee on Human Services</li><li class="rcbItem">Committee on Parks</li><li class="rcbItem">Committee on Public Safety</li><li class="rcbItem">Committee on Public Works</li><li class="rcbItem">County Council</li><li class="rcbItem">County Council - Special Meeting</li><li class="rcbItem">Executive Committee</li><li class="rcbItem">Public Hearing</li><li class="rcbItem">Special Committee on Education</li><li class="rcbItem">Special Committee on Elections</li><li class="rcbItem">Special Committee on Government Reform</li><li class="rcbItem">Special Committee on Insurance</li><li class="rcbItem">Special Committee on Marketing</li><li class="rcbItem">Special Committee on Parks and Public Works</li><li class="rcbItem">Special Committee on Property Assessments</li><li class="rcbItem">Special Committee on Reapportionment</li><li class="rcbItem">Special Committee on the Councils of Governments</li><li class="rcbItem inactiveComboItem" style="color:Gray;">Committee on Campaign Finance Reform (inactive)</li></ul>

bonfirefan:
@thoo is working on this

dpimpinella:
Just to avoid any confusion, @thoo and @bonfirefan, I mistakenly referenced this issue in my PR for another issue, so please ignore that. I have not done any work on this specific issue. Thanks!

thoo:
@dpimpinella np. It happens all the time. :)

thoo:
I don't see csv file at `local_outputs`. I can see all the parsed text in CLI.

pjsier:
[...]

Issue #576, opened 2018-11-09
Spider: Housing Authority of the City of Pittsburgh


bonfirefan:
Housing Authority of the City of Pittsburgh
hacp.org, http://www.hacp.org/public-information/board-info
Spider name: pitt_housing

This is particularly relevant given the [housing summit this weekend in Pittsburgh.](housingsummitpgh.org)

pjsier:
Closed by #579

bonfirefan:
It looks like they have changed their URL to http://www.hacp.org/public-information/board-info
The http://www.hacp.org/public-information/board-info URL now throws a 404 error. Does someone want to modify the existing pitt_housing spider that @dpimpinella created? @dpimpinella you get first dibs, of course.

dpimpinella:
@bonfirefan I'll take care of this one!

dpimpinella:
@bonfirefan I won't be able to complete this within a reasonable time frame, so i'm putting it back up for grabs.

pjsier:
Moved to https://github.com/bonfirefan/city-scrapers-pitt/issues/7

PR #579, opened 2018-11-13
#578 Add pitt_housing spider (WIP)

dpimpinella:
This pull request addresses issue #576 , containing a spider that crawls the Housing Authority of the City of Pittsburgh Board of Commisioners meetings page.

Marked as WIP since this is my first PR, but both the spider and test file are complete (and hopefully correct). Let me know if you need me to change anything with the files, format of this PR etc. Thanks!

### DIFF
--- a/city_scrapers/spiders/pitt_housing.py
+++ b/city_scrapers/spiders/pitt_housing.py
@@ -38,7 +38,7 @@ class PittHousingSpider(Spider):
                 'end': {'date': None, 'time': None, 'note': None},
                 'all_day': False,
                 'location': {
-                    'address': '',
+                    'address': 'Civic Building, 200 Ross St., Pittsburgh, PA, 15219',
                     'name': '',
                     'neighborhood': ''
                 },
--- a/city_scrapers/spiders/pitt_housing.py
+++ b/city_scrapers/spiders/pitt_housing.py
@@ -35,9 +35,9 @@ class PittHousingSpider(Spider):
                 'event_description': 'Board of Commissioners meeting',
                 'classification': BOARD,
                 'start': self._parse_start(item),
-                'end': None,
+                'end': '',
                 'all_day': False,
-                'location': None,
+                'location': '',
                 'documents': self._parse_documents(item),
                 'sources': self._parse_sources(item),
             }
--- a/city_scrapers/spiders/pitt_housing.py
+++ b/city_scrapers/spiders/pitt_housing.py
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
 import re
 import scrapy
-import dateutil.parser 
+from datetime import date
 from html.parser import HTMLParser
+
+from dateutil.parser import parse as dtparse
+
 from city_scrapers.spider import Spider
+from city_scrapers.constants import *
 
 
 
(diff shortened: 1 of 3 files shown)


dpimpinella:
Seems that I need to change 'end' and 'location' from None to empty strings to match the event schema. I'll make these changes on my fork, but will hold off on any more commits here, in case I need to make some other changes based on feedback.

city_scrapers/spiders/pitt_housing.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
+
+        for item in self._create_meetings(response): 
+
+            data = {
+                '_type': 'event',
+                'name': 'Housing Authority of the City of Pittsburgh',
```
pjsier: This isn't currently documented, but we want to avoid using the agency name as the meeting name, so in this case `Board of Commissioners` would work as the name, and you can leave the description blank since they don't provide a unique one for each meeting

city_scrapers/spiders/pitt_housing.py
```diff
@@ -0,0 +1,145 @@
...
+    def _parse_documents(self, item):
+        """
+        Parse or generate documents.
+        """
+
+        # Create documents, which will be a list of dictionaries corresponding 
+        # to each agenda or minutes document found.
+        documents = []
+        for doc in item[1]: 
+            if re.search(r'Agenda',doc):
```
pjsier: Since you aren't using a regular expression here, something like `agenda in doc.lower()` might be more clear. You should be able to remove `re` in this file generally since it looks like it's only used for string lookups

city_scrapers/spiders/pitt_housing.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return documents
+
+    def _parse_sources(self, item):
+        """
+        Parse or generate sources.
+        """
+        return [{'url': self.start_urls[0], 'note': ''}]
+
+class MyHTMLParser(HTMLParser):
```
pjsier: Could you explain the purpose of this a bit? I'm not sure this needs to be split into a parse subclass, and it makes things a bit harder to maintain when there's a lot of variations across scrapers. If there is a good use case it might be worth seeing if this can be generalized for multiple scrapers (that doesn't need to be in this PR though)

tests/test_pitt_housing.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+from datetime import date, time
+
+from tests.utils import file_response
+from city_scrapers.constants import *
+from city_scrapers.spiders.pitt_housing import PittHousingSpider
+
+
+# def test_tests():
```
pjsier: This and the "Uncomment below" comment can be removed

city_scrapers/spiders/pitt_housing.py
dpimpinella: Sure thing.  Most meeting data were links and were formatted like this whenever scrapy grabbed their markup:

<p><a href="http://static1.firemandev.info.s3.amazonaws.com/galleries/general/2018%20Board%20Information/Board_Minutes_-_January_2018.pdf">Jan. 25, 2018 -- Minutes</a></p>

But those that were not links were formatted differently:

<p>July 31, 2018 -- Minutes</p>
 <p><strong>August 2018 -- Board does not meeting in August.\xa0</strong></p>

There was still quite a bit of HTML parsing to be done, so I used HTMLParser (which seemed like it can only be used through subclassing) to grab the links and dates, taking into account the differences in markup.  I was having trouble with keeping the events in chronological order, so this was the solution I came up with to grab everything we need in one pass.

pjsier: Makes sense! It doesn't add a huge amount of complexity, so if it works in this case I think that's fine

pjsier:
Thanks for making these changes! Merging

city_scrapers/spiders/pitt_housing.py
dpimpinella: Great, thanks for reviewing this!

PR #596, opened 2018-11-19
Alle league

### LINKED ISSUE #590: Spider: Allegheny League of Municipalities
URL: http://alleghenyleague.org/
Spider name: alle_league_of_municipalities
Agency name: Allegheny League of Municipalities

thoo:
Close #590

### DIFF
--- a/city_scrapers/spiders/pitt_allegheny_league.py
+++ b/city_scrapers/spiders/pitt_allegheny_league.py
@@ -2,8 +2,9 @@
 import os
 import unicodedata
 from datetime import datetime
-from city_scrapers.spider import Spider
+
 from city_scrapers.constants import BOARD, COMMITTEE, NOT_CLASSIFIED
+from city_scrapers.spider import Spider
 
 
 class PittAlleghenyLeagueSpider(Spider):
--- a/city_scrapers/spiders/pitt_allegheny_league.py
+++ b/city_scrapers/spiders/pitt_allegheny_league.py
@@ -71,7 +71,7 @@ class PittAlleghenyLeagueSpider(Spider):
         """
         start, _ = item[2].split('-')
         datetime_obj = datetime.strptime(start.strip(),
-                                 '%B %d, %Y @ %I:%M %p')
+                                         '%B %d, %Y @ %I:%M %p')
         return {
             'date': datetime_obj.date(),
             'time': datetime_obj.time(),
@@ -84,8 +84,8 @@ class PittAlleghenyLeagueSpider(Spider):
         """
         _, end = item[2].split('-')
         datetime_obj = datetime.strptime(('{date} {time}'
-                                 .format(date=item[0], time=end)),
-                                 '%B %d, %Y %I:%M %p')
+                                          .format(date=item[0], time=end)),
+                                         '%B %d, %Y %I:%M %p')
         return {
             'date': datetime_obj.date(),
             'time': datetime_obj.time(),
--- a/city_scrapers/spiders/pitt_allegheny_league.py
+++ b/city_scrapers/spiders/pitt_allegheny_league.py
@@ -1,0 +1,132 @@
+# -*- coding: utf-8 -*-
+import os
+import unicodedata
+from datetime import datetime
+from city_scrapers.spider import Spider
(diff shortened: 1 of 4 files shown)


city_scrapers/spiders/pitt_allegheny_league.py
```diff
@@ -0,0 +1,133 @@
...
+from city_scrapers.constants import BOARD, COMMITTEE, NOT_CLASSIFIED
+from city_scrapers.spider import Spider
+
+
+class PittAlleghenyLeagueSpider(Spider):
+    name = 'pitt_allegheny_league'
+    agency_name = 'Allegheny League of Municipalities'
+    timezone = 'America/Chicago'
+    allowed_domains = ['alleghenyleague.org']
+    start_urls = ['http://alleghenyleague.org/']
```
pjsier: It looks like the full calendar of events is on this page http://alleghenyleague.org/calendar-of-events/ so we should probably scrape from there instead of the home page so that we can get more than the first 3 items. We could also go one month into the future, and you can see [`chi_low_income_housing_trust_fund`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_low_income_housing_trust_fund.py) for an example of this

city_scrapers/spiders/pitt_allegheny_league.py
```diff
@@ -0,0 +1,133 @@
...
+            'address': item[-1],
+            'name': '',
+            'neighborhood': '',
+        }
+
+    def _parse_documents(self, item):
+        """
+        Parse or generate documents.
+        """
+        return [{'url': item, 'note': ''}]
```
pjsier: Since there can be multiple documents with different information, we should always have a note for documents. In this case, it looks like it's just giving a link to Google Calendar, so there aren't any documents available and this can be dropped. `sources` is a bit confusing since it's really just the link the page was scraped from which is why we're allowing there to be an empty note

city_scrapers/spiders/pitt_allegheny_league.py
thoo: Thanks for the link: [`chi_low_income_housing_trust_fund`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_low_income_housing_trust_fund.py).  [The page](http://alleghenyleague.org/calendar-of-events/) I am scraping has a click-button to go to the next month but it looks like a javascript button without a url-link. So the url would be the same for both the current and next month. Is it still possible to scrape info from  the next month?
 https://github.com/City-Bureau/city-scrapers/blob/be48a646b75891c0e0c4f4e9ff605bf92353fd21/city_scrapers/spiders/chi_low_income_housing_trust_fund.py#L41-L47
This one looks like it is looking for a href. In my case, I think ``the action to click`` is required. Any suggestion?? I think I can invoke `onclick` in `Selenium`.

pjsier: It looks like you should be able to submit a POST request to http://alleghenyleague.org/wp-admin/admin-ajax.php. If you send these parameters it should return JSON with HTML of the table in the `data` property:

```
action: simcal_default_calendar_draw_grid
id: 417
month: (month integer)
year: (year integer)
```

Selenium is a bit of a heavy dependency, and submitting a POST request should be able to get you the same information

thoo: @pjsier  Btw, I was trying to find this link: http://alleghenyleague.org/wp-admin/admin-ajax.php at http://alleghenyleague.org/calendar-of-events/ using `Inspect` but could not find it under Network tag. Am I looking the wrong page or wrong location? Thanks.

pjsier:
@thoo it might make sense to hold off on this one for now, commented on #590, but not sure this is an open public meeting

thoo:
Ah I see. It looks like these meeting are for the board members. I will check with Bonnie.

pjsier:
@thoo We moved all Pittsburgh/Allegheny County scrapers to a separate repo, so if you're still interested in this one feel free to open it as a PR there! https://github.com/bonfirefan/city-scrapers-pitt/

PR #605, opened 2018-11-21
WIP

### LINKED ISSUE #600: Spider: Port Authority of Allegheny County
URL: https://www.portauthority.org/paac/CompanyInfoProjects/BoardofDirectors/MeetingAgendasResolutions.aspx
Spider name: alle_port_authority
Agency name: Port Authority of Allegheny County

thoo:
Close #600.
[URL-link](https://www.portauthority.org/paac/CompanyInfoProjects/BoardofDirectors/MeetingAgendasResolutions.aspx)
Spider name: *alle_port_authority*
Agency name: *Port Authority of Allegheny County*

### DIFF
--- a/city_scrapers/spiders/alle_port_authority.py
+++ b/city_scrapers/spiders/alle_port_authority.py
@@ -1,0 +1,193 @@
+# -*- coding: utf-8 -*-
+from city_scrapers.spider import Spider
+from lxml import html
+import unicodedata
+from datetime import datetime, timedelta
+from dateutil.parser import parse
+from city_scrapers.constants import BOARD, COMMITTEE
+
+class AllePortAuthoritySpider(Spider):
+    name = 'alle_port_authority'
+    agency_name = 'Port Authority of Allegheny County'
+    timezone = 'America/New_York'
+    allowed_domains = ['www.portauthority.org']
+    start_urls = ['https://www.portauthority.org/paac/CompanyInfoProjects/BoardofDirectors/MeetingAgendasResolutions.aspx']
+    event_year = str(datetime.now().year)
+
+    def _build_datatable(self,response):
+        alist_tbody = (response
+                 .xpath('//table[1]/tbody//td')
+                 .extract()
+                )
+
+        atable=[]
+        arow=[]
+
+        for item in alist_tbody:
+            tree = html.fragment_fromstring(item)
+            #pdb.set_trace()
+            text = tree.text_content()
+
+            url = tree.xpath('//a/@href')
+            find_att_b = tree.xpath('//b/text()|//strong/text()')
+            if len(find_att_b)>=1:
+                continue;
+            if url:
+                arow.append('{text}: {url}'
+                            .format(text=text, url=url[0]))
+            else:
+                arow.append('{text}'.format(text=unicodedata.normalize("NFKD", text)))
+            if len(arow) == 6:
+                atable.append(arow)
+                arow = []
+        #import pdb;pdb.set_trace()
+        return atable
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` a dict that follows a modified
+        OCD event schema (docs/_docs/05-development.md#event-schema)
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
+        atable = self._build_datatable(response)
+        #import pdb;pdb.set_trace()
+        for row in atable:
+            data = {
+                'timezone': self.timezone,
+                '_type': 'event',
+                'name': self._parse_name(row),
+                'event_description': self._parse_description(row),
+                'classification': self._parse_classification(row),
+                'start': self._parse_start(row),
+                'end': self._parse_end(row),
+                'all_day': self._parse_all_day(row),
+                'location': self._parse_location(row),
+                'documents': self._parse_documents(row),
+                'sources': self._parse_sources(row),
+            }
+
+            data['status'] = self._generate_status(data)
+            data['id'] = self._generate_id(data)
+
+            yield data
+
+        # self._parse_next(response) yields more responses to parse if necessary.
+        # uncomment to find a "next" url
+        # yield self._parse_next(response)
+
+
+    def _parse_name(self, item):
+        """
+        Parse or generate event name.
+        """
+        return item[0]
+
+    def _parse_description(self, item):
+        """
+        Parse or generate event description.
+        """
+        return ''
+
+    def _parse_classification(self, item):
+        """
+        Differentiate board meetings from public hearings.
+        """
+        meeting_title = item[0].lower()
+        if 'committee' in meeting_title:
+            return COMMITTEE
+        return BOARD
+
+    def _parse_start_datetime(self, item):
+        """
+        Parse the start datetime.
+        """
+        if 'cancel' in item[2].lower():
+            return ''
+
+        if not item[1].strip():
+            if 'stakeholder' in item[0].lower():
+                time = '8:30 a.m.'
+            if 'performance oversight' in item[0].lower():
+                time = '9:00 a.m.'
+            else:
+                time = '9:30 a.m.'
+        else:
+             time = item[1]
+        date = ('{year} {date}'
+                .format(year=self.event_year, date=item[2]))
+
+        time_string = '{0} {1}'.format(date, time)
+        return (parse(time_string))
+
+    def _parse_start(self, item):
+        datetime_obj = self._parse_start_datetime(item)
+        if not datetime_obj:
+            return ''
+        return {
+            'date': datetime_obj.date(),
+            'time': datetime_obj.time(),
+            'note': ''
+        }
+
+    def _parse_end(self, item):
+        """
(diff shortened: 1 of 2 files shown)


city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class AllePortAuthoritySpider(Spider):
+    name = 'alle_port_authority'
+    agency_name = 'Port Authority of Allegheny County'
+    timezone = 'America/New_York'
+    allowed_domains = ['www.portauthority.org']
+    start_urls = ['https://www.portauthority.org/paac/CompanyInfoProjects/'
+                  'BoardofDirectors/MeetingAgendasResolutions.aspx']
+    event_year = str(datetime.now().year)
```
pjsier: The `str` call here is unnecessary since you're using it in the `format` function later

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+                       .xpath('//table[1]/tbody//td')
+                       .extract()
+                       )
+
+        atable = []
+        arow = []
+
+        for item in alist_tbody:
+            tree = html.fragment_fromstring(item)
+            # pdb.set_trace()
```
pjsier: Looks like a leftover comment that can be removed

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+            text = tree.text_content()
+
+            url = tree.xpath('//a/@href')
+            find_att_b = tree.xpath('//b/text()|//strong/text()')
+            if len(find_att_b) >= 1:
+                continue
+            if url:
+                arow.append('{text}: {url}'
```
pjsier: A dictionary with `name` and `url` keys might be more clear here? A dictionary might be a better fit for the rows more generally since you have a sense of what's in each column

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+        `parse` should always `yield` a dict that follows a modified
+        OCD event schema (docs/_docs/05-development.md#event-schema)
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
+        atable = self._build_datatable(response)
+        for row in atable:
+            data = {
+                'timezone': self.timezone,
```
pjsier: `timezone` isn't necessary to include here. Let me know if you see anything that says otherwise in the docs and I can update that

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+
+            yield data
+
+    def _parse_name(self, item):
+        """
+        Parse or generate event name.
+        """
+        return item[0]
+
+    def _parse_description(self, item):
```
pjsier: You can remove this function since it always returns a static value and just include `event_description: ''` in the item dictionary

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+        meeting_title = item[0].lower()
+        if 'committee' in meeting_title:
+            return COMMITTEE
+        return BOARD
+
+    def _parse_start_datetime(self, item):
+        """
+        Parse the start datetime.
+        """
+        if 'cancel' in item[2].lower():
```
pjsier: From looking at some archives (http://www.portauthority.org/paac/CompanyInfoProjects/BoardofDirectors/2016PublicMeetingArchive.aspx) it seems like they'll leave the time in even if a meeting is canceled sometimes. We should still check and see if the time is available, and use the default meeting time if not

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+        if not datetime_obj:
+            return ''
+        return {
+            'date':  datetime_obj.date(),
+            'time': ((datetime_obj + timedelta(hours=3))
+                     .time()),
+            'note': 'Estimated 3 hours after start time'
+        }
+
+    def _parse_all_day(self, item):
```
pjsier: This can also be removed like `_parse_description`

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Parse or generate all-day status. Defaults to False.
+        """
+        return False
+
+    def _parse_location(self, item):
```
pjsier: This works, but it might be good to check to make sure a part of this text is in the page (maybe just the ZIP and room?) and throw an error if we can't find it. That way we'll know if the meeting location ever changes

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        return {
+            'address': ('Neal H. Holmes Board Room, '
+                        '345 Sixth Avenue, Fifth Floor, '
+                        'Pittsburgh, PA 15222-2527'),
+            'name': '',
+            'neighborhood': '',
+        }
+
+    def _parse_documents(self, item):
```
pjsier: We should also collection Resolutions if they're available here

city_scrapers/spiders/alle_port_authority.py
thoo: My script won't work on any other year except 2018 because the table format keeps changing every or every other year.  I can still check for the time but the date would be missing in year 2018. Would it still be useful???

pjsier: In that case I think it's fine to focus on 2018, thanks for checking!

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,179 @@
...
+
+        for row in atable:
+            data = {
+                '_type': 'event',
+                'name': self._parse_name(row),
+                'event_description': '',
+                'classification': self._parse_classification(row),
+                'start': self._parse_start(row),
+                'end': self._parse_end(row),
+                'all_day': '',
```
pjsier: `all_day` should return a boolean, so `False` in this case

city_scrapers/spiders/alle_port_authority.py
pjsier: Just to clarify here, this works, but we shouldn't return any meeting that doesn't have a date or time

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,179 @@
...
+        for item in alist_tbody:
+            tree = html.fragment_fromstring(item)
+            text = tree.text_content()
+
+            url = tree.xpath('//a/@href')
+            find_att_b = tree.xpath('//b/text()|//strong/text()')
+            if len(find_att_b) >= 1:
+                continue
+            if url:
+                arow.append('{name}: {url}'.format(name=text, url=url[0]))
```
pjsier: We should still use a dictionary instead of a formatted string here to keep things more clear

city_scrapers/settings/base.py
```diff
@@ -19,7 +19,7 @@
 USER_AGENT = 'City Scrapers [development mode]. Learn more and say hello at https://www.citybureau.org/city-scrapers/'
 
 # Obey robots.txt rules
-ROBOTSTXT_OBEY = True
+ROBOTSTXT_OBEY = False
```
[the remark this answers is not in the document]
thoo: Thanks.

city_scrapers/spiders/alle_port_authority.py
thoo: This is just for internal usage. I am building a 2D list from a table from HTML. The final url will be a dictionary. It is just regenerating a html table as a 2D list.

thoo:
I think this will not record the meeting without date. 
 https://github.com/thoo/city-scrapers/blob/1fec8fc1d0735c3580ffcfe6bdacc8cf4a461d47/city_scrapers/spiders/alle_port_authority.py#L81-L82

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -78,6 +78,8 @@ def parse(self, response):
             data['status'] = self._generate_status(data)
             data['id'] = self._generate_id(data)
 
+            if not (data['start'] and data['end']):
```
thoo: I think this will remove the meeting without datetime.

city_scrapers/spiders/alle_port_authority.py
```diff
@@ -0,0 +1,182 @@
...
+            documents.append({'note': 'Minutes', 'url': details.split(' ')[-1]})
+
+        agenda = item[3]
+        if agenda.startswith('Agenda: http'):
+            documents.append({'note': 'Agenda', 'url': agenda.split(' ')[-1]})
+
+        resolution = item[4]
+        if resolution.startswith('Resolutions: http'):
+            documents.append({'note': 'Resolution', 'url': resolution.split(' ')[-1]})
+
```
thoo: The final data are in dict.

PR #608, opened 2018-11-26
chi_board_elections spider

zarifmahmud:
I created a spider for the Chicago Board of Elections, but I'm not sure if I'm on the right track with how it's supposed to behave. I'm just returning the dates of the meetings right now.


pjsier:
Thanks for the status update! We also want to get the upcoming meeting listed on the site, so [`det_economic_development_corporation`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/det_economic_development_corporation.py) might be a good example of scraping both upcoming and past meetings on separate pages.

For the past meetings, it looks like most meetings start at 9:30am (with some exceptions at 10am) unless 7 am/7 pm is listed so it should be safe to default the start time to 9:30am unless otherwise noted.

For the upcoming meetings you should be able to use a regular expression to find the date and time inside the description. The [`_parse_start` method of `chi_ssa_1`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_ssa_1.py#L50) is a simple example of how to do that. Hope that helps!

zarifmahmud:
Hey, the spider is tentatively finished, and I've also finished the test file, which were tested on some HTML version of the meetings pages. The only thing left to do is set the environment variables, but I couldn't find the details of what's required on the city-scrapers docs.

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` a dict that follows a modified
+        OCD event schema (docs/_docs/05-development.md#event-schema)
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
+        if response.url == \
```
pjsier: This conditional could be simplified into something like

```python
if 'minutes' in response.url:
    ...
else:
    ...
```

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                "https://app.chicagoelections.com/pages/en/meeting-minutes-and-videos.aspx":
+            yield from self._prev_meetings(response)
+        if response.url == \
+                "https://app.chicagoelections.com/pages/en/board-meetings.aspx":
+            yield from self._next_meeting(response)
+
+        # self._parse_next(response) yields more responses to parse if necessary.
```
pjsier: Since you aren't using this you can remove it

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+        # uncomment to find a "next" url
+        # yield self._parse_next(response)
+
+    def _next_meeting(self, response):
+        next_meeting = response.xpath('//div[@class="copy"]/text()').extract()[2]
+        meetingdate = re.search(r'\d{1}:.{24}', next_meeting).group(0)
```
pjsier: The regex here is a bit too specific, since one character being added or removed will break it. Try something like `\d{1,2}:.*\d{4}` to look for 1 or 2 digits followed by a colon (for the start of the time) ending with a four digit number for the year

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+        data = {
+            '_type': 'event',
+            'name': "Chicago Board of Election Commissioners",
+            'event_description': "Meeting",
```
pjsier: You can leave `event_description` blank since one isn't provided, and `name` should be "Electoral Board". More info here https://cityscrapers.org/docs/development/#name

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+            'classification': COMMISSION,
+            'start': self._parse_start(meetingdate),
+            'end': {},
```
pjsier: This should include the date, since we know it isn't multi-day, but the time can be blank. Here's an example https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_ssa_1.py#L41

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+            'all_day': False,
+            'location': self._parse_location(response),
+            'documents': [],
```
pjsier: The meeting notice and meeting agenda can be scraped here, and we have a good amount of examples of this in different scrapers

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+            'sources': [{
+                'url': "https://app.chicagoelections.com/pages/en/board-meetings.aspx",
```
pjsier: you can use `response.url` as a shortcut for the `url` here instead of hard-coding it

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                'note': ''
+            }],
+        }
+
+        data['status'] = self._generate_status(data)
+        data['id'] = self._generate_id(data)
+        yield data
+
+    def _prev_meetings(self, response):
+        meetings = response.xpath("//a/text()").extract()
```
pjsier: We're going to want to pull the actual `a` element here so we can get the documents. You'll also need to check that you only add each date once, since the minutes and video documents both list the same day

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+        meetingdates = [
```
pjsier: Not sure if I'm following what's going on here, but splitting on - or – might be safer

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+            meeting[22:]
+            for meeting in meetings
+            if "Minutes" not in meeting and "mode" not in meeting and meeting is not " "
+        ]
+        for meetingdate in meetingdates:
+            meetingdate = "9:30 a.m. on " + meetingdate
+            data = {
+                '_type': 'event',
+                'name': "Chicago Board of Election Commissioners",
```
pjsier: Same comment as above regarding the name and description

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                'event_description': "Meeting",
+                'classification': COMMISSION,
+                'start': self._parse_start(meetingdate),
+                'end': {},
```
pjsier: Same comment for `end` as above

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                'all_day': False,
+                'location': self._parse_location(response),
+                'documents': [],
```
pjsier: You'll want to scrape the minutes, video, and any other document for each meeting

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                'sources': [{
+                    'url':
```
pjsier: Same comment as above about `response.url`

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+                    'note': ''
+                }],
+            }
+
+            data['status'] = self._generate_status(data)
+            data['id'] = self._generate_id(data)
+
+            yield data
+
+    def _parse_next(self, response):
```
pjsier: You can delete this since it isn't being used

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Get next page. You must add logic to `next_url` and
+        return a scrapy request.
+        """
+        next_url = None  # What is next URL?
+        return scrapy.Request(next_url, callback=self.parse)
+
+    def _parse_name(self, item):
```
pjsier: This, `_parse_description`, `_parse_end`, `_parse_all_day, and `_parse_sources` can be removed since they aren't being used

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        return ''
+
+    def _parse_all_day(self, item):
+        """
+        Parse or generate all-day status. Defaults to False.
+        """
+        return False
+
+    def _parse_location(self, item):
```
pjsier: This works, but we should add a sanity check here that looks for `69 W Washington` (more flexible than that though) and returns an error if it isn't in the description. That way we'll know when/if the location changes. This likely only applies to `_next_meeting` though

tests/test_chi_board_elections.py
```diff
@@ -0,0 +1,137 @@
+from datetime import date, time
+
+import pytest
+from tests.utils import file_response
+
+from city_scrapers.constants import COMMISSION
+from city_scrapers.spiders.chi_board_elections import ChiBoardElectionsSpider
+
+test_response = file_response('files/chi_board_elections.html')
```
pjsier: `freezegun` should be added here so that the test results don't depend on when it runs. We've got examples of this in some of the other scraper test files

zarifmahmud: Should I run freezegun twice, since I'm looking at two pages?

pjsier: You can just call the `stop()` method after the second spider fixture is created, that way it's running for both

zarifmahmud: Oh okay, I updated the test file

city_scrapers/spiders/chi_board_elections.py
zarifmahmud: I'm just going to scrape the Regular Board Meeting Agenda and not the Electoral Board Meeting Agenda or Electoral Board Meeting Notice here, because I think those two are out of date (for Sept. 27 rather than Nov 27)

zarifmahmud:
Hey, sorry for taking a bit of time with this one. After running some more tests a few days ago, I realized that I wasn't getting back all the dates, so I've been working on the regex for the past few days when I have the time, to get all the dates, take care of weird unicode characters, and also handle the few exception dates. The code's more complicated now, but it should return back all of the meetings now, along with their relevant documents.

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,208 @@
...
+                            'url': response.url,
+                            'note': ''
+                        }],
+                    }
+                    data['end'] = {
+                        'date': data['start']['date'],
+                        'time': None,
+                        'note': '',
+                    }
+                    if "Minutes" in meeting:
```
pjsier: This can be moved into the `_parse_documents` method, and we should pull the links to the videos as well. You should be able to generalize it by checking if a given node is a link tag, and then using the first part of the string before the dash (so "Board Meeting Minutes - ..." becomes "Board Meeting Minutes). You could do that with something like

```python
meeting_str.split(' - ')[0]
```

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,208 @@
...
+                        minuteslink = re.search(r'="(.+)"', meeting).group(1)
+                        data["documents"] = self._parse_documents(minuteslink)
+                    data['status'] = self._generate_status(data)
+                    data['id'] = self._generate_id(data)
+                    yield data
+                if not self._different_time(meeting):  # To handle the odd 7 AM/7PM meetings
+                    prevdate = meetingdate
+                if "Nov. 13" in meetingdate:  # To handle the one date that falls outside of scraper
```
pjsier: Good catch on the behavior here. We'll probably want something more general though. We have an example of iterating through text nodes (anything not directly contained by a tag) in an older version of [`chi_teacherpension`](https://github.com/City-Bureau/city-scrapers/blob/f18c0e9297b6756686e7ca7c74668c104dbe9219/city_scrapers/spiders/chi_teacherpension.py#L36)

zarifmahmud: The spider brings back video links now, but I when I tried to iterate through the text nodes, by using text() and span in XPATH, I got the text back, but no links, so I wasn't sure how to fix that

pjsier: Gotcha, yeah this one is a bit tricky. I got it to work locally with this:

```python
meetings = response.xpath("//a|//span/text()").extract()
for meeting in meetings:
  if not meeting.strip():
    continue # ignores any of the empty strings that are returned from /text()
  ...
```

Let me know if that helps! You might have to change up some other pieces that are expecting t to be a link and not text, and you'll need to remove the " - To be considered..." text

zarifmahmud:
I've updated the files, so it returns video links now, and it iterates through the text nodes. The only thing I couldn't figure out what how to get rid of this HTML space character "&#160;",  which causes some dates to be processed twice, because the program sees them as a date.

pjsier:
Thanks for this update! Could you share an example of where you're running into the non-breaking space issue so I can take a look?

zarifmahmud:
Yeah, for sure. Right when the prev_meetings function is processing a meeting, it checks the next meeting on the list, and sees if it has the same date. If it does, then it parses it for another document, as it probably has a Youtube link or minutes agenda. But some dates like April 24, 2018 are written as both "April 24, 2018" and "April&#160;24, 2018". This is also the case for April 10, 2018. I'd ran a find and replace for the "\xa0" character, but it wasn't working for this character.

pjsier:
In that case it sounds like the best option is comparing the date objects that are the output of `_parse_start` instead of the strings. You could just assign the output of `_parse_start` to `meetingdate` and set `start` to `meetingdate`. The parsed results should match

zarifmahmud:
Hey, I forgot to comment yesterday, but the program can handle those non-breaking space characters now

city_scrapers/spiders/chi_board_elections.py
```diff
@@ -0,0 +1,199 @@
...
+            except AttributeError:  # Sometimes meetings will return None
+                continue
+
+    def _parse_name(self, item):
+        """
+        Parse or generate event name.
+        """
+        return ''
+
+    def _parse_classification(self, item):
```
pjsier: This and `_parse_name` can be deleted

tests/test_chi_board_elections.py
```diff
@@ -0,0 +1,162 @@
...
+def test_classification_prev(item):
+    assert item['classification'] is COMMISSION
+
+
+@pytest.mark.parametrize('item', parsed_items_prev)
+def test__type_prev(item):
+    assert parsed_items_prev[0]['_type'] == 'event'
+
+
+# def test_start_full():
```
pjsier: This can be remove if it isn't being used

zarifmahmud:
Awesome, I've pushed those last changes

pjsier:
Great, thanks for sticking with this! We've got a lot of really good information from this scraper that will be especially relevant with the upcoming election. Doing a squash and merge now to keep the git history clear

zarifmahmud:
Thanks for all the help in making this too. It was my first time using Regex and XPATH, so all the advice you gave me was super helpful

PR #616, opened 2018-12-08
0560 spider chi ssa 21

### LINKED ISSUE #560: Spider: Chicago Special Service Area #21 Lincoln Square Ravenswood
URL: http://www.lincolnsquare.org/SSA-no-21-Commission-meetings
Spider Name: `chi_ssa_21`
Agency Name: Chicago Special Service Area #21 Lincoln Square Ravenswood

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

MrDiggles2:
Closes #560 

Adds completed spider chi_ssa_21 with tests.

Please let me know if you have any questions.


city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
+# -*- coding: utf-8 -*-
+from city_scrapers.spider import Spider
+from city_scrapers.constants import COMMISSION
+import dateutil.parser
+
+
+class ChiSsa21Spider(Spider):
+
+    name = 'chi_ssa_21'
+    agency_name = 'Lincoln Square Neighborhood Improvement Program'
```
pjsier: The agency would be "Chicago Special Service Area #21 Lincoln Square Ravenswood". SSAs get complicated when they have the nonprofit administrator, so we're just going with the SSA name here

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+        <https://city-bureau.github.io/city-scrapers/06_event_schema.html>.
+
+        Change the `_parse_id`, `_parse_name`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.xpath('//div[@id="content-327081"]/p'):
+
+            data = {
+                '_type': 'event',
+                'name': self._parse_name(item),
```
pjsier: This is also a bit confusing for SSAs, but it seems like it makes the most sense to use "Lincoln Square Neighborhood Improvement Program" here. We mainly want to avoid repeating information from the agency name, and in other cases we've said "SSA Governing Commission", but it looks like they're using that name here.

You can also get rid of the `_parse_name` method and just return the static string here since it doesn't change

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+        Parse or generate event description.
+        """
+        description = 'N/A'
+
+        # The itinerary of the meeting is always stored in the <ul>
+        # element immediately following
+        detailElement = item.xpath('following-sibling::*[1]')
+        name = detailElement.xpath('name()').extract_first()
+
+        if (name == 'ul'):
```
pjsier: The parsing here is great! The output is a little confusing though. If we could split each segment with newlines and then the topic and detail with a colon so that:

"2019 Budget Adjustments[Review and Approval]:PaintWorks[Artist Selection]"

becomes:

2019 Budget Adjustments: Review and Approval
PaintWorks: Artist Selection

that would be ideal

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+                    ]
+                )),
+                detailElement.xpath('li')
+            ))
+
+            description = ':'.join(topics)
+
+        return description
+
+    def _parse_classification(self, item):
```
pjsier: Similar to `_parse_name`, this method can be removed since it's returning a static value

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+            'note': 'estimated 2 hours after start time'
+        }
+
+        return ret
+
+    def _parse_date(self, item):
+        rawDate = item.xpath('strong/text()').extract_first()
+        return dateutil.parser.parse(rawDate)
+
+    def _parse_all_day(self, item):
```
pjsier: This can also be removed

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+        location = ''.join(
+            item.xpath('em//text()').extract()
+        ).strip()
+
+        if not location:
+            location = defaultLocation
+
+        return {
+            'address': location,
+            'name': '',
```
pjsier: When possible, we should split the name "Bistro Campagne" from the address and include it here. It looks like they're relatively consistently split by a comma, so we could check for that above.

Also, since we can assume this will always be in Chicago, we should add "Chicago, IL" to the end of the address if it isn't already present. That helps the accuracy of our geocoding later on

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if url:
+            return [{'url': url, 'note': 'Meeting notes'}]
+        return []
+
+    def _parse_sources(self, item):
+        """
+        Parse or generate sources.
+        """
+        return [{'url': '', 'note': ''}]
```
pjsier: You can remove this method and just return `[{'url': response.url, 'note': ''}]` in the `parse` method above

city_scrapers/spiders/chi_ssa_21.py
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+    def _parse_documents(self, item):
+        """
+        Parse or generate documents.
+        """
+        url = item.xpath('a/@href').extract_first()
+
+        if url:
+            return [{'url': url, 'note': 'Meeting notes'}]
```
pjsier: "Minutes" is probably a more fitting description here

MrDiggles2:
@pjsier - Thanks for the quick review! I updated the PR with your suggested changes.

city_scrapers/settings/base.py
```diff
@@ -27,7 +27,7 @@
 # Configure item pipelines
 ITEM_PIPELINES = {
     'city_scrapers.pipelines.CityScrapersItemPipeline': 200,
-    # 'city_scrapers.pipelines.CsvPipeline': 400,
+    'city_scrapers.pipelines.CsvPipeline': 400,
```
pjsier: Looks like this snuck in there, but if you re-add the comment it's good

city_scrapers/spiders/chi_ssa_21.py
pjsier: Thanks for making those changes! Just two more small items. If you could default the description to an empty string instead of "N/A" and include a space after the colon so that "2019 Budget Adjustments:Review and Approval" becomes "2019 Budget Adjustments: Review and Approval" this looks good

pjsier:
Thanks, this looks great! Merging now

PR #702, opened 2019-03-10
WIP: Clarify the comment in 04-development.md about hardcoding values

kiwisquash:
Per discussion in https://github.com/City-Bureau/city-scrapers/pull/697#discussion_r261839741, I am suggesting a small change in the [Development doc](https://cityscrapers.org/docs/development/#a-write-parse-methods-in-the-spider).

### DIFF
--- a/docs/_docs/04-development.md
+++ b/docs/_docs/04-development.md
@@ -317,7 +317,9 @@ class ChiHousingSpider(Spider):
     # ...
 ```
 
-Often a value for meetings returned by a spider will be the same regardless of meeting content (an example is that most meetings will always have `False` for the `all_day` value). In this case, feel free to remove the `_parse_*` method for that field, and simply include the value in each dictionary (so `'all_day': False` in this example rather than `'all_day': self._parse_all_day(item)`).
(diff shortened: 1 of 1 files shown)


docs/_docs/04-development.md
```diff
@@ -317,7 +317,9 @@ class ChiHousingSpider(Spider):
     # ...
 ```
 
-Often a value for meetings returned by a spider will be the same regardless of meeting content (an example is that most meetings will always have `False` for the `all_day` value). In this case, feel free to remove the `_parse_*` method for that field, and simply include the value in each dictionary (so `'all_day': False` in this example rather than `'all_day': self._parse_all_day(item)`).
+Often a value for meetings returned by a spider will be the same regardless of meeting content (an example is that most meetings will always have `False` for the `all_day_value`). For fields like `classification`, `all_day`, and `title` (sometimes), feel free to remove the `_parse_*` method for that field, and simply include the value in each dictionary (so `'all_day': False` in this example rather than `'all_day': self._parse_all_day(item)`).
```
pjsier: Small thing here, but I'd change "`all_day_value`" to "`all_day` value"

kiwisquash: Oops! Good catch.

kiwisquash: Fixed in the next commit.

PR #705, opened 2019-03-17
WIP: 0677 spider cook medical examiner

Anphisa:
I finished a first version of the spider (@pjsier). Some of the scraping feels too specific to me, specifically how time and location are extracted (lines 27-33).
I'm sure a lot of the scraping could be done more elegantly. Please let me know if something's not the way it should be!
Tests and linting run fine for me.

### DIFF
--- a/city_scrapers/spiders/cook_medical_examiner.py
+++ b/city_scrapers/spiders/cook_medical_examiner.py
@@ -1,0 +1,115 @@
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from city_scrapers.constants import ADVISORY_COMMITTEE
+
+import re
+import datetime
+
+
+class CookMedicalExaminerSpider(CityScrapersSpider):
+    name = "cook_medical_examiner"
+    agency = "Cook County Medical Examiner's Advisory Committee"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cookcountyil.gov"]
+    start_urls = ["https://www.cookcountyil.gov/service/medical-examiners-advisory-committee"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        # Venue and address are in a separate field above the meeting data
+        #venue = response.css(".views-field-field-location-title .field-content *::text").get()
+        #address = response.css(".views-field-field-address *::text").getall()
+
+        title = response.css("#page-title *::text").get().strip().replace("\\", "")
+        meetings = response.css(".field-items strong *::text").getall()
+        dates = []
+        for i, item in enumerate(meetings):
+            print(i, item)
+            if "Time" in item:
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/cook_medical_examiner.py
```diff
@@ -0,0 +1,100 @@
...
+                times = re.findall("(1[0-2]|0?[1-9]):([0-5]\\d)\\s*([AaPp][Mm])", item)
+                start_time = times[0]
+                end_time = times[1]
+            if "Location" in item:
+                venue = item.split(": ")[1]
+                street = meetings[i + 1] + "; " + meetings[i + 2]
+                address = {
+                    "name": venue,
+                    "address": street.replace(u'\xa0', u' '),
+                    "coordinates": {
```
pjsier: `coordinates` was part of an older version of the items returned, so it can be removed here

city_scrapers/spiders/cook_medical_examiner.py
```diff
@@ -0,0 +1,100 @@
...
+        return dt
+
+    def _parse_end(self, date, end_time):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        if "Date" in date:
+            return
+        dt = datetime.datetime.strptime(date + str(end_time), "%A, %B %d, %Y('%I', '%M', '%p')")
+        return dt
+
+    def _parse_links(self, item):
```
pjsier: If there aren't any links this should return an empty list instead of a list with one empty item. This site is an interesting case because the detail page is the agenda for the coming meeting.

I'd lean toward leaving this blank because the content of the page changes over time and each meeting doesn't get a unique URL so it would be obsolete after each new meeting, but if I'm missing something let me know!

city_scrapers/spiders/cook_medical_examiner.py
Anphisa: Okay!

city_scrapers/spiders/cook_medical_examiner.py
Anphisa: I can't find the detail page! :D Do you mean the agenda list? I have included that as a meeting description for the first meeting. Or is there a link to another agenda?

pjsier: I just meant the page you're currently scraping is the detail page, and I think your solution of including the agenda as the description works best, so returning an empty list here should be good

Anphisa: Alright, thank you for your quick answers! I will leave it as it is and push the new version with the corrections.

pjsier: Sounds great! It looks like you’ve addressed all my feedback here though, by corrections do you just mean fixing the errors in the CI build? Looks like that’s just a style issue with yapf

Anphisa: Yes, hopefully I've now fixed the styling issue as well :)

pjsier:
These changes look great, thanks @Anphisa! Merging now

Anphisa:
Thanks for your fast help, @pjsier! This was fun! :)

Issue #719, opened 2019-04-15
Spider: Illinois Opioid Crisis Response Advisory Council


pjsier:
URL: https://www.dhs.state.il.us/page.aspx?item=97186
Links URL: https://www.dhs.state.il.us/page.aspx?item=104682
Spider Name: `il_opioid_response`
Agency Name: Illinois Opioid Crisis Response Advisory Council

It's important that we pull the URLs here, and it's safe to assume that the time and location are constant

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

channiemills:
Hi @pjsier - I'd like to contribute and will take a look at this issue.

pjsier:
@channiemills sounds great!

pjsier:
Closed by #724

PR #724, opened 2019-04-17
0719 spider il opioid crisis ac

channiemills:
Implements `il_opioid_response` for https://github.com/City-Bureau/city-scrapers/issues/719

I'm ready for a review on this. I'm pulling the meeting agenda and links from the [links URL]( https://www.dhs.state.il.us/page.aspx?item=104682). https://github.com/City-Bureau/city-scrapers/issues/719 mentions [this page](https://www.dhs.state.il.us/page.aspx?item=97186) as the primary URL, which appears to just list the most recent agenda. Let me know if there's anything additional I should be pulling from that page or any feedback in general.

### DIFF
--- a/city_scrapers/spiders/il_opioid_response.py
+++ b/city_scrapers/spiders/il_opioid_response.py
@@ -85,13 +85,15 @@ class IlOpioidResponseSpider(CityScrapersSpider):
         """Parse or generate location."""
         return [
             {
-                "address": "401 S. Clinton Street, 7th Floor Executive Conference Room,"
-                " Chicago, IL 60607",
+                "address":
+                    "401 S. Clinton Street, 7th Floor Executive Conference Room,"
+                    " Chicago, IL 60607",
                 "name": "Illinois Department of Human Services Clinton Building",
             },
             {
-                "address": "100 S. Grand Avenue East, 3rd Floor Executive Video Conference Room,"
-                " Springfield, IL 62762",
+                "address":
+                    "100 S. Grand Avenue East, 3rd Floor Executive Video Conference Room,"
+                    " Springfield, IL 62762",
                 "name": "Illinois Department of Human Services Harris Building",
             },
         ]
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/il_opioid_response.py
```diff
@@ -0,0 +1,144 @@
...
+                links=self._parse_links(item, response),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
```
pjsier: This should have a default like “Advisory Council”, or a unique meeting name if it has one (doesn’t seem like this does though)

city_scrapers/spiders/il_opioid_response.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return [
+            {
+                "address":
```
pjsier: Right now we’re only supporting one location as a dictionary for meetings, and in this case I think we can just use the Chicago one.

city_scrapers/spiders/il_opioid_response.py
```diff
@@ -0,0 +1,144 @@
...
+                    " Springfield, IL 62762",
+                "name": "Illinois Department of Human Services Harris Building",
+            },
+        ]
+
+    def _parse_links(self, item, response):
+        """Parse or generate links for meeting agenda and minutes."""
+        links = []
+        agenda = {
+            "type": "agenda",
```
pjsier: The type attribute can be removed here, since we only include title and href for links. Was it mentioned anywhere in the docs? We’ve changed the setup a bit so if it’s mentioned somewhere I’m happy to remove it to reduce confusion

channiemills: I added the `type` key as a way to differentiate between a link to a meeting agenda vs a link to meeting minutes. The page lists both, is there a better way to do this?

pjsier: Gotcha, thanks for clarifying! If you know for sure which links will be what, it’s fine to just make “Agenda” or “Minutes” the title property if the existing  link text isn’t useful

channiemills: Fair point, that can be determined by the link text.

channiemills:
Interesting, it looks like the travis test kicked off after you requested changes and then continues to kick off automatically on additional commits. Thanks for the feedback - ready for review again.

pjsier:
@channiemills these changes look great, thanks for the PR! Merging now

channiemills:
Thank you! Happy to contribute.

PR #731, opened 2019-04-22
0721 spider det transportation

nbearman:
Here's my attempt at a spider for the Detroit Department of Transportation. I used the det_city.py mixin for the spider, and I modeled the spider and tests off of the Detroit Zoning Appeals and Detroit City Planning spiders. However, since the Department of Transportation is under the "Department" filter on the calendar and documents pages (and not the "Government" filter, as is the case with the above examples),  I had to modify det_city.py. With this modification, the mixin can accommodate filtering by a department ID as well; if no such ID is provided by a spider using the Detroit city mixin, the existing behavior should still work.

I believe that the existing Detroit Police Department spider could use this new filtering behavior as well, because it's meetings, too, are found with the "Department" filter. Currently, it looks like the Detroit Police Department spider filters for the "Board of Police Commissioners," which misses many of the Police Department meetings and documents. I used the Police Department spider in addition to my new spider to test the new filtering behavior, but I'm not pushing any modifications to that spider.

I ran into a couple problems which seemed beyond the scope of my spider, but maybe someone can let me know if they are errors on my part.

Firstly, events with "All Day" as the time (https://detroitmi.gov/events/free-fare-day) caused an error with my spider. As a result, the page was not parsed by the spider.

Secondly, some descriptions don't parse right (https://detroitmi.gov/node/17986), returning a blank description. For most events, the default description parsing seemed to work. In the case of the above example, the HTML for the event page appeared to have a description where the default parsing behavior was looking for it, but I thought maybe it was nested to deeply for the parser to find it.


city_scrapers/spiders/det_transportation.py
```diff
@@ -0,0 +1,16 @@
...
+
+class DetTransportationSpider(DetCityMixin, CityScrapersSpider):
+    name = "det_transportation"
+    agency = "Detroit Department of Transportation"
+    dept_cal_id = '166'
+    agency_cal_id = 'All'
+    dept_doc_id = '166'
+    agency_doc_id = 'None'
+
+    def _parse_classification(self, response):
```
pjsier: You can override this to return `FORUM` when it's a Community Input meeting, `ADVISORY_COMMITTEE` when it's an Advisory Council meeting and then it can default to `BOARD`.

city_scrapers/spiders/det_transportation.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class DetTransportationSpider(DetCityMixin, CityScrapersSpider):
+    name = "det_transportation"
+    agency = "Detroit Department of Transportation"
+    dept_cal_id = '166'
+    agency_cal_id = 'All'
+    dept_doc_id = '166'
+    agency_doc_id = 'None'
+
```
pjsier: To avoid the issue with all day events you can override the `parse_event_page` method to check and see if a page has a valid start time and ignore it if it doesn't. It would go something like this

```python
def parse_event_page(self, response):
  try:
    self._parse_start(response)
  except AttributeError:
    return
  return super().parse_event_page(response)
```

nbearman:
I made the changes you requested to classifications; now I'm looking at the title of the meeting, and if it says "input meeting," I'll call that a forum, and similarly with the advisory council. I thought about just using "meeting," but that appeared to capture some events which were not necessarily forums, like this change review meeting (https://detroitmi.gov/events/january-service-changes-final-meeting). I used "input meeting" instead of "community input meeting" because I think leaving the filter as broad as possible will correctly identify more potential events. Thanks for catching this; I wasn't sure from the contribution guide what each classification really meant, so I used "Board," as does the zoning spider.

I made the all-day event fix you suggested, and the error that had caught my attention went away.

Finally, could you elaborate on your suggestion for handling descriptions? Most of my descriptions look good, besides the blank ones, which you said are fine. I wasn't really sure what you meant by looking for irregular white space.

Thanks for your timely feedback! I appreciate the help.

pjsier:
Thanks for the updates! We could definitely clarify the classifications in the docs.

For descriptions, I think I was misunderstanding the issue. I thought you were getting descriptions that were weird combinations of whitespace, but if the issue is just that the description is sometimes blank I think this is good to go. I'll check it again and then merge

PR #740, opened 2019-05-07
[Docs] Change yapf flag to what Travis runs

mattpair:
`yapf --in-place --recursive ./city_scrapers/ ./tests/` caused Travis to fail as seen in the attached screenshot.

Proposing a change in the documentation to what Travis runs (`yapf --diff --recursive ./city_scrapers/ ./tests/`) so a user can make the appropriate changes.

### DIFF
--- a/docs/_docs/04-development.md
+++ b/docs/_docs/04-development.md
@@ -211,7 +211,7 @@ We use [`flake8`](http://flake8.pycqa.org/en/latest/){:target="\_blank"}, [`isor
 ```bash
 $ flake8
 $ isort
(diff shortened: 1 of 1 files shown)


pjsier:
`yapf --in-place` actually fixes the issue, but `yapf --diff` is used in the CI to check whether there are any changes. Locally `--in-place` should still be used, but differences from Travis and what happens locally are usually related to the version of `yapf` being different between the main branch and the forked branch, which it looks like was the case here. Thanks for mentioning this though! Closing since it's a different issue

PR #741, opened 2019-05-08
0729 cook north shore mosquito

mattpair:


city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,98 @@
...
+        month_day = self._clean_html_tags(item.split('–')[0])
+        date = month_day + ' ' + year + ' ' + '7 pm'
+        return datetime.strptime(date, '%B %d %Y %I %p')
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
```
pjsier: It looks like this is pulling whether the meeting is cancelled, but that should probably just be used for the status instead. It doesn't look like there are notes on the time here otherwise

city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,98 @@
...
+        date = item.split('–')
+        if len(date) > 1:
+            date = date[1]
+            if 'View' not in date:
+                return self._clean_html_tags(date)
+        return ""
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
```
pjsier: Since the page includes the address text here, but it can sometimes be hard to format, you can implement a check to make sure it hasn't changed like in [`chi_labor_retirement_fund`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_labor_retirement_fund.py#L28)

Also, since we're hard coding it with checks, it should include the state and ZIP code to make geocoding easier

city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,98 @@
...
+            "name": "NSMAD Office",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = []
+        after_dash = item.split('–')
+        if len(after_dash) > 1:
+            raw_link = after_dash[1].split()
+            if len(raw_link) > 1 and 'a' in raw_link[0]:
```
pjsier: The agenda isn't posted there now, but it looks like they post it so this should check if "agenda" or "minutes" is in the link and use whichever is as the title

city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,98 @@
...
+                minutes_link = re.search(r'\"(.+?)\"', raw_link[1]).group(1)
+                minutes = {"href": minutes_link, "title": "Minutes"}
+                links.append(minutes)
+        return links
+
+    def _parse_source(self, response):
+        """Parse or generate source."""
+        return response.url
+
+    def _clean_html_tags(self, item):
```
pjsier: I didn't mention this on the last one, but it's safer to get the text attributes directly rather than parsing them out with regex. Here's an example of what that looks like in [`chi_municipal_retirement`](https://github.com/City-Bureau/city-scrapers/blob/3201ff0030271a6364ed38c2d16659c11cbe85b1/city_scrapers/spiders/chi_municipal_retirement.py#L50)

city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def _parse_classification(self):
+        """Parse or generate classification from allowed options."""
+        return BOARD
+
+    def _parse_start(self, item, response):
+        """Parse start datetime as a naive datetime object."""
+        year = '2019'
+        calendar_title = response.xpath('//*[@id="post-68"]/div/h2[1]').extract()[0]
+        if '2019' not in calendar_title:
```
pjsier: It looks like the header will presumably always include the year, so we should parse it out directly by looking for a four digit number rather than throwing an error when the year changes

city_scrapers/spiders/cook_north_shore_mosquito.py
mattpair: How do I use this for status instead? I've got the check modified and the right constants imported but how can I only sometimes overwrite what `_get_status` returns?

pjsier: It's not super clear, but you can pass an optional `text` keyword argument to `_get_status` that checks if any form of "cancel" or "rescheduled" is in the string and sets it to `CANCELLED` if there is ([here it is in `city-scrapers-core`](https://github.com/City-Bureau/city-scrapers-core/blob/master/city_scrapers_core/spiders/spider.py#L45)).

Here's an example of it in [`chi_community_development`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_community_development.py#L50), and you can see where "Cancelled" shows up next to the dates being scraped [on the agency page](https://web.archive.org/web/20190512050940/https://www.chicago.gov/city/en/depts/dcd/supp_info/community_developmentcommission.html)

city_scrapers/spiders/cook_north_shore_mosquito.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        self._parse_location(response)
+        calendar = response.xpath('//*[@id="post-68"]/div/ul/li/ul')
+        for index, item in enumerate(calendar.extract()[0].split('<li>')[1:]):
+            print(index)
```
pjsier: Is this ready for review again @mattpair? Other than removing this print statement it looks good to go

PR #765, opened 2019-06-17
fix typo

categulario:
just fixes a typo in the docs


pjsier:
@categulario thanks for the PR!

PR #780, opened 2019-07-03
0627 spider chi ohare noise

GeorgeDubuque:
Problem:
Having trouble yielding meetings after all detail pages for each meeting have been scraped. Because of scrapy's use of Twisted which deals with concurrent network calls, yielding requests causes them to be downloaded and processed concurrently making it difficult to determined when the meeting items are ready to be yielded.

Goal:
Determine when all detail pages have been scraped and data has been processed so I can yield all the meetings at the same time.

Current solution:
yield meetings even if all of the meeting info has not been scraped then yield them again when the rest of the data is scraped

### DIFF
--- a/city_scrapers/settings/base.py
+++ b/city_scrapers/settings/base.py
@@ -24,6 +24,9 @@ ROBOTSTXT_OBEY = True
 # Disable cookies (enabled by default)
 COOKIES_ENABLED = False
 
+LOG_LEVEL='ERROR'
+LOG_ENABLED=True
+
 # Throttle results by default
 AUTOTHROTTLE_ENABLED = True
 AUTOTHROTTLE_START_DELAY = 0.5
--- a/city_scrapers/settings/base.py
+++ b/city_scrapers/settings/base.py
@@ -24,9 +24,6 @@ ROBOTSTXT_OBEY = True
 # Disable cookies (enabled by default)
 COOKIES_ENABLED = False
 
-LOG_LEVEL='ERROR'
-LOG_ENABLED=True
-
 # Throttle results by default
 AUTOTHROTTLE_ENABLED = True
 AUTOTHROTTLE_START_DELAY = 0.5
--- a/city_scrapers/spiders/chi_ohare_noise.py
+++ b/city_scrapers/spiders/chi_ohare_noise.py
@@ -11,52 +11,56 @@ class ChiOhareNoiseSpider(CityScrapersSpider):
     timezone = "America/Chicago"
     allowed_domains = ["www.oharenoise.org"]
     meetings = dict()
-    start_urls = ["https://www.oharenoise.org/about-oncc/agendas-and-minutes",
-                  "https://www.oharenoise.org/about-oncc/oncc-meetings/month.calendar/"]
+    start_urls = ["https://www.oharenoise.org/about-oncc/agendas-and-minutes"]
+    requests = []
 
     def parse(self, response):
         """
         `parse` should always `yield` Meeting items.
         """
-        if "agendas" in response.url:
-            table_rows = response.selector.xpath(".//tbody//tr")
-            for item in table_rows:
-                date = self._parse_start(item)
-                url_date = date.strftime("%Y/%m/%d")
-                meeting = Meeting(
-                    title=self._parse_title(item),
-                    links=self._parse_links(item),
-                    start=self._parse_start(item),
-                    source=self._parse_source(response)
-                )
+        table_rows = response.selector.xpath(".//tbody//tr")
 
-                meeting["status"] = self._get_status(meeting)
-                meeting["id"] = self._get_id(meeting)
-                self.meetings[url_date] = meeting
-        else:
-            curr_date = datetime.date.today()
-            months_urls = [datetime.date(month=curr_date.month-2, day=1, year=curr_date.year),
-                           datetime.date(month=curr_date.month-1, day=1, year=curr_date.year),
-                           curr_date,
-                           datetime.date(month=curr_date.month+1, day=1, year=curr_date.year)
-                           ]
-            for date in months_urls:
-                yield Request(
-                        "https://www.oharenoise.org/about-oncc/oncc-meetings/month.calendar/" +
-                        date.strftime("%Y/%m/%d"), callback=self._parse_calendar)
-            print("done parsing months")
+        for item in table_rows:
+            date = self._parse_start(item)
+            url_date = date.strftime("%Y/%m/%d")
+            meeting = Meeting(
+                title=self._parse_title(item),
+                links=self._parse_links(item),
+                start=self._parse_start(item),
+                source=self._parse_source(response)
+            )
 
-    def _yield_meetings(self):
-        print(len(self.meetings))
-        for date_key, meeting in self.meetings.items():
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
             yield meeting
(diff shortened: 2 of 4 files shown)


pjsier:
@GeorgeDubuque thanks for the PR! Now I've got a clearer idea of your issue. It looks like what you need to do is similar to what's happening in `chi_buildings` here: https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_buildings.py#L57

First you scrape the documents page, then you can add the information for the meeting to the `meta` `kwarg` so that you can reference it between requests and then you pass that along to the detail page function to pull the rest of what you need. Let me know if that's unclear, and I'll make some other comments as well

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,167 @@
...
+                start=self._parse_start(item),
+                source=self._parse_source(response)
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+            yield meeting
+            self.meetings[url_date] = meeting
+        curr_date = datetime.date.today()
+        months_urls = [datetime.date(month=curr_date.month-2, day=1, year=curr_date.year),
```
pjsier: You might want to use `datetime.timedelta` and days here, otherwise this will break on months like January or December

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,167 @@
...
+                       datetime.date(month=curr_date.month-1, day=1, year=curr_date.year),
+                       curr_date,
+                       datetime.date(month=curr_date.month+1, day=1, year=curr_date.year)
+                       ]
+        for date in months_urls:
+            yield Request(
```
pjsier: You might need to add the `dont_filter=True` `kwarg` here as well so that it doesn't do the normal duplicate filtering

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,167 @@
...
+                    "https://www.oharenoise.org/about-oncc/oncc-meetings/month.calendar/" +
+                    date.strftime("%Y/%m/%d"), callback=self._parse_calendar, priority=1)
+
+    def _parse_months(self):
```
pjsier: This seems duplicated from the previous section

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,167 @@
...
+                    title=self._parse_title(name_date_time, is_calen=True),
+                    start=start,
+                    description=description,
+                    location=location
+                    )
+            self.meetings[date_key] = meeting
+        yield meeting
+
+    def _not_empty(self, string):
+        if string is not None and not str.isspace(string):
```
pjsier: Empty strings and `None` evaluate to `False`, so instead of having a function for this you can just use `if string and string.strip():`

city_scrapers/spiders/chi_ohare_noise.py
GeorgeDubuque: Oh wow, I didn't think of that 🤦‍♂

GeorgeDubuque:
So in that solution, I would be yielding meetings after I add details from the detail page? The only issue I have with that is some of the meetings on the documents page will not be on the calendar and therefore won't have a detail page (because they are further back then two months) and some of the meetings on the calendar won't be on the documents page because they are in the future. How would I yield those ones?

pjsier:
Instead of using `request.meta` you could create a dictionary of dates/datetimes that populates meeting details as they come up, and once you start to go through the calendar you can yield requests for everything with a detail page and yield meetings for everything that doesn't show up on the calendar. On the detail page you could pull information from the dictionary of datetimes for the meeting (if available) or populate it with defaults.

It could be something like this in `chi_ssa_29` https://github.com/City-Bureau/city-scrapers/blob/51db93bfc876ed51dc27df651832bce53bf4c6ac/city_scrapers/spiders/chi_ssa_29.py except you would set the dictionary to `self.meetings_map` (or something similar) in the beginning.

GeorgeDubuque:
Hey sorry for the late response. The implementation I have now is similar to your suggestion on the pr. I currently add to the meetings dictionary from the agendas/minutes page (key indexed by the date) and then append the data from the details page to the meetings if they exist in the dictionary already. Would it be ok if I just used the requests library instead of concurrent scrapy requests when scraping the calendars page? This would allow me to know when to yield the meetings

city_scrapers/settings/base.py
```diff
@@ -24,6 +24,9 @@
 # Disable cookies (enabled by default)
 COOKIES_ENABLED = False
 
+LOG_LEVEL='ERROR'
```
pjsier: We'll need this removed from the PR before merging. Fine for your local testing though

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,183 @@
...
+                source=self._parse_source(response)
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+            self.meetings[url_date] = meeting
+
+        self._parse_months()
+        for date, meeting in self.meetings.items():
+            yield meeting
```
pjsier: There are still several missing fields on the `Meeting` objects when they're yielded here. `all_day` and `classification` are two of them, and while `all_day` can be `False`, `_parse_classification` will need to be implemented

GeorgeDubuque: Gotchya will do!

GeorgeDubuque: So I'm a little confused about how to classify each meeting, is there a specific formula to determine which classification meetings should be based on their title?

GeorgeDubuque: For now I'm just going to parse the title for the classification constants and define them that way. Let me know if this is ok. Having fun working on this!

pjsier: We've got a list of the classification constants and what they mean here https://cityscrapers.org/docs/development/#classifications, but usually it lines up with the titles

GeorgeDubuque:
Hey! Just updated the changes you requested. Let me know if you need anything else.

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,215 @@
+from city_scrapers_core import constants
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from scrapy.selector import Selector
+import requests
```
pjsier: Looks like you're getting an error because `isort` is failing, but if you run that locally it should fix it

pjsier:
@GeorgeDubuque it's looking good so far, but you still need to implement the tests. Let me know if you have any questions on that!

GeorgeDubuque:
> @GeorgeDubuque it's looking good so far, but you still need to implement the tests. Let me know if you have any questions on that!

Sounds good Pat, what kind of tests should be implemented for this spider? What I am thinking is, for example, testing whether or not the data for a single meeting is being consolodated properly between the agendas page and the calendar page. Also running a test to count the total number of meetings and make sure it lines up with how many there should be.

pjsier:
@GeorgeDubuque those tests make sense to me in addition to the base ones we have that are currently commented out

pjsier:

PR #781, opened 2019-07-03
[WIP] 0565 spider chi ssa 27 new changes

nmschorr:
I believe I have everything as far as git and the spider itself ready to review again.  I'll get back to the pytest tests after this.


city_scrapers/spiders/chi_ssa_27.py
```diff
@@ -0,0 +1,170 @@
...
+        items, minutes_pdf, meeting_location, meeting = [], "", "", None
+        the_url_source = response.url
+
+        items_list = (response.css(".panel-body").getall()[4]).split("<p>")
+
+        for i in items_list:
+            items.append(re.sub("\n|\r|</p>", "", i))
+
+        for item in items:
+            if re.search(r'<div', item):
```
pjsier: We should try to avoid parsing HTML directly with regular expressions. You can use `' ' .join(item.css('*::text').extract())` to pull out all nested text in an element

city_scrapers/spiders/chi_ssa_27.py
```diff
@@ -0,0 +1,170 @@
...
+                continue  # skip these
+
+            if re.search(r"meetings are held at", item):
```
pjsier: This can just be a substring search, something like `if 'meetings are held at' in item`:

city_scrapers/spiders/chi_ssa_27.py
```diff
@@ -0,0 +1,170 @@
...
+                meeting_location = self._parse_location(item)
+                continue
+
+            if re.search(r'http', item):
```
pjsier: This can also just be a substring search, but it might make more sense to use CSS selectors or XPath to check if there are links in an element. Here's a quick example of that https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_ssa_21.py#L122

city_scrapers/spiders/chi_ssa_27.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+        yield meeting
+
+    # We can just use "Commission" here, but if it's a committee meeting it should
+    # return the name of the committee instead.
+
+    def pdf_parse(self, item):
+        re1 = '((?:http|https)(?::\\/{2}[\\w]+)(?:[\\/|\\.]?)(?:[^\\s"]*))'  # HTTP URL 1
```
pjsier: This can likely be removed and replaced with just checking for links, and then checking if the `href` attribute ends with `.pdf`. Something like `link.attrib['href'].endswith('.pdf')`. You can also extract the `href` attribute with a CSS selector to do something like `item.css('a::attr(href)')` to get all of the `href` attributes of child links

city_scrapers/spiders/chi_ssa_27.py
```diff
@@ -0,0 +1,170 @@
...
+        return ""
+
+    @staticmethod
+    def _parse_classification():
+        """Parse or generate classification from allowed options."""
+        # This can return COMMISSION generally or COMMITTEE for the committee meetings
+        return COMMISSION
+
+    def _parse_start(self, item):
+        start_datetime = None
```
pjsier: This should use the `datetime.strptime` function

.gitignore
```diff
@@ -1,135 +0,0 @@
-# Byte-compiled / optimized / DLL files
```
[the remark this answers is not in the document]
nmschorr: wow I did everything I could think of to get rid of this.

city_scrapers/spiders/chi_ssa_27.py
nmschorr: Will do.

city_scrapers/spiders/chi_ssa_27.py
nmschorr: I was doing it that way originally, but got it in my head that I needed to use re for everything. Happy to hear that's not the case.  :)

city_scrapers/spiders/chi_ssa_27.py
nmschorr: ok, great, thanks!

Issue #573, opened 2018-11-07
Spider: Chicago Special Service Area #51 Chatham


pjsier:
URL: http://www.cbatechworks.org/
Spider Name: `chi_ssa_51`
Agency Name: Chicago Special Service Area #51 Chatham

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

yu-zhou-7:
I can work on this one

pjsier:
Sounds good!

pjsier:
Closed by #794, thanks @yu-zhou-7!

PR #794, opened 2019-07-21
0573 spider chi ssa 51

yu-zhou-7:
Spider for issue #573

### DIFF
--- a/city_scrapers/spiders/chi_ssa_51.py
+++ b/city_scrapers/spiders/chi_ssa_51.py
@@ -13,7 +13,6 @@ class ChiSsa51Spider(CityScrapersSpider):
     allowed_domains = ["www.cbatechworks.org"]
     start_urls = ["http://www.cbatechworks.org/"]
     location = {"address": "806 East 78th Street, Chicago IL 60619", "name": "QBG Foundation"}
-    last_parsed_date = None
 
     def parse(self, response):
         """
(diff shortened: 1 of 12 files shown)


city_scrapers/spiders/chi_ssa_51.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ChiSsa51Spider(CityScrapersSpider):
+    name = "chi_ssa_51"
+    agency = "Chicago Special Service Area #51 Chatham"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cbatechworks.org"]
+    start_urls = ["http://www.cbatechworks.org/"]
+    location = {"address": "806 East 78th Street, Chicago IL 60619", "name": "QBG Foundation"}
+    last_parsed_date = None
```
pjsier: It looks like this isn't referenced anywhere outside of the `parse` and `_parse_date` function, and we mainly use this to maintain state between requests. It might be a little cleaner to declare this variable inside the `parse` function, and have `_parse_date` return both the parsed date and `last_parsed_date`, updating it if it has changed.

city_scrapers/spiders/chi_ssa_51.py
```diff
@@ -0,0 +1,75 @@
...
+
+            meeting = Meeting(
+                title="Commission",
+                description="",
+                classification=COMMISSION,
+                start=start,
+                end=end,
+                all_day=False,
+                time_notes="",
+                location=self.location,
```
pjsier: It would be good to add a basic `_validate_location` function here like we have in [`chi_ssa_34`](https://github.com/City-Bureau/city-scrapers/blob/c23f1fd17b1ced1c01dff295c8cdf1ca191f0244/city_scrapers/spiders/chi_ssa_34.py#L58). Since this SSA clearly includes the address which doesn't change between meetings, we try to do a sanity check to make sure the address is the same. Otherwise we want to throw an error so that we get a notification from Sentry

yu-zhou-7:
Updated based on suggestions!

PR #820, opened 2019-08-20
0798-spider-il_lottery

ozzysChiefDataScientist:
This is my first spider for city-scrapers so I thought I'd check to see if I'm heading in the right general direction.
Currently, this spider only extracts meeting dates and agendas for **upcoming** meetings. 
Should I add the ability to scrape past meetings and agendas as well? I can refactor this to make it more similar to the ChiHousingAuthoritySpider, which has separate methods for scraping upcoming and past meetings.

### DIFF
--- a/city_scrapers/spiders/il_lottery.py
+++ b/city_scrapers/spiders/il_lottery.py
@@ -1,0 +1,135 @@
+import re
+from datetime import datetime, time
+
+import dateutil.parser
+from city_scrapers_core.constants import BOARD
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class IlLotterySpider(CityScrapersSpider):
+    name = "il_lottery"
+    agency = "Illinois Lottery Control Board"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.illinoislottery.com"]
+    start_urls = ["https://www.illinoislottery.com/illinois-lottery/lottery-control-board/"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        if re.search(r'122 South Michigan Avenue, 19th Floor', response.text) is None:
+            raise ValueError('Meeting address has changed')
+
+        meetings = self._parse_upcoming(response)
+
+        for item in meetings:
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description='',
+                classification=BOARD,
+                start=self._parse_start(item),
+                end=None,
+                all_day=False,
+                time_notes='',
+                location={
+                    'name': 'Chicago Lottery Office',
+                    'address': '122 South Michigan Avenue, 19th Floor, Chicago, IL 60603'
+                },
+                source=response.url,
+            )
+            meeting['id'] = self._get_id(meeting)
+            meeting['status'] = self._get_status(meeting, text=item)
+            meeting['links'] = self._parse_agenda_links(response, date=meeting['start'].date())
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        suffix = ''
+        if 'QTR' in item:
+            suffix = " Quarterly Meeting"
+        elif 'Special' in item:
+            suffix = ' Special Meeting'
+        return "Lottery Control Board{}".format(suffix)
+
+    @staticmethod
+    def parse_time(source):
+        """Returns times given string with format h:mm am|pm"""
+        time_regex = re.compile(r'([1-9]):([0-5][0-9])\s?(am|pm)')
+        hour, minute, period = time_regex.search(source).groups()
+
+        hour = int(hour)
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/il_lottery.py
```diff
@@ -0,0 +1,135 @@
...
+        meeting_lines = response.xpath(meeting_xpath).css('p').get().split("<br>")
+
+        # only keep <br> lines that include a weekday
+        weekdays = ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday']
+        meeting_lines = [x for x in meeting_lines if any(weekday in x for weekday in weekdays)]
+
+        return meeting_lines
+
+    def _parse_agenda_links(self, response, date):
+        '''
```
pjsier: This is a good approach, but it looks like it's re-parsing the full response for every meeting. It might be a little more straightforward to run this before iterating through the meetings and return a dictionary of date keys and link list values, so that in the for loop you can just have something like:

```python
meeting['links'] = date_link_map.get(meeting['start'].date(), [])
```

city_scrapers/spiders/il_lottery.py
```diff
@@ -0,0 +1,135 @@
...
+
+        dt = dateutil.parser.parse(
+            '{mo} {day} {year}'.format(
+                mo=m.group('month'), day=m.group('day'), year=m.group('year')
+            )
+        )
+        return dt.date()
+
+    def _parse_start(self, item):
+        """Parse start date and time."""
```
pjsier: It's up to you on this one because it seems like this is working, but we've usually been cleaning up date strings and then using `datetime.strptime` to do the parsing since the dates are usually a relatively standard format. If it seems like this fits more dates though it's fine to leave as-is

ozzysChiefDataScientist: Thanks for the tip! I updated parse_day() to use datetime.strptime

city_scrapers/spiders/il_lottery.py
ozzysChiefDataScientist: Thanks for the suggestion! I created two dictionaries, agenda_links and minutes_links, that store date keys and link values. Those dictionaries are created before iterating through the meetings.

pjsier: This update looks good! It looks like it's missing some cases where there are only minutes but no agenda though, like October 10, 2012. You should be able to combine the agenda and minutes dictionaries using a `defaultdict` of lists of links, similar to what we do in [`chi_il_medical_district`](https://github.com/City-Bureau/city-scrapers/blob/42fab8ff5a9763a29f9192ef2cdda31228f74970/city_scrapers/spiders/chi_il_medical_district.py#L108), that way you only have to iterate through once, checking if "Agenda" or "Minutes" is in the link and then just pull links from the `defaultdict` with `meeting['links'] = date_link_map[meeting['start'].date()]` to make it even simpler. Let me know if that example isn't clear!

city_scrapers/spiders/il_lottery.py
```diff
@@ -0,0 +1,189 @@
...
+        links = {}
+        for link in response.xpath('//p//a'):
+            link_text_selector = link.xpath('text()')
+            if link_text_selector:
+                link_text = link_text_selector.get()
+                if link_text_substring in link_text:
+                    link_date = self.parse_day(link_text)
+                    link_href = link.xpath('@href').get()
+                    links[link_date] = {
+                        "href": 'https://www.illinoislottery.com{}'.format(link_href),
```
pjsier: One shortcut you can use here is to create the full URL is `response.urljoin(link_href)` which automatically adds the domain of the response to any relative URLs

city_scrapers/spiders/il_lottery.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        if re.search(r'122 South Michigan Avenue, 19th Floor', response.text) is None:
```
pjsier: You should be able to do this search without regex, so it can just be

```python
if "122 South Michigan Avenue" not in response.text:
    ...
```

city_scrapers/spiders/il_lottery.py
ozzysChiefDataScientist: Nifty, thanks!

city_scrapers/spiders/il_lottery.py
ozzysChiefDataScientist: Done!

city_scrapers/spiders/il_lottery.py
ozzysChiefDataScientist: Thanks for the tip! I implemented the `defaultdict` in the [latest commit](https://github.com/City-Bureau/city-scrapers/pull/820/commits/b7e82f758bc417cdf32445ec5c468789df8694ee).

ozzysChiefDataScientist:
Thanks! I just want to edit _parse_time() to use datetime.strptime() before submitting the final PR>

PR #829, opened 2019-09-11
0804-spider-il_liquor_control

kwonhur:
Hi!

This is my shot at building a spider for il_liquor_control. I would love to know if there is anything to fix or add. I am a new developer, so any opinion on my coding is also welcomed. 

Thanks,


city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+    def _parse_location(self, loc_str):
+        """Parse or generate location."""
+        return {
+            "address": '{} Chicago, IL'.format(' '.join(loc_str.split()[:4])),
+            "name": "",
+        }
+
+    def _parse_links(self, item):
```
pjsier: Ignore this if it's just a work in progress, but we'll also need to pull the board meeting minutes from this page https://www2.illinois.gov/ilcc/Divisions/Pages/Legal/Meeting-minutes.aspx. Also, it looks like the agendas are eventually posted on the detail page (had to find this from an older meeting), so we'll want to scrape those as well https://www2.illinois.gov/ilcc/Events/Pages/Board-Meeting-4-17-19.aspx

One thing I'm noticing is that starting on the February 21, 2019 meeting https://www2.illinois.gov/ilcc/Events/Pages/Board-Meeting-2-21-19.aspx all of the meetings listed in the minutes also have a detail page at a standard URL. One way you could handle this is to scrape the minutes page and for all dates on or after February 21, 2019 grab the minutes link, pass it to the request, and then parse the detail page the same way you already are (example here https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_buildings.py#L56-L64). Alternatively you could just parse every date listed with minutes with defaults, but I think starting in February is your best bet

city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,86 @@
...
+        """Parse or generate links."""
+        return [{"href": "", "title": ""}]
+
+    def _parse_source(self, response):
```
pjsier: This can be dropped if you aren't using it

city_scrapers/spiders/il_liquor_control.py
kwonhur: Thanks for your feedback! 
I was navigating through to see if I can use your approach, but it seems like they sometimes delete details page. For example, 
[https://www2.illinois.gov/ilcc/Events/Pages/Board-Meeting-3-21-19.aspx](https://www2.illinois.gov/ilcc/Events/Pages/Board-Meeting-3-21-19.aspx)
does not exist. 

On the other hand, for the same meeting date, an agenda link exists.
[https://www2.illinois.gov/ilcc/Events/EventDocuments/3-21-2019%20ILCC%20Agenda.pdf](https://www2.illinois.gov/ilcc/Events/EventDocuments/3-21-2019%20ILCC%20Agenda.pdf)

I can still try to implement your approach hoping they would not delete individual meeting page in the future. Let me know what you think. Thanks!

city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,161 @@
...
+        meeting detail page is still valid. If valid scrape through _prev_meeting_w_page. If not
+        valid, scrape available info from _prev_meeting_wo_page.
+        """
+
+        if response.url == self.start_urls[0]:
+            for future_meeting_href in response.css('div.soi-link-item a::attr(href)').extract():
+                yield response.follow(future_meeting_href, callback=self._next_meeting)
+
+        if response.url == self.start_urls[1]:
+            for idx, past_meeting_date in enumerate(response.css("div.link-item ::text").extract()):
```
pjsier: A simpler way of doing this might be to iterate over the links and pull out the link text inside the loop, rather than re-parse the response each time. Something like

```python
for past_meeting in response.css("div.link-item a"):
    past_meeting_text = past_meeting.css("*::text").extract_first()
    minutes_href = past_meeting.attrib["href"]
    ...
```

Also, it looks like they only started publishing the detail pages in 2019, so we can parse the year out of the date here and exit out of the loop if the meeting date is before then. We've handled this differently on different spiders, but generally it's easier to only parse meetings within the past year on a consistent basis because we can pull all old documents in a single archival run

city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,161 @@
...
+                dt_object = datetime.datetime.strptime(past_meeting_date, "%B %d, %Y Minutes")
+                minute_href = response.css("div.link-item a::attr(href)").extract()[idx]
+                meeting_url = ("https://www2.illinois.gov/ilcc/Events/Pages/Board-"
+                               "Meeting-{0}-{1}-{2:%y}.aspx".format(
+                                dt_object.month, dt_object.day, dt_object))
+                if requests.get(meeting_url).status_code == 200:
```
pjsier: An easier way of doing this than requesting the same page twice to check the error code might be to override scrapy's behavior and allow parsing 404 pages. That way, we can handle missing pages in the same method by just setting defaults. You should be able to do this by adding this property under `start_urls` to the spider:

```python
custom_settings = {"HTTPERROR_ALLOW_ALL": True}
```

city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,161 @@
...
+                links=self._parse_link(response, self._parse_start(dt_str), None),
+                source=response.url,
+            )
+
+        meeting["status"] = self._get_status(meeting)
+        meeting["id"] = self._get_id(meeting)
+
+        yield meeting
+
+    def _prev_meeting_w_page(self, response, minute_href):
```
pjsier: This is a smart setup for passing arguments to the callback! One thing that might make it more obvious that you're using `kwargs` here is to use the Python convention for keyword arguments like this:

```
def _prev_meeting_w_page(self, response, **kwargs):
    minutes_href = kwargs["minutes_href"]
```

You don't have to assign the variable that way, but it's just to demonstrate that you'll have any keyword arguments available as keys in the `kwargs` dictionary

city_scrapers/spiders/il_liquor_control.py
```diff
@@ -0,0 +1,161 @@
...
+        links = []
+        agenda_href = response.css("div.item.link-item.bullet a::attr(href)").extract()
+        hidden_agenda_href = ("https://www2.illinois.gov/ilcc/Events/EventDocuments/{0.month}"
+                              "-{0.day}-{0.year}%20ILCC%20Agenda.pdf".format(dt_object))
+        if agenda_href:
+            links.append({
+                    'title': 'Agenda',
+                    'href': agenda_href[0]
+                    })
+        elif requests.get(hidden_agenda_href).status_code == 200:
```
pjsier: I think it's fine to remove this and only get the agenda if it's available on the detail page. The detail page is more useful when we only have minutes

kwonhur:
New commit is up, but Travis CI test fails on Python 3.5 for some reason. Anything I can change to get it passed? It seem to be fine 3.6 and 3.7. I was good on my machine as well.

PR #830, opened 2019-09-11
Created bare bones spider for il_police_professionalism

sameerchandra:
This is my first commit. Just wanted to make sure I did everything right till this point so that I can start working on the parse methods. Any advice is welcome. Thanks!

### DIFF
--- a/city_scrapers/spiders/il_police_professionalism.py
+++ b/city_scrapers/spiders/il_police_professionalism.py
@@ -1,5 +1,5 @@
-from time import strptime
 from datetime import datetime
+import re
 
 from city_scrapers_core.constants import (COMMISSION)
 from city_scrapers_core.items import Meeting
@@ -20,46 +20,70 @@ class IlPoliceProfessionalismSpider(CityScrapersSpider):
         Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
         needs.
         """
-        for item in response.xpath("//*[contains(text(),\"Police Professionalism\")]"):
-            meeting = Meeting(
-                title=self._parse_title(item),
-                description=self._parse_description(item),
-                classification=self._parse_classification(item),
-                start=self._parse_start(item),
-                end=None,
-                all_day=self._parse_all_day(item),
-                time_notes=self._parse_time_notes(item),
-                location=self._parse_location(item),
-                links=self._parse_links(item),
-                source=self._parse_source(response),
-            )
+        list_items = response.xpath("//li[contains(@style,'padding-bottom')]")
+        relavent_links = list_items.xpath(
+            'a[(.//*|.)[contains(text(),"Police Professionalism")]]/@href'
+        ).extract()
+        # titles = relavent_links = list_items.xpath(
+        #     'a[(.//*|.)[contains(text(),"Police Professionalism")]]/text'
+        # ).extract()
+        for i in range(0, len(relavent_links)):
+            link = response.urljoin(relavent_links[i])
+            yield response.follow(link, self._parse_item)
 
-            meeting["status"] = self._get_status(meeting)
-            meeting["id"] = self._get_id(meeting)
-
-            yield meeting
+    def _parse_item(self, response):
+        """`_parse_item` should always `yield` Meeting items"""
+        paragraphs = response.xpath('//p')
+        title = response.xpath('//h3/text()').get()
+        meeting = Meeting(
+            title=self._parse_title(response),
+            description=self._parse_description(response, paragraphs),
+            classification=self._parse_classification(title),
+            start=self._parse_start(response, paragraphs),
+            end=None,
+            time_notes='',
+            all_day=False,
+            location=self._parse_location(response),
+            links=self._parse_links(response),
+            source=response.url,
+        )
+        meeting['id'] = self._get_id(meeting)
+        meeting['status'] = self._get_status(meeting, title)
+        return meeting
 
     def _parse_title(self, item):
         """Parse or generate meeting title."""
         return "Commission on Police Professionalism"
 
-    def _parse_description(self, item):
+    def _parse_description(self, item, paragraphs):
+        return_list = []
+        # paragraphs = paragraphs[0:len(paragraphs) - 1]
+        for i in paragraphs:
+            xpath_selector = i.xpath('text()').get()
+            if xpath_selector is not None:
+                return_list.append(xpath_selector.strip().replace(u'\xa0', u''))
+        final_description = " ".join(return_list)
         """Parse or generate meeting description."""
-        return ""
+        return final_description
 
     def _parse_classification(self, item):
         """Parse or generate classification from allowed options."""
         return COMMISSION
 
-    def _parse_start(self, item):
+    def _parse_start(self, item, paragraphs):
         """Parse start datetime as a naive datetime object."""
-        title = item.xpath('text()').get()
-        date = "".join(title[title.index('(') + 1:title.index(')')].split(',')[1:])
-        date = date.strip(' ')
-        date = date.replace('p.m.', 'PM')
-        date = date.replace('a.m.', 'AM')
-        finalDate = datetime.strptime(date, "%B %d %Y %H:%M %p")
-        return finalDate
+        for p in paragraphs.getall():
+            date_match = re.search(r"[a-zA-Z]{3,10} \d{1,2}([a-z]{2})?,? \d{4}", p)
+            time_match = re.search(r"\d{1,2}(:\d{2})? ?[apm\.]{2,4}", p)
+            if date_match:
+                final_date = date_match.group(0)
+            if time_match:
+                final_time = time_match.group(0)
+        final_time = final_time.replace('a.m.', 'AM')
+        final_time = final_time.replace('p.m.', 'PM')
+        final_datetime = final_date + ' ' + final_time
+        final_datetime = datetime.strptime(final_datetime, "%B %d, %Y %H:%M %p")
+        return final_datetime
 
     def _parse_time_notes(self, item):
         """Parse any additional notes on the timing of the meeting"""
@@ -72,17 +96,13 @@ class IlPoliceProfessionalismSpider(CityScrapersSpider):
     def _parse_location(self, item):
         """Parse or generate location."""
         return {
-            "address": "",
-            "name": "Commission on Police Professionalism",
+            "address": "301 S 2nd St, Springfield, IL 62701",
+            "name": "Illinois State Capitol",
         }
 
     def _parse_links(self, item):
         """Parse or generate links."""
-        link = item.xpath('@href').get()
-        return [{
-            "href": "https://www.isp.state.il.us/media/" + link,
-            "title": "Commission on Police Professionalism"
-        }]
+        return [{"href": "", "title": "Commission on Police Professionalism"}]
 
     def _parse_source(self, response):
         """Parse or generate source."""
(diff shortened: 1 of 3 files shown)


pjsier:
@sameerchandra looks good so far!

sameerchandra:
Hi, wrote some parse functions. Can you please check if they are correct? Thanks!

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+    start_urls = ["https://www.isp.state.il.us/media/openmtgs.cfm"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.xpath("//*[contains(text(),\"Police Professionalism\")]"):
```
pjsier: Sometimes agencies will change their markup a bit so that the actual text might be nested inside of a `strong` tag inside the `a` tag, so the selector `"li[style] a"` might be safer to just find links with the `style` attribute on them which only seems to match document links that we're interested in. Then you can get the text of each item with:

```python
item_text = " ".join(item.css("*::text").extract())
```

This way, you pull all of the text within the item, even if it's nested. From there you can check if "Police Professionalism" is included

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+            meeting = Meeting(
```
pjsier: To pull the meetings, we might actually want to parse the detail pages. We can do that by using the meeting links we parsed above and yielding scrapy requests. Here's an example in [`det_library_commission`](https://github.com/City-Bureau/city-scrapers/blob/aa2caea9337d3a509d13302d4a0523c2654f39e6/city_scrapers/spiders/det_library_commission.py#L29-L30)

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+                start=self._parse_start(item),
+                end=None,
+                all_day=self._parse_all_day(item),
+                time_notes=self._parse_time_notes(item),
+                location=self._parse_location(item),
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
```
pjsier: This is where you would pass any text that might indicate whether the meeting is cancelled (likely from the detail page now). It would look something like:

```python
meeting["status"] = self._get_status(meeting, text=item_text)
```

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return COMMISSION
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        title = item.xpath('text()').get()
+        date = "".join(title[title.index('(') + 1:title.index(')')].split(',')[1:])
```
pjsier: The datetimes follow a pretty standard format here, so regex might be easier to parse out the start time information like this (although this case was more complicated, so you'd probably only need the first few lines) https://github.com/City-Bureau/city-scrapers/blob/51db93bfc876ed51dc27df651832bce53bf4c6ac/city_scrapers/spiders/chi_ssa_29.py#L84

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+        date = date.strip(' ')
+        date = date.replace('p.m.', 'PM')
+        date = date.replace('a.m.', 'AM')
+        finalDate = datetime.strptime(date, "%B %d %Y %H:%M %p")
```
pjsier: Small style thing, but Python convention is to use snake case instead of camel case, so this would be `final_date`

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "",
```
pjsier: The address would need to be parsed from the detail page here, and it looks like the location name is generally Illinois State Capitol. Since the location is generally static, we can also just have a static location that we verify is on the page or throw an error if it's not available. Here's an example of that https://github.com/City-Bureau/city-scrapers/blob/51db93bfc876ed51dc27df651832bce53bf4c6ac/city_scrapers/spiders/chi_ssa_29.py#L105-L108

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,89 @@
...
+            "name": "Commission on Police Professionalism",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        link = item.xpath('@href').get()
+        return [{
+            "href": "https://www.isp.state.il.us/media/" + link,
```
pjsier: It's generally easier to use `response.urljoin` to create the URL that would result from clicking the link on the page than statically adding the base path

city_scrapers/spiders/il_police_professionalism.py
sameerchandra: The problem with the above regex is that I'm not able to extract the time

city_scrapers/spiders/il_police_professionalism.py
sameerchandra: I wrote the following code which basically does the same thing

```
for item in response.xpath("//li[contains(@style,'padding-bottom')]"):
            link = item.xpath('a[(.//*|.)[contains(text(),"Police Professionalism")]]/@href'
                              ).extract()
```

city_scrapers/spiders/il_police_professionalism.py
sameerchandra: Done! It is working as intended

city_scrapers/spiders/il_police_professionalism.py
pjsier: There's an example of extracting the time in the same method that might be useful https://github.com/City-Bureau/city-scrapers/blob/51db93bfc876ed51dc27df651832bce53bf4c6ac/city_scrapers/spiders/chi_ssa_29.py#L96

sameerchandra: Great! Thanks a lot! This looks much better than the one I did

city_scrapers/spiders/il_police_professionalism.py
sameerchandra: Cool! Thanks for the suggestion

sameerchandra:
@pjsier Hi, made some changes as you suggested. Please take a look!

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -20,46 +20,70 @@ def parse(self, response):
...
 
-    def _parse_description(self, item):
+    def _parse_description(self, item, paragraphs):
+        return_list = []
+        # paragraphs = paragraphs[0:len(paragraphs) - 1]
+        for i in paragraphs:
+            xpath_selector = i.xpath('text()').get()
+            if xpath_selector is not None:
+                return_list.append(xpath_selector.strip().replace(u'\xa0', u''))
+        final_description = " ".join(return_list)
```
pjsier: Small thing, but you can just return this directly instead of assigning it to a variable first

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -20,46 +20,70 @@ def parse(self, response):
...
-        finalDate = datetime.strptime(date, "%B %d %Y %H:%M %p")
-        return finalDate
+        for p in paragraphs.getall():
+            date_match = re.search(r"[a-zA-Z]{3,10} \d{1,2}([a-z]{2})?,? \d{4}", p)
+            time_match = re.search(r"\d{1,2}(:\d{2})? ?[apm\.]{2,4}", p)
+            if date_match:
+                final_date = date_match.group(0)
+            if time_match:
+                final_time = time_match.group(0)
+        final_time = final_time.replace('a.m.', 'AM')
```
pjsier: A quicker way of doing this would be to just replace periods and make everything upper case
```python
final_time = time_match.group(0).replace('.', '').upper()
```

You also might want to set a default value for time like `final_time = '12:00AM' at the beginning of the function so that it won't error out if a time isn't found

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -72,17 +96,13 @@ def _parse_all_day(self, item):
...
         }
 
     def _parse_links(self, item):
         """Parse or generate links."""
-        link = item.xpath('@href').get()
-        return [{
-            "href": "https://www.isp.state.il.us/media/" + link,
-            "title": "Commission on Police Professionalism"
-        }]
+        return [{"href": "", "title": "Commission on Police Professionalism"}]
```
pjsier: Assuming this is a placeholder for now? This is a weird one because the agenda is also the source link, but since that's the case you can return it that way

```python
def _parse_links(self, response):
    return [{"href": response.url, "title": "Agenda"}]
```

sameerchandra:
Hi, made some changes as you suggested. Please review! Thanks

pjsier:
@sameerchandra the changes are looking good so far, but looks like you need to run the [linting commands](https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools) and get the tests to pass. Let me know when that's working!

sameerchandra:
Hi @pjsier, I made linting changes as suggested. I still haven't wrapped my head over what tests to write. This is my first time

pjsier:
In general we just stick to implementing all of the tests from the template which should cover most of what we’re looking for. They’re simple and mostly to make sure things are basically working and that any changes don’t break existing functionality

sameerchandra:
Hi, I also added tests for the script. Please take a look. Thanks!

city_scrapers/spiders/il_police_professionalism.py
```diff
@@ -0,0 +1,107 @@
...
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        list_items = response.xpath("//li[contains(@style,'padding-bottom')]")
+        relavent_links = list_items.xpath(
+            'a[(.//*|.)[contains(text(),"Police Professionalism")]]/@href'
+        ).extract()
+        # titles = relavent_links = list_items.xpath(
```
pjsier: Small thing but this can be removed

tests/test_il_police_professionalism.py
```diff
@@ -0,0 +1,79 @@
...
+spider = IlPoliceProfessionalismSpider()
+
+freezer = freeze_time("2019-09-11")
+freezer.start()
+
+item = spider._parse_item(test_response)
+
+freezer.stop()
+
+# def test_tests():
```
pjsier: This and the below comment can be removed as well

pjsier:
Looks great, thanks for your work on this!

sameerchandra:
S

> Looks great, thanks for your work on this!

Sure! Thanks for your advice! I got to do web scraping and also got to make my first contribution to open source

PR #832, opened 2019-09-13
Created barebones spider for il_board_of_examiners

mahesh174:
Created a barebone spider. Wanted to make sure that everything is good.  advice is welcome. Thanks.!!

### DIFF
--- a/city_scrapers/spiders/il_board_of_examiners.py
+++ b/city_scrapers/spiders/il_board_of_examiners.py
@@ -50,22 +50,14 @@ class IlBoardOfExaminersSpider(CityScrapersSpider):
 
     def _parse_start(self, item):
         """Parse start datetime as a naive datetime object."""
-        date_str = item.css('.minuteDate::text').get()
+        date_str = re.sub(r"(?<=\d)[a-z]+", "", item.css('.minuteDate::text').get())
         time_str = item.css('.minuteTime::text').get()
-        date_ar = date_str.rsplit(" ")
-        date = re.sub("\D","",date_ar[1])
-        year = re.sub("\D","",date_ar[2])
-        month = 13
-        in_time = datetime.strptime(time_str, "%I:%M %p")
-        out_time = datetime.strftime(in_time, "%H:%M")
+        datetimeObj = ''
         try:
-            month = strptime(date_ar[0],'%B').tm_mon
+            datetimeObj = strptime(date_str+':'+time_str,"%B %d, %Y:%I:%M %p")
         except:
-            month = strptime(date_ar[0],'%b').tm_mon
-
-        temp = year+'-'+str(month)+'-'+date+'T'+out_time+':'+'00'
-        datetimeObj = strptime(temp,'%Y-%m-%dT%H:%M:%S')
-
+            datetimeObj = strptime(date_str+':'+time_str,"%b %d, %Y:%I:%M %p")
+            
         return datetimeObj
 
     def _parse_end(self, item):
@@ -83,7 +75,7 @@ class IlBoardOfExaminersSpider(CityScrapersSpider):
     def _parse_location(self, item):
         """Parse or generate location."""
         seperator = ","
-        [name,address] = seperator.join(item.css(".minutes")[3].css(".minuteLocation").css("p::text").extract()).rsplit("\n",1)
+        [name,address] = seperator.join(item.css(".minuteLocation p::text").extract()).rsplit("\n",1)
         
         return {
             "name": self.remove_special_chars(name),
@@ -105,8 +97,4 @@ class IlBoardOfExaminersSpider(CityScrapersSpider):
 
     def remove_special_chars(string_val):
         """Remove all special chars from string"""
-        return re.sub(
-                r'\s+',
-                ' ',
-                re.sub(r'(\n)|(--em--)|(--em)|(em--)', ' ', string_val),
-                ).strip()
+        return re.sub(r'\s+',' ',string_val).strip()
--- a/city_scrapers/spiders/il_board_of_examiners.py
+++ b/city_scrapers/spiders/il_board_of_examiners.py
@@ -1,0 +1,76 @@
+from city_scrapers_core.constants import NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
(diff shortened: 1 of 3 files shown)


pjsier:
Looks good so far, thanks!

mahesh174:
Hi, I've written some parsing functions. Can you look at those and give some comments or changes, please. Thanks.!

city_scrapers/spiders/il_board_of_examiners.py
```diff
@@ -0,0 +1,112 @@
...
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
```
pjsier: It looks like this could be more simple if it parsed the strings as they were, since they seem to follow a standard format. Is there a reason something like this this wouldn't work (haven't tested this specific code)?

```python
# Remove ordinal date suffix
date_str = re.sub(r"(?<=\d)[a-z]+", "", item.css('.minuteDate::text').get())
time_str = item.css('.minuteTime::text').get()

return datetime.strptime(date_str + time_str, "%B %d, %Y%I:%M %p)
```

city_scrapers/spiders/il_board_of_examiners.py
```diff
@@ -0,0 +1,112 @@
...
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        seperator = ","
```
pjsier: This also looks like it could be simplified a bit by changing the CSS selector:

```python
loc_parts = item.css(".minuteLocation p::text").extract()
name = loc_parts[0]
address = " ".join(loc_parts[1:])
```

city_scrapers/spiders/il_board_of_examiners.py
```diff
@@ -0,0 +1,112 @@
...
+            return [{"href": agenda_link, "title": "Agenda"},{"href": mom_link, "title": "Minutes"}]
+        else:
+            return [{"href": agenda_link, "title": "Agenda"}]
+
+    def _parse_source(self, response):
+        """Parse or generate source."""
+        return response.url
+
+    def remove_special_chars(string_val):
+        """Remove all special chars from string"""
```
pjsier: Is there a reason for splitting out the `--em--` strings? We use the `r'\s+'` a good amount, but  I've only seen the other on Legistar scrapers

city_scrapers/spiders/il_board_of_examiners.py
mahesh174: This looks much simpler and it works Thanks..!!

city_scrapers/spiders/il_board_of_examiners.py
mahesh174: `[name,address] =",".join(item.css(".minuteLocation p::text").extract()).rsplit("\n",1)` I could use this.

`item.css(".minuteLocation p::text").extract()` and `item.css(".minuteLocation").css("p::text").extract()` both gives the same result. But the initial one has lesser selectors. Thanks!!

city_scrapers/spiders/il_board_of_examiners.py
mahesh174: `r'\s+'` works, Splitting out `--em--` strings have nothing specific here though.

mahesh174:
Hi, the meeting description/agenda is in the form of pdf for download. I've added those as links. will that be fine?

pjsier:
Yep, that works!

mahesh174:
Hi, I've updated the changes and written some test. please review,
Thanks.

city_scrapers/spiders/il_board_of_examiners.py
```diff
@@ -0,0 +1,104 @@
...
+
+            yield meeting
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
```
pjsier: This should return `BOARD` since it's the Board of Examiners. You can see more info in [the docs](https://www.ilboe.org/board-information/board-meetings/)

tests/test_il_board_of_examiners.py
```diff
@@ -0,0 +1,75 @@
...
+)
+spider = IlBoardOfExaminersSpider()
+
+freezer = freeze_time("2019-09-13")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+freezer.stop()
+"""
```
pjsier: You can remove this comment

city_scrapers/spiders/il_board_of_examiners.py
```diff
@@ -0,0 +1,104 @@
...
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
```
pjsier: It doesn't look like this is working consistently (you can see some of the output in the [last test run](https://travis-ci.org/City-Bureau/city-scrapers/jobs/586594978)). One way to parse this out would be to first check if there are numbers anywhere in the location string and start the address there for cases like:
Illinois CPA Society
550 W. Jackson Blvd.
9th Floor Building Meeting Room
Chicago, Illinois

If it starts with a number then the whole thing is the address. For addresses that don't include any numbers like the one below you can default to using the last line as the address if there is more than one line.

Northern Illinois University
DeKalb, IL

mahesh174:
Hi @pjsier, I've updated the code with the changes you requested. And can you help me why pipeline got failed for 3.5 version?

pjsier:
@mahesh174 this looks great! There have been some unrelated issues with Python 3.5 tests occasionally failing on Travis CI, but this is good to merge. Thanks!

PR #838, opened 2019-09-22
0670 spider chi midway noise

bendiller:
Hi! I believe this spider and it's tests are finished. This is my first ever PR, so please help me learn if I've done something incorrectly or even just sub-optimally! In addition to the spider and tests, I did complete the yapf/isort/flake8 and pytest steps as specified in the instructions at https://cityscrapers.org/docs/development/.

One note about the code - yapf expanded what had been a fairly compact set of definitions for the links in the test_links() test into a much larger block of code - it went from about 100 lines to closer to 500 lines. I find that section to be much less clear and clean-looking as a result of the yapf changes.

I'm happy to re-work anything that you're unhappy with - looking forward to your feedback!

### DIFF
--- a/city_scrapers/spiders/chi_midway_noise.py
+++ b/city_scrapers/spiders/chi_midway_noise.py
@@ -13,7 +13,9 @@ class ChiMidwayNoiseSpider(CityScrapersSpider):
     agency = "Chicago Midway Noise Compatibility Commission"
     timezone = "America/Chicago"
     allowed_domains = ["www.flychicago.com"]
-    start_urls = ["https://www.flychicago.com/community/MDWnoise/AdditionalResources/pages/default.aspx"]  # noqa
+    start_urls = [
+        "https://www.flychicago.com/community/MDWnoise/AdditionalResources/pages/default.aspx"
+    ]  # noqa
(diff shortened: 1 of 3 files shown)


bendiller:
Whoops, it looks like I used f-strings and Python 3.5 doesn't support those. I'll get that straightened out.

pjsier:
Thanks @bendiller! Everything looks good to go here so I'll go ahead and merge this. Agreed that `yapf` can sometimes be wonky, but the pros of having a consistent style that we can check in a CI outweigh the cons. We may actually move over to `black` soon, which I've only been avoiding because it doesn't support Python 3.5

bendiller:
Hey great! I understand what you mean about `yapf` - I wasn't complaining, more just explaining why that section might have looked so strange, haha.

Anyway - question for you, is there a particular issue you're wanting to see closed? I'd be happy to take a look. Or should I just pick another one at random?

pjsier:
Ah gotcha, yeah I've though the same thing about `yapf` a few times. And I don't think there are any in particular that are higher priority, but I would say feel free to comment on any that look like they've been inactive for over a month since we have a few of those

Issue #803, opened 2019-07-21
Spider: Illinois Investment Policy Board


pjsier:
URL: https://www2.illinois.gov/sites/iipb/Pages/MeetingInformation.aspx
Spider Name: `il_investment_policy`
Agency Name: Illinois Investment Policy Board

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

wraikes:
Hello, I'd like to work on this issue!

pjsier:
@wraikes sounds good, all yours!

pg:
@wraikes Are you working on this issue? If not, I would like to take it on. How does that sound @pjsier?

pjsier:
@pg if you don't hear back in a day or so that works for me!

wraikes:
@pg sorry for the delay, I'm still working on this one, thank you!

pg:
Oh, shoot! I just saw this... I got excited and was eager to help out for #hacktoberfest. I ended up working on this one first to get an understanding of the project. I think I have something working.

pjsier:
@wraikes sorry about the miscommunication here, but happy to help you get started on another spider if you're still interested! And I'll take a look at your PR, @pg

wraikes:
@pjsier no worries! I'll just work on another issue.  thank you.

pjsier:
Closed by #858, thanks @pg!

PR #858, opened 2019-10-04
0803 spider il investment policy

pg:
Here's a shot at #803. Let me know how it looks.

### DIFF
--- a/city_scrapers/spiders/il_investment_policy.py
+++ b/city_scrapers/spiders/il_investment_policy.py
@@ -1,0 +1,194 @@
+import re
+from collections import defaultdict
+from datetime import datetime, time
+from enum import Enum
+
+from city_scrapers_core.constants import BOARD, COMMITTEE, NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class IlInvestmentPolicySpider(CityScrapersSpider):
+    name = "il_investment_policy"
+    agency = "Illinois Investment Policy Board"
+    timezone = "America/Chicago"
+    allowed_domains = ["www2.illinois.gov"]
+    start_urls = ["https://www2.illinois.gov/sites/iipb/Pages/MeetingInformation.aspx"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        self._validate_location(response)
+        links_map = self._parse_links(response)
+
+        for meeting in self._parse_upcoming_meetings(response, links_map):
+            yield meeting
+
+        for item in response.xpath(
+            "//h2[text()='Agendas']/following-sibling::ul"
+            "[not(preceding-sibling::h2[text()='Minutes'])]"
+            "/li/p"
+        ):
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description="",
+                classification=self._parse_classification(item),
+                start=self._parse_start(item),
+                end=None,
+                all_day=False,
+                time_notes="Confirm start time with agency.",
+                location={
+                    "name": "James R. Thompson Center",
+                    "address": "100 W. Randolph St., Room 16-503, Chicago, Illinois"
+                },
+                source=response.url
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            meeting_type = self._parse_meeting_type(meeting['title'])
+            meeting['links'] = links_map.get((meeting['start'].date(), meeting_type), [])
+
+            yield meeting
+
+    def _parse_upcoming_meetings(self, response, links_map):
+        meetings = []
+        upcoming = response.xpath("//strong[text()='2019 Board Meetings']/following::ul[1]")
+        for item in upcoming.xpath("./li"):
+            ddate_str = item.xpath("text()").get().strip()
+            ddate = datetime.strptime(ddate_str, "%m/%d/%y").date()
+            if (ddate, IlInvestmentPolicyMeetingType.IIPB) not in links_map.keys():
+                meeting = Meeting(
+                    title="Investment Policy Board",
+                    description="",
+                    classification=BOARD,
+                    start=datetime.combine(ddate, time(11, 00)),
+                    end=None,
+                    all_day=False,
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,157 @@
...
+            day = date(meeting['start'].year, meeting['start'].month, meeting['start'].day)
+            meeting_type = self._parse_meeting_type(meeting['title'])
+            meeting['links'] = links.get((day, meeting_type), [])
+
+            yield meeting
+
+    @staticmethod
+    def _parse_title(item):
+        """Parse or generate meeting title."""
+        return item.xpath("text()[normalize-space(.)]").get().lstrip(' -').strip()
```
pjsier: Since the titles can sometimes return odd results like "IIPB Board Meeting Agenda" or "COMMITTEE ON SUDAN AND IRAN RESTRICTIONS" even though we know it's constrained to a few values, we should do something similar to what you're doing in `_parse_meeting_type` to pull the title. If there isn't a match for any of the predefined options we can return the existing text

city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        link_text = item.xpath("a/text()[normalize-space(.)]").get()
+        return datetime.combine(self._parse_date(link_text), time(13, 30))
+
+    @staticmethod
+    def _parse_time_notes(item):
+        """Parse any additional notes on the timing of the meeting"""
+        return "The meeting start time may be earlier than 1:30pm when multiple meetings are " \
```
pjsier: From looking at some of the minutes it seems like meetings are almost always on the same day. It might make sense to have a more generic note here like "Confirm start time with agency" and then maybe default the committees to an earlier time. Not sure if you're seeing a different pattern though, so I'm open to suggestions! Ideally we'd want to list a time earlier rather than later

city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,157 @@
...
+               "scheduled for the same day."
+
+    def _parse_links(self, response):
+        """
+        Extracts links for a specific meeting date
+        """
+        link_map = defaultdict(list)
+        for item in response.xpath("//h2[text()='Agendas']/following-sibling::ul/li/p"):
```
pjsier: We'll want to get the minutes here as well, but that should be doable the same way as you've done here by just modifying the XPath and making sure it's matching the date formats

city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        self._validate_location(response)
+        links = self._parse_links(response)
+
+        for item in response.xpath(
```
pjsier: It looks like this is missing the upcoming meeting in December, and we'll want to get that as well. One way of handling this would be to get all of the meeting dates from the keys in the `links` `defaultdict`, add any dates found in the upcoming meetings section, and then iterate through that similar to this:

https://github.com/City-Bureau/city-scrapers-cle/blob/a499b5cb2ca28e73a73823b3e67d478f2b87da0c/city_scrapers/spiders/cuya_planning.py#L31-L40

Let me know if that makes sense!

city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,157 @@
...
+                    "name": "James R. Thompson Center",
+                    "address": "100 W. Randolph St., Room 16-503, Chicago, Illinois"
+                },
+                source=response.url
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            day = date(meeting['start'].year, meeting['start'].month, meeting['start'].day)
```
pjsier: It might be quicker to just call `meeting['start'].date()` in `links.get` unless I'm missing something

pg: That ends up including the start time (1:30pm), but the links map doesn't care for / include the time. I guess you could get the 'start' date and then zero out the hours and minutes to make it match what the links map expects.

city_scrapers/spiders/il_investment_policy.py
pg: This does capture the meeting minutes, too. They end up mapped to the same date and meeting type.

city_scrapers/spiders/il_investment_policy.py
pjsier: `meeting["start"]` should be a `datetime` object, so `date()` should return a `datetime.date` without the time included that should fit the links map. It should essentially be a shortcut for what you're doing here

city_scrapers/spiders/il_investment_policy.py
pjsier: Ah you're right, my mistake. I misunderstood the XPath but looking at the output in Travis CI I'm seeing the minutes

city_scrapers/spiders/il_investment_policy.py
pg: Ah, true! Thanks!

pg:
Sorry, I'm realizing I probably shouldn't be force-pushing the single commit. My aim was to keep the commit history clean when you merged, but I guess you can squash commits on the PR. Now this review isn't as straight-forward as changes have been made.

city_scrapers/spiders/il_investment_policy.py
```diff
@@ -0,0 +1,194 @@
...
+        iipb_substrings=['iipb', 'policy', 'board']
+    ):
+        """Parse or generate meeting title."""
+        title = item.xpath("text()[normalize-space(.)]").get().lstrip(' -').strip()
+        if any(substring in title.lower() for substring in ise_substrings):
+            return "Committee on Sudan and Iran Restrictions"
+        elif any(substring in title.lower() for substring in ibr_substrings):
+            return "Committee on Israel Boycott Restrictions"
+        elif any(substring in title.lower() for substring in iipb_substrings):
+            return "IIPB Board Meeting Agenda"
```
pjsier: This line should return "Investment Policy Board" instead of "Agenda" being in the title

pg:
Fixed that last issue and the builds passed. :-)

PR #860, opened 2019-10-04
0843

firejava:


firejava:
failed on python 3.5, but no scrapy installed on travis?

PR #861, opened 2019-10-05
0843 spider chi_ssa_61

firejava:
trying again for failed travis build


city_scrapers/spiders/chi_ssa_61.py
```diff
@@ -0,0 +1,65 @@
...
+                continue
+            meeting = Meeting(
+                title=self.agency,
+                description='',
+                classification=COMMISSION,
+                start=start,
+                end=None,
+                time_notes='',
+                all_day=False,
+                location=self.location,
```
pjsier: We'll want to add a `_validate_location` function like [here](https://github.com/City-Bureau/city-scrapers/blob/c23f1fd17b1ced1c01dff295c8cdf1ca191f0244/city_scrapers/spiders/chi_ssa_34.py#L58) to double-check the location hasn't changed

tests/test_chi_ssa_61.py
```diff
@@ -0,0 +1,33 @@
...
+)
+spider = ChiSsa61Spider()
+
+freezer = freeze_time("2019-10-04")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+freezer.stop()
+
```
pjsier: We'll want to include all of the existing tests from the template here, not because there's much notable happening but as regression tests to make sure we're aware of anything that will change if we have to update the scraper

city_scrapers/spiders/chi_ssa_61.py
firejava: @pjsier add _validate_location function and Pipfile.lock back in.  Tests pasted on 3.5 this time.

pjsier: @firejava thanks! The updates look great. The last thing is adding back the boilerplate regression tests and then this should be good to go

firejava: added in regression tests

city_scrapers/spiders/chi_ssa_61.py
```diff
@@ -0,0 +1,74 @@
...
+    def parse(self, response):
+        """Parse meetings on upcoming and minutes pages"""
+        self._validate_location(response)
+        today = datetime.now().replace(hour=0, minute=0)
+        for item in response.xpath('//div[@rel="1"]/ul/li/strong'):
+            text = item.xpath('./text()').extract_first()
+            if not re.search(r"[ pam\.]{2,5}", text):
+                continue
+            start = self._parse_start(text)
+            if not start or (start < today):
```
pjsier: Sorry,  I missed this on the first review, but we should also be scraping meetings in the past. They aren't currently posting minutes, but it looks like they [have in the past](https://www.downtownhydeparkchicagoinfo.com/minutes) and we'll want to be able to scrape those and attach them to meetings if they're available

city_scrapers/spiders/chi_ssa_61.py
```diff
@@ -0,0 +1,74 @@
...
+                continue
+            meeting = Meeting(
+                title="commission",
```
pjsier: This should be capitalized as "Commission"

city_scrapers/spiders/chi_ssa_61.py
firejava: Okay, should we scrape the old agendas from this page as well? [Agendas from old website](https://www.downtownhydeparkchicagoinfo.com/agendas)

pjsier: I don't think we need to, since they aren't posted for 2019 yet

firejava: the old minutes are on the old website too, should i still scrape for minutes?  

I can't find any recent minutes or agendas on the new website

pjsier: We can ignore them both for now since they haven’t been posted this year and might not be going forward

firejava:
git comment typo wrong, I added pass meetings, not minutes

city_scrapers/spiders/chi_ssa_61.py
firejava: sounds good

city_scrapers/spiders/chi_ssa_61.py
firejava: fixed this

pjsier:
Everything looks good, thanks @firejava! Merging now

Issue #771, opened 2019-06-25
Spider: Cook County Housing Authority


pjsier:
URL: http://thehacc.org/events/
Doc URL: http://thehacc.org/about/
Spider Name: `cook_housing`
Agency Name: Cook County Housing Authority

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

aminamos:
Hi, I'm very new to Python and open source contributing in general. Currently going through [the guide](https://cityscrapers.org/docs/development/#contribute) and will start looking at this. Thank you!

pjsier:
@aminamos sounds great, feel free to reach out with any questions!

pg:
@pjsier @aminamos Any chance I could take this one over for #hacktoberfest?

pjsier:
@pg all yours if there isn't a response for a day or so!

aminamos:
@pg definitely, happy hacking!

pjsier:
Closed by #862

PR #862, opened 2019-10-05
Add cook_housing (#771)

pg:


city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+
+            yield meeting
+
+    @staticmethod
+    def _parse_title(item):
+        """Parse or generate meeting title."""
+        return item.xpath(".//a/text()[normalize-space(.)]").get().strip()
+
+    @staticmethod
+    def _parse_description(item):
```
pjsier: It looks like the description isn't pulling anything useful, so we can drop it here and just use an empty string

city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+
+        return datetime.combine(ddate, ttimes[0].time()), datetime.combine(ddate, ttimes[1].time())
+
+    @staticmethod
+    def _parse_location(item):
+        """Parse or generate location."""
+        times = item.xpath('.//p[@class="subtitle"]/br/following-sibling::text()').get().strip()
+        address = item.xpath('.//div[@class="info"]/p/text()').get().replace(times, '').strip()
+        name = item.xpath('.//p[@class="subtitle"]/text()').get().strip()
+        return {"address": address, "name": name}
```
pjsier: It looks like this isn't including the city and state here, which we should have in there if they're available or even just default to "Chicago, IL" if it looks like it's consistent

city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+        """Parse or generate meeting description."""
+        return item.xpath('.//p[@class="subtitle"]/following-sibling::p[2]/text()').get().strip()
+
+    @staticmethod
+    def _parse_classification(item):
+        """Parse or generate classification from allowed options."""
+        title = item.xpath(".//a/text()[normalize-space(.)]").get().lower()
+        if 'board' in title:
+            return BOARD
+        elif 'workshop' in title:
```
pjsier: For our purposes we can ignore any meetings including "Workshop" because those seem mostly educational rather than public forums.

city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class CookHousingSpider(CityScrapersSpider):
+    name = "cook_housing"
+    agency = "Cook County Housing Authority"
+    timezone = "America/Chicago"
+    allowed_domains = ["thehacc.org"]
+    start_urls = ["http://thehacc.org/events/"]
```
pjsier: Since you can get different months from the URL, we should try and get 2 months in the past, the current month and then 2 months in the future. Here's an example: https://github.com/City-Bureau/city-scrapers/blob/aa2caea9337d3a509d13302d4a0523c2654f39e6/city_scrapers/spiders/det_library_commission.py#L17-L25

city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    @staticmethod
+    def _parse_title(item):
+        """Parse or generate meeting title."""
+        return item.xpath(".//a/text()[normalize-space(.)]").get().strip()
```
pjsier: If it looks like the title is for a regular board meeting, it might be cleaner to return "Board of Commissioners" here

city_scrapers/spiders/cook_housing.py
```diff
@@ -0,0 +1,91 @@
...
+    def _parse_location(item):
+        """Parse or generate location."""
+        times = item.xpath('.//p[@class="subtitle"]/br/following-sibling::text()').get().strip()
+        address = item.xpath('.//div[@class="info"]/p/text()').get().replace(times, '').strip()
+        name = item.xpath('.//p[@class="subtitle"]/text()').get().strip()
+        return {"address": address, "name": name}
+
+    @staticmethod
+    def _parse_links(item):
+        """Parse or generate links."""
```
pjsier: Links are for content like agendas or minutes, but what's getting included here should actually be returned as the URL string in `_parse_source`

city_scrapers/spiders/cook_housing.py
pg: Thanks for the example!

pg:
I believe I resolved all of your comments.

pjsier:
Looks great, thanks @pg!

PR #863, opened 2019-10-05
0853 spider wayne county lepc

Brian-Lee:
This is my first pull request!

### DIFF
--- a/city_scrapers/spiders/wayne_local_emergency_planning.py
+++ b/city_scrapers/spiders/wayne_local_emergency_planning.py
@@ -1,0 +1,80 @@
+from city_scrapers_core.constants import NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class WayneLocalEmergencyPlanningSpider(CityScrapersSpider):
+    name = "wayne_local_emergency_planning"
+    agency = "Wayne County Local Emergency Planning Committee"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.waynecounty.com"]
+    start_urls = ["https://www.waynecounty.com/departments/hsem/wayne-county-lepc.aspx"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.css(".meetings"):
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description=self._parse_description(item),
+                classification=self._parse_classification(item),
+                start=self._parse_start(item),
+                end=self._parse_end(item),
+                all_day=self._parse_all_day(item),
+                time_notes=self._parse_time_notes(item),
+                location=self._parse_location(item),
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return ""
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        return None
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "",
+            "name": "",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        return [{"href": "", "title": ""}]
+
+    def _parse_source(self, response):
+        """Parse or generate source."""
+        return response.url
--- a/city_scrapers/spiders/wayne_local_emergency_planning.py
+++ b/city_scrapers/spiders/wayne_local_emergency_planning.py
@@ -23,6 +23,8 @@ class WayneLocalEmergencyPlanningSpider(CityScrapersSpider):
 
         # dates in the page before 'Schedule' are not relevant
         relevant_text = response.text[response.text.index("Schedule"):]
+        # check to see if the meeting location still looks right (didn't move)
+        self._validate_location(relevant_text)
         # match dates and try to grab the day of the week if it is present
         days = "Sunday, |Monday, |Tuesday, |Wednesday, |Thursday, |Friday, |Saturday, "
         date_expression = "[A-Z][a-z]{2,8} \\d{1,2},? \\d{4}"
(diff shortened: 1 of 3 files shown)


pjsier:
@Brian-Lee I sent this in the Slack, but it looks like you still need to run the style checks, more info here https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools

Brian-Lee:
> @Brian-Lee I sent this in the Slack, but it looks like you still need to run the style checks, more info here https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools

TY. I think I did the checks, but didn't add one of the reformatted files to my branch before the pull request. Thanks for your help.

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        # the meeting dates I've seen all had the substring 'day, ' as in 'Wednesday, March 6, 2019'
+        meeting_dates = response.xpath('''//p[contains(text(),'day, ')]''')
```
pjsier: This is a good observation, but on occasion we've seen some agencies be inconsistent about including days of the week. It might be safer to get the description text and then search for dates with regex. A pattern we've used a lot has been `[A-Z][a-z]{2,8} \d{1,2},? \d{4}`. The `\d{4}` catches 4 digit numbers and makes it pretty safe to assume there's a year in there.

You can also parse the text content directly without having to manually strip tags, which can be error prone if attributes are added to the elements. The easiest way to do this for XPath is `/text()` or `::text` for CSS selectors and then adding `.extract()` to the selector

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+
+        for item in meeting_dates:
+            # clean off the paragraph tags from the parsed meeting date (item)
+            item = str(item.extract()).replace('<p>', '').replace('</p>', '')
```
pjsier: Mentioned above, but selecting the text content instead of the HTML string is generally a safer bet

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+            # print('------------')
```
pjsier: No worries about these as placeholders while you're working on it, but just leaving this as a reminder to remove them before it's merged

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        prefix = 'Wayne County LEPC Meeting - '
+        # create a unique title for each meeting by combining
```
pjsier: For regular meetings like this, it's actually preferable to not have unique titles since we have the details elsewhere, so "Local Emergency Planning Committee" is fine

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+        ret_val = prefix + item
+        return ret_val
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""  # no description of this specific meeting is shown, so we return an empty string
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return COMMITTEE
```
pjsier: The classifications get tricky, but since this seems like a generally independent body, `ADVISORY_COMMITTEE` might be the safest bet. It looks like it's generally issuing recommendations, but isn't necessarily a direct subgroup of one body like the Finance Committee of a Board of Directors

city_scrapers/spiders/wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,99 @@
...
+        """Parse or generate all-day status. Defaults to False."""
+        return False  # Since the meetings start at 2pm, I assume this must be False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        ''' from webpage:
+            Wayne County Community College, 21000 Northline Road, Taylor,
+            in the MIPSE Building. The MIPSE Building is the fire training
+            facility located at the rear of the campus.'''
+        return {
```
pjsier: Returning a static address makes sense, but we'll want to add a check to make sure it doesn't change with a `_validate_location` method ([here's an example](https://github.com/City-Bureau/city-scrapers/blob/c23f1fd17b1ced1c01dff295c8cdf1ca191f0244/city_scrapers/spiders/chi_ssa_34.py#L58)). Also, since it is static we might as well add the ZIP code 48180 to the end (it makes geocoding later easier)

tests/test_wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,114 @@
...
+
+freezer = freeze_time("2019-10-03")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+freezer.stop()
+
+
+def test_tests():
```
pjsier: Usually we'll want to either remove this entirely or add something like `test_count` that checks how many meetings are returned. That would make sense here, and we could make sure `parsed_items` has a length of 4

tests/test_wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,114 @@
...
+    print("Please write some tests for this spider or at least disable this one.")
+    # assert False
+    assert True  # disabled this test
+
+
+"""
```
pjsier: This can be removed

tests/test_wayne_local_emergency_planning.py
```diff
@@ -0,0 +1,114 @@
...
+def test_source():
+    EXPECTED_SOURCE = "https://www.waynecounty.com/departments/hsem/wayne-county-lepc.aspx"
+    assert parsed_items[0]["source"] == EXPECTED_SOURCE
+
+
+def test_links():
+    assert parsed_items[0]["links"] == []
+
+
+#     assert parsed_items[0]["links"] == [{
```
pjsier: The comments can be removed

Brian-Lee:
Sorry I was out all day. I am looking now.

city_scrapers/spiders/wayne_local_emergency_planning.py
Brian-Lee: Thanks for the tip, I'll get on that,

city_scrapers/spiders/wayne_local_emergency_planning.py
Brian-Lee: I've made some changes on my branch that address these. Thanks for the ideas,

city_scrapers/spiders/wayne_local_emergency_planning.py
Brian-Lee: I'll make that change on my branch now.

city_scrapers/spiders/wayne_local_emergency_planning.py
Brian-Lee: ty. I'll change that.

city_scrapers/spiders/wayne_local_emergency_planning.py
Brian-Lee: Good ideas.

tests/test_wayne_local_emergency_planning.py
Brian-Lee: I think I covered that with my last test:

def test_correct_number_of_items():
    assert len(parsed_items) == 4

tests/test_wayne_local_emergency_planning.py
Brian-Lee: ok

tests/test_wayne_local_emergency_planning.py
Brian-Lee: ok

pjsier:
@Brian-Lee the changes look great, thanks! Merging now

PR #864, opened 2019-10-05
Add chi_ssa_18

AmiMappy:


pjsier:
@AmiMappy is this the PR you want reviewed or is #865? It looks like there's a duplicate

PR #867, opened 2019-10-06
[WIP] draft is finished for the pdf-reading part of the scraper, 

nolanee:
would appreciate any feedback as I go on


city_scrapers/spiders/det_employment_solutions_corp.py
```diff
@@ -0,0 +1,156 @@
...
+                    end_time, date_str, year_str)})
+
+    def _parse_documents(self, response):
+        """Parse agenda and minutes page"""
+        link_map = self._parse_link_map(response)
+        for date in self.meeting_dates:
+            meeting = Meeting(
+                # FIXME may be able to be more specific by by matching
+                # with links found with _parse_link_map?
+                title="Detroit Employment Solutions Corporation",
```
pjsier: This might be difficult, but we'd want to differentiate between meetings of the Board and different committees

nolanee: Okay, I will submit a pr when I have that (and the links) working

PR #871, opened 2019-10-07
WIP 0856 spider det employment solutions corp

nolanee:
I think I'm fairly close to done, but I have run into a snag.

Things seem to work okay when I run the the spider individually, but I know something is wrong because when I run the tests I get the following error:

 ___________________ ERROR collecting tests/test_det_employment_solutions_corp.py ___________________
tests/test_det_employment_solutions_corp.py:31: in <module>
    key=itemgetter("start"),
E   TypeError: 'Request' object is not subscriptable

It seems like I'm no longer getting Meeting objects back like before?


tests/test_det_employment_solutions_corp.py
```diff
@@ -0,0 +1,94 @@
...
+    url="https://www.descmiworks.com/about-us/public-meetings/",
+)
+spider = DetEmploymentSolutionsCorpSpider()
+
+freezer = freeze_time("2019-07-20")
+freezer.start()
+
+spider._parse_schedule_pdf(test_pdf_response)
+parsed_items = sorted(
+    [item for item in spider.parse(test_response)],
```
pjsier: It looks like changing this to use `_parse_documents` should fix your problem. `parse` is yielding another Request object, so you can cut out some steps that the actual scraper is doing

nolanee: Okay, thanks!

nolanee:
So the Travis Cl build failed?

pjsier:
@nolanee it looks like you need to run the linting checks. We've got more info [in the docs](https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools), but you can run:

```bash
flake8
isort
yapf --in-place --recursive ./city_scrapers ./tests
```

And then commit the changes to fix these. Also, if you click Details next to the failing Travis CI check, or click one of the red Xs and then click on one of the jobs (i.e. Python 3.6) you can see the output of the tests to get a sense of what's failing. I can take a closer look once that's fixed!

city_scrapers/spiders/det_employment_solutions_corp.py
```diff
@@ -0,0 +1,183 @@
...
+        """Parse or generate links. Returns a dictionary of month, year tuples and link lists"""
+        link_map = defaultdict(list)
+
+        title_divs = response.css('div[class="meeting-min_inner-wrapper"]')
+        for div in title_divs:
+            date = div.css("p::text").extract_first().strip()
+            date = date.split('/')
+            link_start = datetime(int(date[2]), int(date[0]), int(date[1]))
+            link = div.css("a").xpath('@href').get()
+            print(link)
```
pjsier: This can be removed

city_scrapers/spiders/det_employment_solutions_corp.py
```diff
@@ -0,0 +1,183 @@
...
+            link_map[(link_start.month, link_start.year)].append({"title": "Minutes", "href": link})
```
pjsier: This would need to include the title of the meeting as the key as well as the date. Right now it's including links for the committees for board meetings as well as for the committee meetings

city_scrapers/spiders/det_employment_solutions_corp.py
```diff
@@ -0,0 +1,183 @@
...
+
+        return link_map
+
+    def _parse_title_map(self, response):
+        """parse or generate titles Returns a dictionary of month, year tuples and title lists"""
```
pjsier: Related to the above comment, it looks like there are multiple types of meetings on the same day

nolanee: Thanks for the feedback! I will work on these issues and resubmit (along with more robust tests to detect them).

nolanee: It also looks like there are some meetings on the minutes page that are not public. Should those be included as well?

pjsier: Any in particular that don't seem to be public? I've only skimmed it but I haven't seen any. In general, if we can safely ignore ones that aren't public without accidentally ignoring public ones we should give it a try, but even the conference calls still qualify as public meetings

nolanee: Specifically, the Mayor’s Workforce Development Board meetings don't show up in the public meeting schedule pdf but minutes are listed on the minutes page. I have fixed the link parser to look up by title as well, but it seems like I will then need to find a way to get the titles out of the pdf itself, since it seems difficult to assign the right titles to two meetings that occur on the same day using the minutes page.

pjsier:

Issue #843, opened 2019-10-03
Spider: Chicago Special Service Area #61 Hyde Park


pjsier:
URL: http://www.downtownhydeparkchicago.com/about/
Spider Name: chi_ssa_61
Agency Name: Chicago Special Service Area #60 Hyde Park

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

firejava:
going to give this one a shot

firejava:
#860

pjsier:
Closed by #861

pg:
Hmm...I just noticed Agency Name lists `#60`, but spider name and the issue title list `#61`. I think the correct value is `#61`

PR #874, opened 2019-10-08
Fix typo in agency name (#843)

pg:
Teeny change, but an important one, I suppose (especially for reducing confusion). :-P


pjsier:
Ah thanks for catching this!

PR #875, opened 2019-10-09
Add chi_ssa_18

AmiMappy:


pjsier:
It looks like there are still extra/duplicate files here? There should only be three additional files, and they should all be in subdirectories. Let me know if I can clarify anything there

pjsier:
It's a valid module name, but it's not being imported from the right location. You can import it from `city_scrapers.spiders.chi_ssa_18`, but importing from `chi_ssa_18` is expecting it to be at the root level

tests/test_chi_ssa_18.py
```diff
@@ -0,0 +1,86 @@
...
+from os.path import dirname, join
+
+import sys
+import pytest
+from chi_ssa_18 import ChiSsa18Spider
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.utils import file_response
+from freezegun import freeze_time
+
+sys.path.append('city-scrapers\city_scrapers\spiders')
```
pjsier: We shouldn't modify the path here. The line importing the spider can be the following to fix this:

```python
from city_scrapers.spiders.chi_ssa_18 import ChiSsa18Spider
```

AmiMappy:
I made a commit, but Travis does not seem to check it.

pjsier:
Travis CI only runs on pull requests that aren't drafts, so I'll mark this as ready for review now

AmiMappy:
I am sorry, I will no longer be able to continue with the project.

pjsier:
Thanks for the heads up!

PR #876, opened 2019-10-09
0656-spider-cook_county_board_ethics

pyence:
Please review and comment

### DIFF
--- a/city_scrapers/spiders/cook_county_board_ethics.py
+++ b/city_scrapers/spiders/cook_county_board_ethics.py
@@ -1,3 +1,4 @@
+import re
 from datetime import datetime
 
 import scrapy
@@ -84,7 +85,11 @@ class CookCountyBoardEthicsSpider(CityScrapersSpider):
 
     def _parse_title(self, response):
         """Parse or generate event title."""
-        return response.xpath('//h1/text()').extract_first()
+        title = response.xpath('//h1/text()').extract_first()
+        if re.match(r'(?i)^Cook County(:?) Board of Ethics Meeting$', title):
+            return 'Board of Ethics'
+        else:
+            return re.sub(r'(?i)^Cook County(:?) ', '', title)
 
     def _parse_description(self, response):
         """
--- a/city_scrapers/spiders/cook_county_board_ethics.py
+++ b/city_scrapers/spiders/cook_county_board_ethics.py
@@ -1,10 +1,7 @@
-import re
 from datetime import datetime, timedelta
 
 import scrapy
-from city_scrapers_core.constants import (
-    ADVISORY_COMMITTEE, BOARD, COMMISSION, COMMITTEE, NOT_CLASSIFIED
-)
+from city_scrapers_core.constants import BOARD
 from city_scrapers_core.items import Meeting
 from city_scrapers_core.spiders import CityScrapersSpider
 
@@ -39,7 +36,7 @@ class CookCountyBoardEthicsSpider(CityScrapersSpider):
         meeting = Meeting(
             title=title,
             description=self._parse_description(response),
-            classification=self._parse_classification(title),
+            classification=BOARD,
             start=self._parse_start(response),
             end=self._parse_end(response),
             time_notes='',
@@ -57,22 +54,10 @@ class CookCountyBoardEthicsSpider(CityScrapersSpider):
         Get urls for all Board of ethics meetings on the page.
         """
         return [
-            response.urljoin(href) for href in
-            response.xpath('//a[contains(@href, "board-ethics")]').css('a::attr(href)').extract()
+            response.urljoin(href)
+            for href in response.xpath('//a[contains(text(), "Board of Ethics")]'
+                                       ).css('a::attr(href)').extract()
         ]
-
-    @staticmethod
-    def _parse_classification(name):
-        name = name.upper()
-        if re.search(r'A(C|(DVISORY)) (COMMITTEE|COUNCIL)', name):
-            return ADVISORY_COMMITTEE
-        if 'BOARD' in name:
-            return BOARD
-        if 'COMMITTEE' in name:
-            return COMMITTEE
-        if 'COMMISSION' in name:
-            return COMMISSION
-        return NOT_CLASSIFIED
 
     def _parse_location(self, response):
         """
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/cook_county_board_ethics.py
```diff
@@ -0,0 +1,151 @@
...
+        meeting['status'] = self._get_status(meeting)
+        return meeting
+
+    def _get_event_urls(self, response):
+        """
+        Get urls for all Board of ethics meetings on the page.
+        """
+        return [
+            response.urljoin(href) for href in
+            response.xpath('//a[contains(@href, "board-ethics")]').css('a::attr(href)').extract()
```
pjsier: It might be safer to check the text than the URL here. Either way works, but we've seen some strange URL issues on sites before

city_scrapers/spiders/cook_county_board_ethics.py
```diff
@@ -0,0 +1,151 @@
...
+        ]
+
+    @staticmethod
+    def _parse_classification(name):
```
pjsier: If we're only parsing Board of Ethics, you shouldn't need this and it should always be `BOARD`

tests/test_cook_county_board_ethics.py
```diff
@@ -0,0 +1,65 @@
+from datetime import datetime
+
+from city_scrapers_core.constants import BOARD, PASSED
+from tests.utils import file_response
+
```
pjsier: We should leave the `freezegun` boilerplate in here

tests/test_cook_county_board_ethics.py
```diff
@@ -0,0 +1,65 @@
...
+def test_all_day():
+    assert item['all_day'] is False
+
+
+def test_classification():
+    assert spider._parse_classification('Board of Ethics') == BOARD
+
+
+def test_status():
+    assert item['status'] == PASSED
```
pjsier: Looks like this test is failing, not sure why it would be though

pyence:
Changes implemented. Do hope everything's up to code now

city_scrapers/spiders/cook_county_board_ethics.py
```diff
@@ -0,0 +1,136 @@
...
+class CookCountyBoardEthicsSpider(CityScrapersSpider):
+    name = 'cook_county_board_ethics'
+    agency = 'Cook County Government Board of Ethics'
+    timezone = 'America/Chicago'
+    allowed_domains = ['www.cookcountyil.gov']
+
+    def start_requests(self):
+        today = datetime.now()
+        next_month = today.replace(day=28) + timedelta(days=5)
+        for dt in [today, next_month]:
```
pjsier: This should do a time range of 2 months in the past through 2 months in the future. Here's an example:

https://github.com/City-Bureau/city-scrapers/blob/3ae5c3e2304c9fb47576ba8fa71d05993e2055ae/city_scrapers/spiders/cook_housing.py#L15-L23

tests/test_cook_county_board_ethics.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_location():
+    assert item['location'] == {
+        'name': '',
+        'address': '69 W. Washington Street, Suite 3040 Chicago , IL  60602',
```
pjsier: Small thing, but we should clean up extra spaces and spaces before commas. A quick way of doing that would be something like this:

```python
# Replace all duplicate whitespace characters with a single space, remove leading/trailing spaces
loc_str = re.sub(r"\s+", " ", loc_str).strip()
# Remove all whitespaces that are before commas
loc_str = loc_str.replace(" ,", ",")
```

city_scrapers/spiders/cook_county_board_ethics.py
```diff
@@ -0,0 +1,136 @@
...
+        Parse or generate all-day status. Defaults to false.
+        """
+        date = response.xpath('//span[@class="date-display-single"]/descendant-or-self::*/text()'
+                              ).extract()
+        date = ''.join(date).upper()
+        return 'ALL DAY' in date
+
+    def _parse_title(self, response):
+        """Parse or generate event title."""
+        return response.xpath('//h1/text()').extract_first()
```
pjsier: Any changes in the title are likely unneeded noise, so we should check for things like "Special" in the title and return the title if it's there, but otherwise just confirm that "Board" is in the title and return "Board of Ethics"

tests/test_cook_county_board_ethics.py
pyence: Would you like to retain the comma between 'Chicago' and 'IL 60602'?

pjsier: It's not a priority, so whatever's easiest is fine

city_scrapers/spiders/cook_county_board_ethics.py
pyence: If possible please elaborate on things to check for besides "Special".
Can I just return the title if it's not 'Cook County: Board of Ethics'?

pjsier: We don't want any boilerplate like "Cook County" included, so we can basically do that as long as we also remove "Cook County: " (removing it whether there's a colon or not)

pyence:
Implemented! How's this?

city_scrapers/spiders/cook_county_board_ethics.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        date = response.xpath('//span[@class="date-display-single"]/descendant-or-self::*/text()'
+                              ).extract()
+        date = ''.join(date).upper()
+        return 'ALL DAY' in date
+
+    def _parse_title(self, response):
+        """Parse or generate event title."""
+        title = response.xpath('//h1/text()').extract_first()
+        if re.match(r'(?i)^Cook County(:?) Board of Ethics Meeting$', title):
```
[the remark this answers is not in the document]
pyence: Sorry to keep talking your ear off about this. A match on any title including "Board of Ethics" except when there's a "Special" in there alright with you?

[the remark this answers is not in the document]
pjsier: What we've done a lot of the time is a series of substring checks, so first checking for "Special" and returning "Special Board of Ethics Meeting", then checking for "Board of Ethics", and then falling back to returning the title if those both fail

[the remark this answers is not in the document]
pyence: Alright. I'll do that.

PR #877, opened 2019-10-09
[WIP]genererated barebones spider, test. and html files

Brian-Lee:
adding a spider


pjsier:
@Brian-Lee looks like there might be an issue with your branching here, since I'm seeing files in the diff from the last scraper. Have you updated your local `master` branch and created a new one off of that?

Brian-Lee:
I did, but let me double check that I did it correctly. If I have to, I may just start over.

PR #882, opened 2019-10-11
0805 spider il pollution control

bendiller:
Hi Patrick - 

I believe this spider is ready for your review. This one was a little tricky! I had to implement something that I didn't see in any of the other spiders, which gives me some pause. I'll be happy to discuss this with you and re-work that part if you have a more preferred way to solve the problem.

The difficulty came from the fact that the links for the meeting minutes came from a separate URL, and one which required diving down a few levels to get to the PDFs. Traversing this directory structure yielded a bunch of `Requests`, which would get naturally get scheduled and run long after the primary URL was processed and turned into Meeting objects.

Normally it'd be easy enough to combine these disparate data in an Item Pipeline, but since this project doesn't appear to use one-off pipelines, I looked for other possibilities, eventually settling on connecting to the `spider_idle` signal and using that to trigger parsing of the primary URL (with the bulk of the meeting data), once all the meeting minutes links had been discovered and stored. This allowed me to associate the links to the Meeting objects within the spider. It works well and I haven't found any problems with this implementation.

Anyway, sorry for the novel - I hope I explained it well enough. The code isn't as convoluted as I've made it sound. If you've encountered a problem like this before and would like me to approach the solution differently, please give me some guidance on how and I'll by happy to accommodate. 

Thanks!


bendiller:
Oh, and I almost forgot - I have a couple more considerations/notes for you:
1. I wasn't sure how to classify "hearings". I left them as NOT_CLASSIFIED but perhaps BOARD is appropriate.
2. Locations - those which listed both the Chicago and Springfield offices, I selected the Chicago as the location, believing that to be the main office. Let me know if you'd like something different.
3. Locations - some of them were indeterminate; I assigned "name" and "address" as empty strings for these.

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.link_map = dict()  # Populated by self._parse_minutes()
+        self.link_parse_complete = False
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def from_crawler(cls, crawler, *args, **kwargs):
+        """ Overridden `from_crawler` to connect `spider_idle` signal. """
+        spider = super(IlPollutionControlSpider, cls).from_crawler(crawler, *args, **kwargs)
```
pjsier: In Python 3 you should be able to just call `super()` without specifying the class

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+        crawler.signals.connect(spider.spider_idle, signal=scrapy.signals.spider_idle)
+        return spider
+
+    def spider_idle(self):
+        """ React to `spider_idle` signal by starting JSON parsing after _parse_minutes."""
+        if not self.link_parse_complete:
```
pjsier: First off, the `spider_idle` setup is really well done! We're using that in some of the pipelines in `city-scrapers-core`, but this is a really solid application.

I don't think you need the `link_parse_complete` attribute here though. Because you're disconnecting from the `spider_idle` handle in the method it can only run once which should achieve the same think as a boolean flag for checking whether it's done.

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+                source=self._parse_source(item),
+            )
+
+            meeting["links"] = self._parse_links(meeting)
+            meeting["status"] = self._parse_status(meeting, item)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_status(self, meeting, item):
```
pjsier: The `_get_status` method in the `CityScrapersSpider` implements a lot of this behavior, so you can get the same result by removing this method and passing the text as a `kwarg`, so this would work:

```python
meeting["status"] = self._get_status(meeting, text=" ".join([item['CalendarTypeDesc'], item['Description']]).lower())
```

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+        elif meeting["start"] < datetime.now():
+            return PASSED
+        else:
+            return TENTATIVE
+
+    def _parse_classification(self, title):
+        """Parse or generate classification from allowed options."""
+        if "Board" in title:
+            return BOARD
+        elif "Seminar" in title:
```
pjsier: I think we can ignore Seminars here, since it isn't really what we're looking for. I'm not sure if hearings are open to the public, but I think they should be safe to ignore too

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+            return {
+                "address": "",
+                "name": "",
+            }
+
+    def _parse_links(self, meeting):
+        """ Associate Meeting objects with previously-scraped links """
+        key = meeting['start'].date()
+        if key in self.link_map:
+            return [{"href": self.link_map[key], "title": "Meeting Minutes - {}".format(key)}]
```
pjsier: Since the minutes are connected to a meeting with a date already, we can just make the title "Minutes"

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _parse_source(self, item):
+        """Parse or generate source."""
+        rel_url = scrapy.Selector(text=item["Description"]).xpath(".//a/@href").get()
+        if rel_url:
+            return urljoin("https://{}".format(self.allowed_domains[0]), rel_url)
+        else:
+            return self.json_url
```
pjsier: We should default this to the HTML page calendar since that would be easier for people to refer back to than the raw JSON https://pcb.illinois.gov/ClerksOffice/Calendar

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+        """
+        for item in response.xpath("//iframe/@src"):
+            yield scrapy.Request(item.get(), callback=self._parse_minutes)
+
+    def _parse_minutes(self, response):
+        """ Traverse tree of URLs and populate self.link_map """
+        for item in response.xpath("//td[@class='name']/a"):
```
pjsier: I don't think we want to scrape all the way back through the archive every time, since we'd only need to pull that information once. It would be great to just pull minutes for the past two years displayed

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,164 @@
...
+            raise scrapy.exceptions.DontCloseSpider
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+        """
+        for item in response.xpath("//iframe/@src"):
+            yield scrapy.Request(item.get(), callback=self._parse_minutes)
+
+    def _parse_minutes(self, response):
```
pjsier: It looks like the most recent board meeting agenda is posted on this page: https://pcb.illinois.gov/CurrentAgendas

It would be great to pull this using the same `spider_idle` setup and parse the PDF for the board meeting date to associate so we can get an agenda as well as minutes

city_scrapers/spiders/il_pollution_control.py
bendiller: Good point! I'll fix that.

city_scrapers/spiders/il_pollution_control.py
bendiller: Thanks for the kind words! I bet you're right, I'll look into it. That boolean was in there with a handful of other stuff when I was in the "how on earth do I make this work" phase of development, and I never even considered it might've become redundant once I got it working.

city_scrapers/spiders/il_pollution_control.py
bendiller: Oh, interesting. I figured the hearings would be where the juicy stuff happened, haha. I'll get these cleaned up.

city_scrapers/spiders/il_pollution_control.py
bendiller: Yep that makes much more sense.

city_scrapers/spiders/il_pollution_control.py
bendiller: Sure thing, will do.

city_scrapers/spiders/il_pollution_control.py
bendiller: Wow, you're right! The embarrassing part is I looked at the `CityScrapersSpider._get_status` method first and somehow reached the conclusion that it wouldn't work. And then proceeded to write something almost identical.

bendiller:
Two things:
1. I'm not sure why that's failing. Travis CI details show that one of the test_links() assertions comes back false, but all the tests pass on my machine, and the test HTML file containing those links does still contain the link that threw the error.

2. I haven't implemented scraping the Agenda links yet, but I wanted to push the fixes for the other comments for your review in the meantime. Of particular note: these changes removed ~60% of the overall `Meetings` and kind of removed the need for some of the logic, particularly in `parse_source()` and `parse_location()`. Several of the looked-for conditions no longer appear in the remaining `Meeting` objects. I wanted to get your approval before removing those pieces of logic, however. It's possible they could be useful still in the future, but not with the current data set. Best to get it out of there for now and re-consider if it causes a problem down the road? Or leave it in?

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,160 @@
...
+            return {
+                "address": "",
+                "name": "",
+            }
+
+    def _parse_links(self, meeting):
+        """ Associate Meeting objects with previously-scraped links """
+        key = meeting['start'].date()
+        if key in self.link_map:
+            return [{"href": self.link_map[key], "title": "Minutes - {}".format(key)}]
```
pjsier: To clarify what I meant here, I think "title" can just be "Minutes" without adding in the date. I realize it was unclear before

bendiller: 1. Hmm, I'm still confused though, because this build passed the previous time around, and I'd expect it to have been subject to the same issue. Either way, I'll change the test to something more reasonable like you suggested. Probably not a lot of added value in the way I've written these.

2. Yep, no problem, I'll leave em in! Should be reasonably bullet-proof, at least within the generally brittle nature of web scraping.

3. Whoops, I'll go ahead and get the date out of that title.

bendiller:
Okay, I'm gonna stop spamming attempts on this thing here, sorry for that - I'll just wait for feedback from you. Each one of these commits have passed all tests on my end; I tried to remove the dependency on `FakeDate`, but either something else is going on or I just don't understand it well enough. I don't understand CI well enough to have any intuition about why this is failing.

Worth mentioning - I'm testing the links completely independently from the rest of the parsing to keep it as simple and atomic as possible - it's really only testing the results of the `_parse_minutes` method. Those links are not being associated with Meeting objects. Maybe I need to do that.

Anyway, like I said, I'm worried you're getting spammed with notifications about this so in case that's true, I'm going to drop this until I hear from you.

pjsier:
Looks like it's probably the last point you mentioned about `_parse_minutes` not being called. Since that isn't called, you'll probably need to have a line that runs `spider._parse_minutes` right before it calls `_parse_json` when setting up the fixtures. Here's an example of that (except you won't need to assign the result since it doesn't return a value):

https://github.com/City-Bureau/city-scrapers-akr/blob/f32b12bb9791082e199e7a428e6818037b3c4754/tests/test_akr_metro_transportation_study.py#L11-L27

bendiller:
Well, `_parse_minutes ` does get called - but I opted to call it within `test_links` since that's the only test that utilizes it. I wonder if that's the problem - maybe I need to move it outside. If that doesn't fix it alone, then I'll run the `_parse_links` method to associate the links with `Meeting` objects.

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            return {
+                "address": "",
+                "name": "",
+            }
+
+    def _parse_links(self, meeting):
+        """ Associate Meeting objects with previously-scraped links """
+        key = meeting['start'].date()
+        if key in self.link_map:
```
pjsier: The tests just made me notice that this needs to return an empty list instead of `None` if nothing is found

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,160 @@
...
+                href = item.xpath("@href")[0].get()
+                text = item.xpath("b/text()")[0].get().strip()
+                if not any([(year in text) for year in self.relevant_years]):
+                    continue  # Link does not contain documents from recent years
+                if text[-4:] == ".pdf":
+                    text = text[:-4]
+            except IndexError:
+                continue
+
+            url = urljoin("https://{}".format(self.allowed_domains[0]), href)
```
pjsier: Missed this earlier, but looking at it now from the tests and unless I'm missing something `response.urljoin(href)` should be quicker here

city_scrapers/spiders/il_pollution_control.py
bendiller: Whoops, alrighty, I'll get that in there.

city_scrapers/spiders/il_pollution_control.py
bendiller: I seem to remember there being a problem with that approach but I'm not 100% sure there; I'll take a look and verify.

bendiller:
I think I'm out of ideas on the CI problem for the moment. I tried moving the `_parse_minutes` call between the `freezer.start` and `.stop` calls, and I've also now implemented the actual stitching together of the separate sets of results (meeting objects and links). 

I'm really sorry for the million commits and pushes. Each of them pushes has passed all tests on my machine (and I've even deliberately broken the `sample_link` string to make it fail as a sanity check). And the strangest part is that the first push of the PR passed (well, second push - first had f-strings problems).

pjsier:
Not a problem! My best guess is that it's related to there being several possible outcomes from `_parse_minutes` including potential network requests, but for now I think it's safe to set `link_map` manually in the test setup and just verify that `_parse_links` is working from that. Once that and the other comments are resolved this should be good to go

city_scrapers/spiders/il_pollution_control.py
bendiller: I've implemented this feature and it appears to work! We'll have to hope the next PDF they post shares a similar format - obviously with a sample size of 1 it's hard to make predictions.

bendiller:
I believe all the comments and requests are complete at this point. My only remaining question is whether we ought to delete `il_pollution_control.html` from the test files. It's no longer used - it was there to be scraped for the links test, but it's no longer being read by anything.

Thanks!

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,203 @@
...
+        if key in self.agenda_map:
+            links.append({"href": self.agenda_map[key], "title": "Agenda"})
+
+        return links
+
+    def _parse_source(self, item):
+        """Parse or generate source."""
+        rel_url = scrapy.Selector(text=item["Description"]).xpath(".//a/@href").get()
+        if rel_url:
+            return urljoin("https://{}".format(self.allowed_domains[0]), rel_url)
```
pjsier: Here's another spot where `response.urljoin` can be used instead as long as `response` is passed to `_parse_source`

city_scrapers/spiders/il_pollution_control.py
```diff
@@ -0,0 +1,203 @@
...
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from PyPDF2 import PdfFileReader
+
+
+class IlPollutionControlSpider(CityScrapersSpider):
+    name = "il_pollution_control"
+    agency = "Illinois Pollution Control Board"
+    timezone = "America/Chicago"
+    allowed_domains = ["pcb.illinois.gov"]
```
pjsier: Small thing, but we'll be getting rid of `allowed_domains` going forward because we don't need to restrict our crawlers as much since they're pretty targeted. If you could drop this while fixing the other `urljoin` method that'd be great!

tests/test_il_pollution_control.py
```diff
@@ -0,0 +1,95 @@
...
+from os.path import dirname, join
+
+import pytest
+from city_scrapers_core.constants import BOARD, CANCELLED, PASSED, TENTATIVE
+from city_scrapers_core.utils import file_response
+from freezegun import freeze_time
+
+from city_scrapers.spiders.il_pollution_control import IlPollutionControlSpider
+
+test_minutes_response = file_response(
```
pjsier: Agreed that this can be removed and the `il_pollution_control.html` file can be deleted

city_scrapers/spiders/il_pollution_control.py
bendiller: Alrighty! Just for my own benefit, what makes response.urljoin() preferred?

pjsier: It's mostly because we're moving away from including `allowed_domains` since it's generally unnecessary for us, but it's also more explicit about including the protocol and domain from the current response rather than hard-coding values to start

city_scrapers/spiders/il_pollution_control.py
bendiller: Done!

tests/test_il_pollution_control.py
bendiller: Done!

pjsier:
Thanks for this @bendiller! Merging now

PR #884, opened 2019-10-11
WIP: 0848-spider-wayne_state_university

moisesjsanchez:
Hello, this is my first time writing a scraper with python using scrapy. I am having some issues getting the code to run as of now, I get the error "response not defined" when running PyTest.  I was wondering if I can please get some feedback about what I have written so far? Thank you!

### DIFF
--- a/city_scrapers/spiders/wayne_state_university.py
+++ b/city_scrapers/spiders/wayne_state_university.py
@@ -1,0 +1,94 @@
+from city_scrapers_core.constants import BOARD, COMMITTEE
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+import re
+
+class WayneStateUniversitySpider(CityScrapersSpider):
+    name = "wayne_state_university"
+    agency = "Wayne State University"
+    timezone = "America/Chicago"
+    allowed_domains = ["http://bog.wayne.edu"]
+    start_urls = ["http://bog.wayne.edu/"]
+    location = {
+            "name": "McGregor Memorial Conference Center", 
+            "address": "Room BC 495 Gilmour Mall, Detroit, MI 48202",
+        }
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        self._validate_location(response)
+        for item in response.xpath('//*[@id="content"]/div[1]/div/h2').re('((?:January|Feburary|March|April|May|June|July|August|October|September|November|December)\s+\d{1,2},\s+\d{4})'):
+            meetings = self._parse_meeting_types(response)
+            for meeting_type in meetings: 
+                meeting = Meeting(
+                    title = meeting_type[0],
+                    description=self._parse_description(item),
+                    classification= meeting_type[-1], 
+                    start= meeting_type[1],
+                    end= "Ending Time Not Specified",
+                    all_day=self._parse_all_day(item),
+                    time_notes=self._parse_time_notes(item),
+                    location=self.location,
+                    links=self._parse_links(item,response),
+                    source=self._parse_source(response),
+                )
+                yield meeting
+                meeting["status"] = self._get_status(meeting)
+                meeting["id"] = self._get_id(meeting)
+
+            
+
+    def _parse_meeting_types(self, response):
+        """ Parse meeting titles and start times. Returns a list of tuples including titles and meeting times """
+        meeting_details = []
+        meeting_types = []
+        for item in response.css('#content .border-t li').extract(): #interate through meeting times and types
+            start_time = re.search('(\d+:\d{1,2}\s(?:a.m|p.m))', item)
+            title = re.search('(\d+:\d{1,2}\s(?:a.m.|p.m.)\s-\s)(.+)$', item)
+            if ("Board" not in item or "board" not in item):
+                meeting_classification = COMMITTEE
+            else:
+                meeting_classification = BOARD
+            meeting_types.append([title,start_time,meeting_classification])
+        return meeting_types
+    
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return "See source to confirm details. Times may change"
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False 
+
+    def _validate_location(self,response):
+        if "Room BC" not in response.xpath('//*[@id="content"]/div[1]/div/p[1]'):
+            raise ValueError("The location has changed! Please check source!")
+
+    def _parse_links(self, item, response):
+        """Parse or generate links."""
+        links = {}
+        href = response.xpath('//*[@id="content"]/div[2]').re('\/meetings\/\d{4}\/\d{2}-\d{2}').pop()
+        title = item + " Meeting"
+        links.update({href: title})
+        if len(response.css('#content .lg\:flex li a').extract()) > 0: #scape meeting notes
+            next_page = "http://bog.wayne.edu/" + href 
+            documents = re.search('//*[@id="content"]/div[2]/ul', next_page)
+            for items in documents:
+                title = documents.xpath('//a/text').pop()
+                href = documents.xpath('//a/@href').pop()
+                links.update({href: title})
+        return links
+
+    def _parse_source(self, response):
+        """Parse or generate source."""
+        return "http://bog.wayne.edu/"
--- a/tests/test_wayne_state_university.py
+++ b/tests/test_wayne_state_university.py
@@ -1,0 +1,78 @@
+from datetime import datetime
+from os.path import dirname, join
+
+import pytest
+from city_scrapers_core.constants import BOARD, COMMITTEE
+from city_scrapers_core.utils import file_response
+from freezegun import freeze_time
+
+from city_scrapers.spiders.wayne_state_university import WayneStateUniversitySpider
+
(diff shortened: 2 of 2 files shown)


city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        
+        #one large all encompassing meeting
+        meeting_dates = response.xpath('//*[@id="content"]/div[1]/div/h2').re('((?:January|Feburary|March|April|May|June|July|August|October|September|November|December)\s+\d{1,2},\s+\d{4})')
```
pjsier: An easier way of pulling this would be to look at all of the `li` elements in the bottom section which you can get with:

```python
response.css(".px-4 ul > li")
```

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+        for item in meeting_dates:
```
pjsier: When there's a link in the `li` element, you'll want to `yield` a request to another page that parses detail pages (one example here: https://bog.wayne.edu/meetings/2019/09-20)

You can check if the `item` has a link in it by seeing if `len(item.css("a"))` is more than 0

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+            meeting = Meeting(
```
pjsier: For each date you'll want to yield 3 meetings, one for each of the separate categories (since they have distinct start times). You can parse those in the beginning instead of re-running it each time in `_parse_start`.

This could look like a method that does something like this:

```python
def _parse_meeting_types(self, response):
    """Parse meeting titles and start times. Returns a list of tuples including title and start time"""
    meeting_details = []
    for item in response.css("#content .border-t li"):
        title, start_time = self._parse_title_start_time(item) # Placeholder function
        meeting_details.append(title, start_time)
    return meeting_details
```

Then you could iterate first through dates, and then inside that loop create another loop that iterates through each of the meeting types

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+        """Parse start datetime as a naive datetime object."""
+        time = response.xpath('//*[@id="content"]/div[1]/div/ul').re('(\d+:\d{1,2}\s(?:a.m|p.m))')
+        date_time_start = item
+        date_time_end = date + " at " + time[-1]
+
+        return date_time_end
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return "Unless noted otherwise, the time schedule for meetings will be as follows"
```
pjsier: This won't necessarily make sense for someone not looking at the page, so something like "See source to confirm details" might be more clear

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False #no all day meetings to be seen
+
+    def _parse_location(self, item):
```
pjsier: You'll want to use a `_validate_location` function here to verify that it hasn't changed. Here's one example: https://github.com/City-Bureau/city-scrapers/blob/c23f1fd17b1ced1c01dff295c8cdf1ca191f0244/city_scrapers/spiders/chi_ssa_34.py#L58

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+            "name": "495 Gilmour Mall, Detroit, MI 48202", #name does not change 
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        href = response.xpath('//*[@id="content"]/div[2]'.re('\/meetings\/\d{4}\/\d{2}-\d{2}'))
+        title = item + " Meeting"
+        return [{"href": href, "title": title}]
+
+    def _parse_source(self, response):
```
pjsier: You can just return `response.url` here since it would be doing the same thing but a little more explicit (in case there are any redirects in the future)

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,95 @@
...
+        return False #no all day meetings to be seen
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "McGregor Memorial Conference Center", #meeting adresses does not change  
+            "name": "495 Gilmour Mall, Detroit, MI 48202", #name does not change 
+        }
+
+    def _parse_links(self, item):
```
pjsier: This will need to change if you're parsing the meeting detail page, but in general you'll want to include `response` as an argument (like the change in `_parse_start`) and call the `response.urljoin` function on any relative URLs. That way, the return value will be a valid URL that someone could use without having to combine it with the source URL

moisesjsanchez:
Thank you so much for the help so far! However, I just ran into an issue with my git. My current commit is not making to this branch and I have been working on this issue for two days with no prevail. I can try dragging and dropping my current second draft via message on this thread? But if this would be too much I can just leave this project for another user and wait till another issue arises. Thank you once again.

pjsier:
Hmm git issues can be tricky. Here's one resource that can be helpful for debugging: https://sethrobertson.github.io/GitFixUm/fixup.html

Worst case scenario I would say to manually copy the files you've changed out of the repo, re-clone it and then move the files back inside and then commit again. It's not ideal, but it should be relatively straightforward since you've only added a few files here. Happy to help debug on Slack as well!

moisesjsanchez:
Hello. I just made my second draft.  When I test some of these sections in scrapy shell I got most of the functions to work. However, I have an issue with my _validate_location function keep returning the ValueError even through Room BC is found in the xpath?  I also fixed my git issue, I think for the most part.  Thank you for the help once again!

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
...
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return "See source to confirm details. Times may change"
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False 
+
+    def _validate_location(self,response):
+        if "Room BC" not in response.xpath('//*[@id="content"]/div[1]/div/p[1]'):
```
pjsier: Not sure if this is returning a string, you might have to call `extract()`

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
...
+                meeting["id"] = self._get_id(meeting)
+
+            
+
+    def _parse_meeting_types(self, response):
+        """ Parse meeting titles and start times. Returns a list of tuples including titles and meeting times """
+        meeting_details = []
+        meeting_types = []
+        for item in response.css('#content .border-t li').extract(): #interate through meeting times and types
+            start_time = re.search('(\d+:\d{1,2}\s(?:a.m|p.m))', item)
```
pjsier: `start` needs to be a `datetime` object, so you'll need to parse any string you find

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
...
+                    classification= meeting_type[-1], 
+                    start= meeting_type[1],
+                    end= "Ending Time Not Specified",
+                    all_day=self._parse_all_day(item),
+                    time_notes=self._parse_time_notes(item),
+                    location=self.location,
+                    links=self._parse_links(item,response),
+                    source=self._parse_source(response),
+                )
+                yield meeting
```
pjsier: You'll need to `yield` the meeting at the end of the loop once `status` and `id` have been set

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
...
+        self._validate_location(response)
+        for item in response.xpath('//*[@id="content"]/div[1]/div/h2').re('((?:January|Feburary|March|April|May|June|July|August|October|September|November|December)\s+\d{1,2},\s+\d{4})'):
+            meetings = self._parse_meeting_types(response)
+            for meeting_type in meetings: 
+                meeting = Meeting(
+                    title = meeting_type[0],
+                    description=self._parse_description(item),
+                    classification= meeting_type[-1], 
+                    start= meeting_type[1],
+                    end= "Ending Time Not Specified",
```
pjsier: `end` should either return a `datetime` object or `None`. Here's some more context from the docs https://cityscrapers.org/docs/development/#meeting-items

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
+from city_scrapers_core.constants import BOARD, COMMITTEE
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+import re
+
+class WayneStateUniversitySpider(CityScrapersSpider):
+    name = "wayne_state_university"
+    agency = "Wayne State University"
+    timezone = "America/Chicago"
```
pjsier: This should be "America/Detroit"

city_scrapers/spiders/wayne_state_university.py
```diff
@@ -0,0 +1,94 @@
...
+    allowed_domains = ["http://bog.wayne.edu"]
```
pjsier: This is a more recent change, but you can remove this property entirely

pjsier:

PR #885, opened 2019-10-11
Add wayne_land_bank spider

radoslawkrolikowski:
This pull request adds the wayne_land_bank spider, the test file of that spider and target web page as an HTML file. 

-  To access the Board of Directors tab on the following page: https://public-wclb.epropertyplus.com/landmgmtpub/app/base/customPage we have to start with POST request.

- The desired data was included in `<script>` tags, thus to extract the data, the `re` regular expressions were used.

- The target site seems to be under the development, in the future the spider might require update if the website layer was changed or new information were added.

### DIFF
--- a/city_scrapers/spiders/wayne_land_bank.py
+++ b/city_scrapers/spiders/wayne_land_bank.py
@@ -15,7 +15,7 @@ class WayneLandBankSpider(CityScrapersSpider):
     start_urls = ["https://public-wclb.epropertyplus.com/landmgmtpub/app/base/customPage"]
     location = {
         "name": "Wayne County Treasurer's Office",
-        "address": "400 Monroe St 5th Floor, Detroit, MI 48226, United States",
+        "address": "400 Monroe St 5th Floor, Detroit, MI 48226",
     }
 
     def start_requests(self):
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/wayne_land_bank.py
```diff
@@ -0,0 +1,76 @@
...
+        data = re.findall(
+            '(Jan(?:uary)|Feb(?:ruary)|Mar(?:ch)|Apr(?:il)|May|Jun(?:e)|Jul(?:y)'
+            '|Aug(?:ust)|Sep(?:tember)|Oct(?:ober)|Nov(?:ember)|Dec(?:ember))'
+            r'(.*?)(?=\.\s|\\n)'
+            '(.*?)(?=\\n)', data[0]
+        )
+
+        for item in data:
+            meeting = Meeting(
+                title=item[2].strip(" .()"),
```
pjsier: The title should never be an empty string, if it's blank we should default to "Board of Directors". Since these are all for the Board, any additional title should probably be added to "Board of Directors " so that the special meeting becomes "Board of Directors Special Meeting"

city_scrapers/spiders/wayne_land_bank.py
```diff
@@ -0,0 +1,76 @@
...
+                description="",
+                classification=BOARD,
+                start=self._parse_start("".join(item[0:2])),
+                end=None,
+                all_day=False,
+                time_notes="",
+                location=self._parse_location(""),
```
pjsier: The `_parse_location` method here isn't necessary, but a `_validate_location` method would be good here to check if the location has changed. https://github.com/City-Bureau/city-scrapers/blob/c23f1fd17b1ced1c01dff295c8cdf1ca191f0244/city_scrapers/spiders/chi_ssa_34.py#L58-L60

city_scrapers/spiders/wayne_land_bank.py
```diff
@@ -0,0 +1,76 @@
...
+
+class WayneLandBankSpider(CityScrapersSpider):
+    name = "wayne_land_bank"
+    agency = "Wayne County Land Bank"
+    timezone = "America/Detroit"
+    allowed_domains = ["public-wclb.epropertyplus.com"]
+    start_urls = ["https://public-wclb.epropertyplus.com/landmgmtpub/app/base/customPage"]
+    location = {
+        "name": "Wayne County Treasurer's Office",
+        "address": "400 Monroe St 5th Floor, Detroit, MI 48226, United States",
```
pjsier: "United States" isn't necessary here for our purposes

radoslawkrolikowski:
All changes that you suggested have been made. I am only wondering about the location item. I used default location name and address because it looks clearer than this one on the website. After adding `_validate_location` method, maybe it will be good to use scraped location if default wasn't found.
```
loc = re.findall('The Board of Directors holds meetings at (.*?)(?=\\n)', self.response_data[0])[0]
print('Name: {}, Address: {}'.format(loc.split('at')[0].title(), loc.split('at')[1]))
```
Or raising an error as it is right now is better option?

pjsier:
@radoslawkrolikowski thanks for these changes!  For now, since the markup is so irregular I think it's safest to throw the error if the location isn't found. Reviewing now but I think these are good to go

PR #886, opened 2019-10-11
Remove allowed_domains property

pjsier:
The `allowed_domains` is a an optional property carried over from scrapy defaults, but since our scrapers are pretty constrained it ends up just making things more difficult and requires specifying `dont_filter` in a lot of requests. Removing it also makes it easier to take advantage of duplicate filtering in `scrapy-wayback-middleware`


pjsier:
Merging because the CI failure is due to too much output

PR #900, opened 2019-10-19
0575 spider chi ssa 60

bendiller:
Hi Patrick - 

I believe this is ready for a review. A couple notes/questions (as usual):

1. How many months forward and backward would you like this to look? Currently it's set to capture a full year (10 months backward, this month, next month) but this is trivial to change however you prefer.
2. The `meeting["description"]` field contents are probably too noisy to be useful, but I built the functionality real quick so you can make the decision. Let me know if you'd like me to take it out.
3. I didn't find any links for minutes or agendas. Let me know if you see any you'd like me to process.
4. I don't know if I've set the classifications correctly. I...seem not to be very good at that, haha.

Anyway, I'll be happy to make any changes you like of course. Thanks!


city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+
+class ChiSsa60Spider(CityScrapersSpider):
+    name = "chi_ssa_60"
+    agency = "Chicago Special Service Area #60 Albany Park"
+    timezone = "America/Chicago"
+    months_backward = 10
+    months_forward = 1
+
+    start_urls = None  # Generated by `start_requests`
+    custom_settings = {'ROBOTSTXT_OBEY': False}
```
pjsier: It looks like we shouldn't need this if we aren't scraping anything at /wp-admin/?

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class ChiSsa60Spider(CityScrapersSpider):
+    name = "chi_ssa_60"
+    agency = "Chicago Special Service Area #60 Albany Park"
+    timezone = "America/Chicago"
+    months_backward = 10
+    months_forward = 1
+
+    start_urls = None  # Generated by `start_requests`
```
pjsier: If you're using `start_requests` I think you can remove this

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def _parse_classification(self, title):
+        """Parse or generate classification from allowed options."""
+        title = title.lower()
+        if 'advisory' in title:
+            return ADVISORY_COMMITTEE
+        elif 'committee' in title:
+            return COMMITTEE
+        # elif 'economic' in title or 'environmental' in title:
+        elif any(s in title for s in ['economic', 'environmental', 'membership', 'commission']):
```
pjsier: The classifications look good overall! I think we can just return `COMMISSION` here as a default though

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+from city_scrapers_core.spiders import CityScrapersSpider
+from dateutil.relativedelta import relativedelta
+from scrapy import Request
+
+
+class ChiSsa60Spider(CityScrapersSpider):
+    name = "chi_ssa_60"
+    agency = "Chicago Special Service Area #60 Albany Park"
+    timezone = "America/Chicago"
+    months_backward = 10
```
pjsier: This isn't based on anything in particular, but let's say 6 months backward and 3 months forward?

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        event_list = list()
+        for item in response.xpath("//script[@type='application/ld+json']/text()"):
```
pjsier: You could do something like `extract_first` or just using the first index here instead of iterating through just to make it more clear that it should only find one thing

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        allowed = ['meeting', 'committee', 'advisory council']
+        if not any([a in item['name'].lower() for a in allowed]):  # Not a meeting type we want
+            return None
+        return item['name']
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        desc = html.unescape(item["description"])
```
pjsier: I think this is good to leave in. Just to be safe here in case they add any more HTML in, it might be easier to create a `Selector` from the description and extract all the text inside of it like we do here: https://github.com/City-Bureau/city-scrapers/blob/20a12ba5d76186cba65b45f7f764f02393d4a991/city_scrapers/spiders/cook_landbank.py#L59

I'm also seeing some miscellaneous whitespace characters like `\xa0` in the output, so it might make sense to replace all whitespace characters with a single space with `re.sub(r"\s+", " ", string).strip()`

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+            return NOT_CLASSIFIED
+
+    def _parse_time(self, item, index):
+        # index must be either 'startDate' or 'endDate'
+        return datetime.strptime(item[index][:-6], "%Y-%m-%dT%H:%M:%S")
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        return self._parse_time(item, 'startDate')
+        # return None
```
pjsier: Just a note to remote this comment

city_scrapers/spiders/chi_ssa_60.py
```diff
@@ -0,0 +1,129 @@
...
+        elif any(s in title for s in ['economic', 'environmental', 'membership', 'commission']):
+            return COMMISSION
+        else:
+            return NOT_CLASSIFIED
+
+    def _parse_time(self, item, index):
+        # index must be either 'startDate' or 'endDate'
+        return datetime.strptime(item[index][:-6], "%Y-%m-%dT%H:%M:%S")
+
+    def _parse_start(self, item):
```
pjsier: If `_parse_start` and `_parse_end` are essentially the same thing, it might make sense to just call `_parse_time` twice in the `parse` method. The methods are just a starting point, so if some of them are redundant it's fine to consolidate them

tests/test_chi_ssa_60.py
```diff
@@ -0,0 +1,77 @@
...
+    test_str = "You are invited to join North River Commission and the Albany Park Chamber of"
+    assert parsed_items[0]["description"].startswith(test_str)
+
+
+def test_start():
+    assert parsed_items[0]["start"] == datetime(2019, 10, 1, 19, 0)
+
+
+def test_end():
+    print(parsed_items[0]["end"])
```
pjsier: A quick note to remove this

city_scrapers/spiders/chi_ssa_60.py
bendiller: Whoops, good catch. That's leftover from my first approach (used an AJAX call but that turned out to be a bad idea).

city_scrapers/spiders/chi_ssa_60.py
bendiller: Good point.

city_scrapers/spiders/chi_ssa_60.py
bendiller: Deal!

city_scrapers/spiders/chi_ssa_60.py
bendiller: Well, I should've mentioned that looking forward, only October and November are populated on the site currently, so I think we may wind up with some fruitless requests being made by looking 3 months ahead. Probably not a problem, and I'll go ahead and implement it this way, but I figured I should let you know in case you want something different.

city_scrapers/spiders/chi_ssa_60.py
bendiller: That's a good idea.

city_scrapers/spiders/chi_ssa_60.py
bendiller: Hmm, yeah my approach definitely felt clunky. This sounds a lot better. If I'm understanding you properly though, that would also depend on every description being contained within the `<p></p>` tags right? I don't remember off hand if they all were, but I'll take a look.

city_scrapers/spiders/chi_ssa_60.py
bendiller: Sounds good! I wasn't sure if I should prioritize uniformity between spiders or just sensible code in general.

tests/test_chi_ssa_60.py
bendiller: Whoops! Will do.

city_scrapers/spiders/chi_ssa_60.py
pjsier: It should work without any tags if you just check pull everything out with:

```python
" ".join(item.css("*::text").extract())
```

bendiller: Hmm, that worked really well. I'm going to have to dive into the CSS selectors someday, I haven't really ever used them. Thanks!

bendiller:
I believe I've addressed all of your concerns. Let me know if there's anything I missed or any other requests!

city_scrapers/spiders/chi_ssa_60.py
pjsier: I'm partially just more used to them from doing front-end work, but I think they're a lot easier to use and I'd definitely recommend it!

pjsier:
Looks great! Merging now

Issue #912, opened 2019-10-24
Spider: Illinois Arts Council


pjsier:
URL: http://www.arts.illinois.gov/about-iac/governance/council-meetings
Spider Name: il_arts_council
Agency Name: Illinois Arts Council
Meeting Name: Agency Board

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

channiemills:
I'll take this one!

pjsier:
@channiemills sounds good!

pjsier:
Closed by #916

PR #916, opened 2019-10-29
Issue 912: il_arts_council

channiemills:
PR for https://github.com/City-Bureau/city-scrapers/issues/912. 

I did have a question, how should I handle `cancelled` meetings? The [website](http://www.arts.illinois.gov/about-iac/governance/council-meetings) uses the location field to indicate whether a meeting is canceled or not but I wasn't sure if there was a different field I should use on the meeting object to indicate that.

### DIFF
--- a/city_scrapers/spiders/il_arts_council.py
+++ b/city_scrapers/spiders/il_arts_council.py
@@ -35,7 +35,8 @@ class IlArtsCouncilSpider(CityScrapersSpider):
                     source=self._parse_source(response),
                 )
 
-                meeting["status"] = self._get_status(item, text=item.xpath('td[2]/text()').get())
+                meeting_status = item.xpath('td[2]/text()').get() or ""
+                meeting["status"] = self._get_status(meeting, text=meeting_status)
                 meeting["id"] = self._get_id(meeting)
 
                 yield meeting
--- a/city_scrapers/spiders/il_arts_council.py
+++ b/city_scrapers/spiders/il_arts_council.py
@@ -89,8 +89,7 @@ class IlArtsCouncilSpider(CityScrapersSpider):
             location = item.xpath('td[2]/a/text()').get()
 
         jrtc = {
-            "address":
-                "100 West Randolph, Suite 10-500, Chicago, IL 60601",
+            "address": "100 West Randolph, Suite 10-500, Chicago, IL 60601",
             "name": "IACA/JRTC"
         }
 
--- a/city_scrapers/spiders/il_arts_council.py
+++ b/city_scrapers/spiders/il_arts_council.py
@@ -35,7 +35,8 @@ class IlArtsCouncilSpider(CityScrapersSpider):
                     source=self._parse_source(response),
                 )
 
-                meeting["status"] = self._get_status(meeting)
+                meeting["status"] = self._get_status(item,
+                                                     text=item.xpath('td[2]/text()').get())
                 meeting["id"] = self._get_id(meeting)
 
                 yield meeting
--- a/city_scrapers/spiders/il_arts_council.py
+++ b/city_scrapers/spiders/il_arts_council.py
@@ -35,8 +35,7 @@ class IlArtsCouncilSpider(CityScrapersSpider):
                     source=self._parse_source(response),
                 )
 
-                meeting["status"] = self._get_status(item,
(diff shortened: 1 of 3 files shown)


pjsier:
@channiemills thanks for the PR! You can pass the string of the location (or wherever might mention a cancellation) to the `_get_status` function in the `text` kwarg. More info here https://cityscrapers.org/docs/development/#status

It looks like the check for `isort` is failing for now though, but running the linting commands locally should fix that

city_scrapers/spiders/il_arts_council.py
```diff
@@ -0,0 +1,129 @@
...
+                meeting["status"] = self._get_status(meeting)
+                meeting["id"] = self._get_id(meeting)
+
+                yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Board Meeting"
+
+    def _parse_description(self, item):
```
pjsier: We can leave this blank if they aren't providing any info in the description

city_scrapers/spiders/il_arts_council.py
```diff
@@ -0,0 +1,129 @@
...
+                )
+
+                meeting["status"] = self._get_status(meeting)
+                meeting["id"] = self._get_id(meeting)
+
+                yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Board Meeting"
```
pjsier: "Agency Board" might be the best default here. We try to leave "Meeting" out if we can

city_scrapers/spiders/il_arts_council.py
```diff
@@ -0,0 +1,130 @@
...
+    def _parse_location(self, item):
+        """Parse or generate location."""
+
+        location = item.xpath('td[2]/text()').get()
+        if not location:
+            location = item.xpath('td[2]/a/text()').get()
+
+        jrtc = {
+            "address":
+                "James R. Thompson Center, 100 West Randolph, Suite 10-500, Chicago, IL 60601",  # noqa
```
pjsier: We can take out the "James R. Thompson Center" part here and just start with "100 W"

city_scrapers/spiders/il_arts_council.py
```diff
@@ -0,0 +1,130 @@
...
+                "name": "",
+            }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        agenda_link = item.xpath('td/a/@href').get()
+        if agenda_link:
+            title = agenda_link.split('/')[-1].replace('%20', ' ')
+            return [{"href": agenda_link, "title": title}]
+        return [{"href": "", "title": ""}]
```
pjsier: If there aren't any links available this should just return an empty list

channiemills:
Thanks for the feedback, back to you @pjsier

city_scrapers/spiders/il_arts_council.py
```diff
@@ -0,0 +1,129 @@
...
+
+        else:
+            return {
+                "address": "",
+                "name": "",
+            }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        agenda_link = item.xpath('td/a/@href').get()
```
pjsier: I missed this on the first pass, but it looks like some of the URLs being returned are relative (seeing it on the August 2018 meeting). To be safe, we should add `response` as an argument to this function and call `response.urljoin` on the parsed link to make sure that it includes the domain

tests/test_il_arts_council.py
```diff
@@ -0,0 +1,77 @@
...
+    assert parsed_items[0]["location"] == {'address': 'TBA', 'name': 'TBA'}
+
+
+def test_source():
+    assert parsed_items[0][
+        "source"] == "http://www.arts.illinois.gov/about-iac/governance/council-meetings"  # noqa
+
+
+def test_links():
+    print(parsed_items[8]["links"])
```
pjsier: This can be removed

channiemills: 🙈

city_scrapers/spiders/il_arts_council.py
channiemills: Good catch, thanks.

pjsier:
@channiemills looks great, thanks! Merging now

Issue #913, opened 2019-10-24
Spider: Illinois Capital Development Board


pjsier:
URL: https://www2.illinois.gov/cdb/about/boardmeetings
Spider Name: il_capital_development
Agency Name: Illinois Capital Development Board

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

codevbus:
@pjsier  - I'd love to take this.

pjsier:
@codevbus all yours!

pjsier:
Closed by #917, thanks @codevbus!

PR #917, opened 2019-10-30
0913: Spider for issue #913

codevbus:
@pjsier - Everything should be g2g here.

### DIFF
--- a/city_scrapers/spiders/il_capital_development.py
+++ b/city_scrapers/spiders/il_capital_development.py
@@ -21,31 +21,22 @@ class IlCapitalDevelopmentSpider(CityScrapersSpider):
         for item in response.xpath("//tbody/tr"):
             self._validate_location(item)
             meeting = Meeting(
-                title="CDB Board Meeting",
-                description=self._parse_description(item),
+                title="Capital Development Board",
+                description="",
                 classification=BOARD,
                 start=self._parse_start(item),
                 end=self._parse_end(item),
                 all_day=self._parse_all_day(item),
-                time_notes=self._parse_time_notes(item),
+                time_notes="",
                 location=self.location,
                 links=self._parse_links(item),
                 source=self._parse_source(response),
             )
 
-            meeting["status"] = self._get_status(meeting)
+            meeting["status"] = self._get_status(meeting, text=item.xpath('.//td/text()').get())
             meeting["id"] = self._get_id(meeting)
 
             yield meeting
-
-    def _parse_description(self, item):
-        """Parse or generate meeting description."""
-        description = item.xpath('.//td/text()').get()
-        for i in ["Chicago,", "Chicago", "Springfield", "&", "Collinsville"]:
-            if i in description:
-                description = description.replace(i, '')
-        description = description.lstrip()
-        return description
 
     def _parse_start(self, item):
         """Parse start datetime as a naive datetime object."""
@@ -62,14 +53,6 @@ class IlCapitalDevelopmentSpider(CityScrapersSpider):
         end += timedelta(days=1)
         return end
 
-    def _parse_time_notes(self, item):
-        """Parse any additional notes on the timing of the meeting"""
-        time_notes = item.xpath('.//td/text()').get()
-        if any(x in time_notes.lower() for x in ["rescheduled", "cancelled", "changed"]):
-            return time_notes
-        else:
-            return ""
-
     def _parse_all_day(self, item):
         """Parse or generate all-day status. Defaults to False."""
         return False
@@ -83,7 +66,10 @@ class IlCapitalDevelopmentSpider(CityScrapersSpider):
         """Parse or generate links."""
         links = []
         for href in item.xpath('.//ul/li/a'):
-            links.append({"title": href.xpath('text()').get(), "href": href.xpath('@href').get()})
+            links.append({
+                "title": href.xpath('text()').get().strip(),
+                "href": href.xpath('@href').get().strip()
+            })
         return links
 
     def _parse_source(self, response):
--- a/city_scrapers/spiders/il_capital_development.py
+++ b/city_scrapers/spiders/il_capital_development.py
@@ -1,0 +1,91 @@
+from datetime import datetime, timedelta
+
+from city_scrapers_core.constants import BOARD
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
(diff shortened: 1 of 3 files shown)


pjsier:
@codevbus looks like some of the style checks are failing. If you run [these commands](https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools) and then push the changes that should fix it

codevbus:
> @codevbus looks like some of the style checks are failing. If you run [these commands](https://cityscrapers.org/docs/development/#6-run-linting-and-style-checking-tools) and then push the changes that should fix it

The 3 lines throwing errors are contiguous URL strings that will violate line length even if true-left aligned... wasn't sure if I should split them or what.

pjsier:
Ah gotcha, for URLs that don't make sense to split on multiple lines I just add `  # noqa` at the end of the line to exempt it from the checks. I think that's fine for these

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
```
pjsier: It doesn't look like they're including anything useful here, so I think we can ignore this and return an empty string

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+        "address": "100 West Randolph Street, 14th Floor, Chicago, IL 60601"
+    }
+
+    def parse(self, response):
+        """
+        """
+        for item in response.xpath("//tbody/tr"):
+            self._validate_location(item)
+            meeting = Meeting(
+                title="CDB Board Meeting",
```
pjsier: We try to leave "Meeting" out of titles we're populating statically, so "Capital Development Board" might work better here

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+    agency = "Illinois Capital Development Board"
+    timezone = "America/Chicago"
+    start_urls = ["https://www2.illinois.gov/cdb/about/boardmeetings"]
+    location = {
+        "name": "James R. Thompson Center",
+        "address": "100 West Randolph Street, 14th Floor, Chicago, IL 60601"
+    }
+
+    def parse(self, response):
+        """
```
pjsier: Small thing, but it's probably fine to remove this comment if we aren't using it

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+                start=self._parse_start(item),
+                end=self._parse_end(item),
+                all_day=self._parse_all_day(item),
+                time_notes=self._parse_time_notes(item),
+                location=self.location,
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
```
pjsier: We should pass the text that's currently getting pulled in `description` to the `text` kwarg here which will check for any mention of cancellations.

The `_get_status` function checks for any mention of words like cancel and reschedule, and since the notes about "Time Changed" seem to reflect the current details of the row it looks like `_get_status` should cover it

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        end_date = item.xpath('.//th/text()').get()
+        end_time = "12:00 AM"
+        end = datetime.strptime(end_date + end_time, "%B %d, %Y %H:%M %p")
+        end += timedelta(days=1)
+        return end
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
```
pjsier: The text currently getting pulled here makes more sense as input to `_get_status`. We can leave this as an empty string

city_scrapers/spiders/il_capital_development.py
```diff
@@ -0,0 +1,91 @@
...
+    def _validate_location(self, item):
+        """Validate if location has changed"""
+        if "Chicago" not in item.xpath('.//td/text()').get():
+            raise ValueError("Meeting location has changed")
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = []
+        for href in item.xpath('.//ul/li/a'):
+            links.append({"title": href.xpath('text()').get(), "href": href.xpath('@href').get()})
```
pjsier: We'll want to call `strip()` on each of these since it looks like there's some extra whitespace from the test output

city_scrapers/spiders/il_capital_development.py
codevbus: @pjsier - Should "time notes" and "description" be defined as empty strings directly in the Class attributes or as empty returns at the parse functions?

pjsier: It's probably easiest to just include them as empty strings in the `Meeting` kwargs. The parse methods are mostly there as a starter, so there's no reason to leave them in if they aren't useful

codevbus: @pjsier - when I set the Meeting kwargs to empty strings, and pass the xpath selector text to the `_get_status` `text` kwarg I get this test output:

```tests/test_chi_ssa_28.py:20: in <module>
    parsed_items = [item for item in spider.parse(test_response)]
tests/test_chi_ssa_28.py:20: in <listcomp>
    parsed_items = [item for item in spider.parse(test_response)]
city_scrapers/spiders/chi_ssa_28.py:29: in parse
    meeting["status"] = self._get_status(meeting)
../../.virtualenvs/city-scrapers-j7d0QWF1/lib/python3.7/site-packages/city_scrapers_core/spiders/spider.py:55: in _get_status
    if item["start"] < datetime.now():
E   TypeError: '<' not supported between instances of 'NoneType' and 'FakeDatetime'```

codevbus: ignore my deleted comment, my working branch wasn't clean with other changes.

pjsier: I'm not sure why that would have changed the `start` value if you only changed `time_notes` and `description`. Could you post a snippet of what was changed?

pjsier:
Looks good! Merging

Issue #906, opened 2019-10-24
Spider: Chicago Special Service Area #28 Six Corners


pjsier:
URL: https://sixcorners.com/ssa28
Spider Name: chi_ssa_28
Agency Name: Chicago Special Service Area #28 Six Corners

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

codevbus:
@pjsier - I'll grab this one

pjsier:
Sounds good!

pjsier:
Closed by #921, thanks @codevbus!

PR #921, opened 2019-10-31
0906: Spider for issue #906

codevbus:
@pjsier - ready for review!

### DIFF
--- a/city_scrapers/spiders/chi_ssa_28.py
+++ b/city_scrapers/spiders/chi_ssa_28.py
@@ -31,7 +31,7 @@ class ChiSsa28Spider(CityScrapersSpider):
                 all_day=self._parse_all_day(date),
                 time_notes="",
                 location=self.location,
-                links=self._parse_links(item, date, dates),
+                links=self._parse_links(item, date, dates, response),
                 source=self._parse_source(response),
             )
 
@@ -65,7 +65,7 @@ class ChiSsa28Spider(CityScrapersSpider):
         """Parse or generate all-day status. Defaults to False."""
         return False
 
-    def _parse_links(self, item, date, dates):
+    def _parse_links(self, item, date, dates, response):
         """Parse or generate links. Uses the index from the dates list."""
         links = []
         if dates.index(date) == 0:
@@ -80,7 +80,7 @@ class ChiSsa28Spider(CityScrapersSpider):
             if href:
                 links.append({
                     "title": href.xpath('.//text()').get(),
-                    "href": href.xpath('.//@href').get()
+                    "href": response.urljoin(href.xpath('.//@href').get())
                 })
         return links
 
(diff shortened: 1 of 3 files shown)


codevbus:
Just realized it's stripping some links. Lemme fix.

city_scrapers/spiders/chi_ssa_28.py
```diff
@@ -0,0 +1,89 @@
...
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa28Spider(CityScrapersSpider):
+    name = "chi_ssa_28"
+    agency = "Chicago Special Service Area #28 Six Corners"
+    timezone = "America/Chicago"
+    start_urls = ["https://sixcorners.com/ssa28"]
+    location = {"name": "Portage Arts Lofts", "address": "4041 N. Milwaukee Ave. #302"}
```
pjsier: The address part should include "Chicago, IL 60641" since we're adding it statically

city_scrapers/spiders/chi_ssa_28.py
```diff
@@ -0,0 +1,89 @@
...
+            item = response.xpath('//div[@class="col sqs-col-3 span-3"]/div/div/p[2]')
+            meeting = Meeting(
+                title="Six Corners Association Meeting",
+                description="",
+                classification=self._parse_classification(date),
+                start=self._parse_start(item, date),
+                end=self._parse_end(item, date),
+                all_day=self._parse_all_day(date),
+                time_notes="",
+                location=self.location,
```
pjsier: It would be good to add a `_validate_location` function to the initial response so that we know if it changes. Here's an example: https://github.com/City-Bureau/city-scrapers/blob/20a12ba5d76186cba65b45f7f764f02393d4a991/city_scrapers/spiders/chi_ssa_34.py#L57-L59

city_scrapers/spiders/chi_ssa_28.py
```diff
@@ -0,0 +1,89 @@
...
+        them outside of the main parsing loop, and use the list index to iterate through
+        xpath for things like links.
+        """
+        dates = response.xpath('//div[@class="col sqs-col-3 span-3"]/div/div/p[2]/text()').getall()
+        dates = [d.strip(', ')[:-2] for d in dates]
+        dates = [d for d in dates if d]
+        for date in dates:
+            item = response.xpath('//div[@class="col sqs-col-3 span-3"]/div/div/p[2]')
+            meeting = Meeting(
+                title="Six Corners Association Meeting",
```
pjsier: Just "Commission" is probably fine here

city_scrapers/spiders/chi_ssa_28.py
```diff
@@ -0,0 +1,89 @@
...
+                description="",
+                classification=self._parse_classification(date),
+                start=self._parse_start(item, date),
+                end=self._parse_end(item, date),
```
pjsier: We can just include `None` here and let the pipeline fill in the value

codevbus: @pjsier  - which value specifically?

pjsier: The value of `end`, you can get rid of the `_parse_end` function and just include `end=None` for the kwarg

Issue #846, opened 2019-10-03
Spider: Chicago Special Service Area #72 Austin


pjsier:
URL: http://www.av72chicago.com/commissioners--meetings.html
Spider Name: `chi_ssa_72`
Agency Name: Chicago Special Service Area #72 Austin

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

doubledare704:
I'll take this

pjsier:
Weird, it was up yesterday. You can use this Wayback Machine link for testing until it's back up https://web.archive.org/web/20190929023807/http://www.av72chicago.com/commissioners--meetings.html

channiemills:
I'll work on this one!

pjsier:
@doubledare704 are you still working on this? If not you're good to go @channiemills, looks like the site is back up

pjsier:
Closed by #922, thanks @channiemills!

PR #922, opened 2019-11-01
0846 spider chi_ssa_72

channiemills:
Addresses https://github.com/City-Bureau/city-scrapers/issues/846

### DIFF
--- a/tests/test_chi_ssa_72.py
+++ b/tests/test_chi_ssa_72.py
@@ -1,0 +1,82 @@
+from datetime import datetime
+from os.path import dirname, join
+
+import pytest
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.utils import file_response
+from freezegun import freeze_time
+
+from city_scrapers.spiders.chi_ssa_72 import ChiSsa72Spider
+
+test_response = file_response(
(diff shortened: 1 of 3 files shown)


tests/test_chi_ssa_72.py
```diff
@@ -0,0 +1,82 @@
...
+def test_time_notes():
+    assert parsed_items[0]["time_notes"] == ""
+
+
+def test_id():
+    assert parsed_items[0]["id"] == "chi_ssa_72/201904221000/x/advisory_commission"
+
+
+def test_status():
+    assert parsed_items[0]["status"] == "cancelled"
```
pjsier: We should use the `CANCELLED` constant here on the off chance that it changes

tests/test_chi_ssa_72.py
```diff
@@ -0,0 +1,82 @@
...
+        "name": "Westside Health Authority",
+    }
+
+
+def test_source():
+    assert parsed_items[0]["source"] == "http://www.av72chicago.com/commissioners--meetings.html"
+
+
+def test_links():
+    print(parsed_items[1])
```
pjsier: Just a small thing to remove

Issue #873, opened 2019-10-07
Spider: Chicago Committee on Standards and Tests


pjsier:
URL: https://www.chicago.gov/city/en/depts/bldgs/supp_info/committee_on_standardsandtests.html
Spider Name: `chi_standards_tests`
Agency Name: Chicago Committee on Standards and Tests

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

zkunkle:
I'll work on it

travfu:
I'd like to have a go at this

pjsier:
@travfu all yours!

travfu:
@pjsier this url contains pdfs, are there other spiders that also parse pdf so I can see what dependencies were used instead of adding new ones.

pjsier:
We've been using PyPDF2, but I don't think you need to parse the PDFs here. Since it's rare that they post things in advance, we can verify that the 1:30 time hasn't changed and include a note in `time_notes` to confirm details with the agency

pjsier:
Closed by #926, thanks @travfu!

PR #926, opened 2019-11-04
Add 0873 chi_standards_tests

travfu:
Issue #873 - Chicago Committee on Standards and Tests 

Added parser and tests for website stated in issue.

### DIFF
--- a/city_scrapers/spiders/chi_standards_tests.py
+++ b/city_scrapers/spiders/chi_standards_tests.py
@@ -20,7 +20,7 @@ class ChiStandardsTestsSpider(CityScrapersSpider):
         needs.
         """
         for item in response.xpath("//div[@class='col-xs-12']/table[1]//td/p"):
-            if self._pass_filter(item) is False:
+            if not self._pass_filter(item):
                 continue
             meeting = Meeting(
                 title="Committee on Standards and Tests",
@@ -34,7 +34,7 @@ class ChiStandardsTestsSpider(CityScrapersSpider):
                     "address": "121 North LaSalle Street, Room 906, Chicago, IL 60602",
                     "name": "City Hall"
                 },
-                links=self._parse_links(item),
+                links=self._parse_links(item, response),
                 source=response.url,
             )
             meeting["status"] = self._get_status(meeting)
@@ -70,13 +70,13 @@ class ChiStandardsTestsSpider(CityScrapersSpider):
         time_obj = datetime.time(13, 30, 0)
         return datetime.datetime.combine(date_obj, time_obj)
 
-    def _parse_links(self, item):
+    def _parse_links(self, item, response):
         """Parse or generate links."""
         links = []
         a_tag = item.xpath('.//a')
         if a_tag:
             links.append({
-                "href": 'chicago.gov' + a_tag.xpath('.//@href').get(),
+                "href": response.urljoin(a_tag.xpath('.//@href').get()),
                 "title": a_tag.xpath('.//text()').get(),
             })
         return links
--- a/city_scrapers/spiders/chi_standards_tests.py
+++ b/city_scrapers/spiders/chi_standards_tests.py
@@ -4,11 +4,14 @@ from city_scrapers_core.constants import COMMITTEE
 from city_scrapers_core.items import Meeting
 from city_scrapers_core.spiders import CityScrapersSpider
 
+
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/chi_standards_tests.py
```diff
@@ -0,0 +1,82 @@
...
+    ]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.xpath("//div[@class='col-xs-12']/table[1]//td/p"):
+            if self._pass_filter(item) is False:
```
pjsier: `if not self._pass_filter(item)` might be a little cleaner here

city_scrapers/spiders/chi_standards_tests.py
```diff
@@ -0,0 +1,82 @@
...
+            meeting = Meeting(
+                title="Committee on Standards and Tests",
+                description="",
+                classification=COMMITTEE,
+                start=self._parse_start(item),
+                end=None,
+                all_day=False,
+                time_notes="Confirm details with the agency",
+                location={
+                    "address": "121 North LaSalle Street, Room 906, Chicago, Illinois 60602",
```
pjsier: Small thing, but we've been using the state abbreviation `IL` in places where we're statically setting the address. It might also be worth validating that it occurs in room 906 with a `_validate_location` function https://cityscrapers.org/docs/development/#location

city_scrapers/spiders/chi_standards_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        ]
+        date_str = item.xpath('.//text()').get()
+        if any(date_str.lower().startswith(month) for month in months):
+            return True
+        else:
+            return False
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        year = item.xpath(".//../../../../../p[1]/strong/text()").get()
```
pjsier: This might be easier to pull out in `parse` and pass to this function as an argument

city_scrapers/spiders/chi_standards_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        time_obj = datetime.time(13, 30, 0)
+        return datetime.datetime.combine(date_obj, time_obj)
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = []
+        a_tag = item.xpath('.//a')
+        if a_tag:
+            links.append({
+                "href": 'chicago.gov' + a_tag.xpath('.//@href').get(),
```
pjsier: Passing `response` as an argument to this function and then using `response.urljoin` here is a cleaner way of doing this that maintains the protocol part of the link

city_scrapers/spiders/chi_standards_tests.py
```diff
@@ -0,0 +1,82 @@
...
+        "https://www.chicago.gov/city/en/depts/bldgs/supp_info/committee_on_standardsandtests.html"
+    ]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.xpath("//div[@class='col-xs-12']/table[1]//td/p"):
```
pjsier: It looks like this isn't parsing years before 2019. We might not always want to pull all meetings, but we should at least pull the past two years if available and be explicit about where this is being set. It would be good to pull those if possible

city_scrapers/spiders/chi_standards_tests.py
travfu: Understandable, but the website does not include an address for the meetings. The statically set address here was deduced from the the few post-meeting pdfs provided on the website. I scanned through all the provided pdfs and they all had the same address, forming the basis of my assumption that the meetings are all held at that location.

As far as I can tell, the website provides insufficient information to parse or validate the meeting locations ahead of time. 

If there is something that I have missed in parsing/validating meeting locations, please do let me know. Otherwise, I don't know how else I can parse the location and will proceed with updating the state abbreviation.

pjsier: Makes sense, in that case just updating the abbreviation is fine

Issue #681, opened 2019-02-03
Spider: Illinois Sports Facilities Authority


pjsier:
URL: https://www.isfauthority.com/governance/board-meetings/
Spider Name: `il_sports_facilities_authority`
Agency Name: Illinois Sports Facilities Authority

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

jamesarnall:
If unclaimed, I'll take this one on. Let me know if there's anything special about this particular site I need to know.

pjsier:
It's all yours! The start times of past meetings aren't listed so it's probably safest to default to 10AM from a quick look, but otherwise it seems pretty straightforward

chandranathr:
Can I help with this one?

pjsier:
@chandranathr it's been inactive for a while, so go for it!

chandranathr:
Due to current personal issues, I do not have the time to work on this bug. Apologies for your time.

zybex86:
If unclaimed I'll have a go ;)

pjsier:
@zybex86 go for it!

zybex86:
Almost finished it - need to test it, as the author of the site has done some CRAZY HTML bindings... Also, should I scrape the most recent meetings? There is a meeting not mentioned in the list at the bottom...

pjsier:
@zybex86 great! We should scrape the past few years of meetings here, are there any meetings you're referring to in particular? Might be worth adding a check in the loop that the meeting is within the past 2 years or so before yielding it though

zybex86:
I'm referring to the 2-3 meetings in the h2 tags that start with days of the week... They are written just before the `meetings and agendas`. The others are ones in this year and prior. Also the full address is in those above (as it implies that sometimes the meeting is in a different place, but I think the extra files point to that)

pjsier:
Yeah, I think those meetings are a high priority, and the address being at the top of the page is pretty common. We've used [`_validate_location`](https://cityscrapers.org/docs/development/#location) functions for that in the past to make sure that it hasn't changed

zybex86:
I think I finished this. Can sb check the issue?

d-glenz:
Can I help with this spider?

pjsier:
@d-glenz all yours!

pjsier:
Closed by #985, thanks @d-glenz!

PR #928, opened 2019-11-15
Issue #681 Spider

zybex86:
Worked on this for a while now, but am at a standstill, as wrongly parse the site. Also need some help on the regex to correctly parse the site. It is styled in a very specific way...


pjsier:
Is there anything in particular you're looking for help with? It looks like you're generally going in the right direction

zybex86:
I am a little tired after this past two weeks coding at my job so can't seem to think of a way to get the right content out of this website. Also, I noticed that they changed the styling yet again :/ That sucks...

zybex86:
still working on it - just sat down to it :)

zybex86:
now it seems to be ok... still need to escape the None errors and parse the meetings above the <h2>Meetings and Agendas</h2>... and fix the broken tests apparently...

zybex86:
ok this seems ok now - just need to figure out the part above meeting and agendas... Gonna do two parses I think :)

zybex86:
The new site has fixed it's formatting (good old Wordpress). Now just need to do a first parse of the elements above the meetings and agendas :)

zybex86:
Done ;)

zybex86:
Did anyone check this?

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,104 @@
...
+                source=response.url,
+            )
+            try:
+                meeting['status'] = self._get_status(meeting)
+                meeting['id'] = self._get_id(meeting)
+            except TypeError:
+                continue
+            yield meeting
+
+    def _parse_title(self, item):
```
pjsier: You can probably consolidate this and `_parse_h2_title`, and "Meeting" can be removed from the end. It might look a little cleaner to return "Board of Directors" if only "Board" is present

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,104 @@
...
+    def _parse_location(self, item):
+        """Parse the location of the meeting."""
+        return {
+            'name': '',
+            'address':
+                item.css('::text').re_first(r'\b \d* [a-zA-Z].*, [0-9][a-zA-Z].*').lstrip(' ')
+        }
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
```
pjsier: Since the time is sometimes included, it might make sense to consolidate this with `_parse_h2_start` to use any time that's available in the string and defaulting to 10am otherwise

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,104 @@
...
+
+        for item in response.css('div.wpb_text_column div.wpb_wrapper p'):
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description='',
+                classification=BOARD,
+                start=self._parse_start(item),
+                end=None,
+                all_day=False,
+                time_notes='',
```
pjsier: Since we're using default times we should add a note like "See agenda for meeting time"

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        for item in response.css('div.wpb_text_column div.wpb_wrapper div.inner-text h2'):
```
pjsier: In general, it looks like most of the logic between the H2 and general items is the same, so if possible it would be good to consolidate these loops

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,104 @@
...
+            meeting = Meeting(
+                title=self._parse_h2_title(item),
+                description='',
+                classification=BOARD,
+                start=self._parse_h2_start(item),
+                end=None,
+                all_day=False,
+                time_notes='',
+                location=self._parse_location(item),
```
pjsier: A `_validate_location` function might make more sense here, more info in the [docs](https://cityscrapers.org/docs/development/#location)

city_scrapers/spiders/il_sports_facilities_authority.py
zybex86: You are right - titles are similar. Consolidated them

city_scrapers/spiders/il_sports_facilities_authority.py
zybex86: I consolidated them, and I think the times are ok

city_scrapers/spiders/il_sports_facilities_authority.py
zybex86: added it

city_scrapers/spiders/il_sports_facilities_authority.py
zybex86: you mean add items to one list from both responses and then parse?

pjsier: Or even just modify the loop selector to:

```python
for item in response.css('div.wpb_text_column div.wpb_wrapper div.inner-text h2, div.wpb_text_column div.wpb_wrapper p'):
    meeting = Meeting(...
```

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,89 @@
...
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse meeting title."""
+        title = item.css('::text').re_first(r'.*[a-zA-Z] Meeting')
+        if title and 'Board' in title:
+            return "Board of Directors"
+        return title
+
+    def _parse_location(self, item):
```
pjsier: Looks like this can be removed if it isn't present, but we'll still need a `_validate_location` function so that we can flag when the address might have changed

zybex86: pushed changes but didn't have time to finish last time. Now it's done :)

city_scrapers/spiders/il_sports_facilities_authority.py
zybex86: done :)

zybex86:
I think done all - Thanks for the opportunity and for the wait :) Learned a lot!

city_scrapers/spiders/il_sports_facilities_authority.py
```diff
@@ -0,0 +1,79 @@
...
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description='',
+                classification=BOARD,
+                start=self._parse_start(item),
+                end=None,
+                all_day=False,
+                time_notes='See agenda for meeting time',
+                location=self.location,
+                links=self._parse_links(item),
```
pjsier: From the output of the CI run it looks like all of the links are getting associated with one meeting: https://github.com/City-Bureau/city-scrapers/runs/365507248#step:11:73

Not sure what caused this, but the selector I gave you might have been off

zybex86:
Sorry for not doin anything here - If anyone wants they can take over - I won't have time to do anything here till March...

pjsier:

Issue #672, opened 2019-02-03
Spider: Chicago Northwest Home Equity Assurance Program


pjsier:
URL: https://nwheap.com/category/meet-minutes-and-agendas/
Spider Name: `chi_northwest_home_equity`
Agency Name: Chicago Northwest Home Equity Assurance Program

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

GeorgeDubuque:
I would like to take this one!

pjsier:
@GeorgeDubuque sorry, I missed this initially. Right now our policy is for people to take on one at a time, so feel free to start this or the O'Hare scraper and move on to the other once you're done

mingchan96:
Hi. For a class, my partner and I are looking for an issue to contribute to. I was wondering if this issue is still up for grabs?

pjsier:
@mingchan96 this is open, all yours if you're interested!

ekand:
I'd like to claim this one, please.

mingchan96:
@ekand you can have it. My partner and I are currently business with other projects.

pjsier:
@ekand all yours!

ekand:
@pjsier Thanks!
I'll start by studying the contributors guide and try to have something in a pull request in two weeks.

ekand:
@pjsier Well, I'm sorry to do this again, but I'm going to bow out and release this task. I just got a job (yay!) and I'm going to proritize that for now.

pjsier:
@ekand no problem, and congrats on the job!

SubtleHyperbole:
Hey Pj, so I am working on this one (bc the illinois department of corrections seems to have not been doing what they are supposed to do in terms of posting info about public meetings for the last couple years), and I have a question.

It looks like, in general, the response variable used in the test .py is coming from a method called file_response which pulls a saved offline version of the webpage which was created (i think?) when the spider was created on the command line, leaving no way of pulling additional pages which might be needed to completely parse all meetings.

For the example on this issue (chi_northwest_home_equity, i think), the meetings are listed in pages of 10, with each additional page having a /page/2/ or /page/3/ on up. Normally when I would be scraping a site like this, I would use requests to try get a page, checking its status code looking for a 4## and upon reaching that 4##, to stop the scraper. 

[...]

PR #929, opened 2019-11-18
WIP Spider for issue #672

JKReynolds:
Spider for #672, this me and my partner [mingchan96](https://github.com/mingchan96)'s first. We think it looks good, we wrote some tests, everything passed and the parser results look good.

### DIFF
--- a/city_scrapers/spiders/chi_northwest_home_equity.py
+++ b/city_scrapers/spiders/chi_northwest_home_equity.py
@@ -1,0 +1,365 @@
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from datetime import datetime
+
+
+class ChiNorthwestHomeEquitySpider(CityScrapersSpider):
+    name = "chi_northwest_home_equity"
+    agency = "Chicago Northwest Home Equity Assurance Program"
+    timezone = "America/Chicago"
+    allowed_domains = ["nwheap.com"]
+    start_urls = ["https://nwheap.com/category/meet-minutes-and-agendas/"]
+
+    #used to convert month to number month, help with mm/dd/yyyy formatting
+    calendar = {"january": "01",
+                "february": "02",
+                "march": "03",
+                "april": "04",
+                "may": "05",
+                "june": "06",
+                "july": "07",
+                "august": "08",
+                "september": "09",
+                "october": "10",
+                "november": "11",
+                "december": "12"}
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        #the page contains a list of meetings minute, which are past meetings,
+        #and upcomming meetings
+        #we parsed both sections and return the meetings the best way we can
+        futureMeetings = response.xpath("//aside[@id='em_widget-5']/ul/li[not(@class)]")
+        pastMeetings = response.xpath("//div[@class='post-loop-content']")
+        
+        #loop through the meetings minute (past meetings)
+        for post in pastMeetings:
+            meeting = Meeting(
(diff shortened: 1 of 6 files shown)


city_scrapers/spiders/chi_northwest_home_equity.py
```diff
@@ -0,0 +1,375 @@
...
+
+            # not found then default to 12am
+            if not timeString:
+                timeString = "12:00:00AM"
+            else:
+                timeString = timeString.split(" ")
+                timeString = timeString[0] + ":00" + timeString[1].upper()
+
+            dateString = dateStringFourDigit + " " + timeString
+            datetimeFormat = "%m/%d/%Y %H:%M:%S%p"
```
pjsier: If you're using `%p` here then hours should be pulled with `%I` instead of `%H`

city_scrapers/spiders/chi_northwest_home_equity.py
```diff
@@ -0,0 +1,375 @@
...
+            description = ""
+
+        return description
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        # Northwest Home Equity Assurance Program is governing commission appointed by the mayor
+        return COMMISSION
+
+    def _parse_start(self, item, isPast, response):
```
pjsier: There's a lot going on in these methods, it might be cleaner to split out upcoming and past parsing into two functions, since it's essentially already doing that

JKReynolds: Good call we’ll do that.

city_scrapers/spiders/chi_northwest_home_equity.py
JKReynolds: Thanks, didn’t know that.

JKReynolds:
okay, so we’ll go ahead and work on simplifying the code and grabbing those pdf links. I must say that the logic for parsing the upcoming and past meetings would still need to be sufficiently different enough that it is probably not actually going to simplify the parsing that much to switch to parsing the individual pages. Also, the pages linked to in the sidebar are very different than the ones on the main page. 

Additionally, the past meetings sometimes have posts on the sidebar and on the main page. In these cases there would be extra complexity in avoiding duplicates while maximizing info gathered. This is because there is different information in the sidebar posts than in the main page posts.

Either way, we will simplify, get the pdf links and then show you what we’ve got, thanks for the feedback.

pjsier:
I'm not sure why it would be different? Here's an upcoming meeting page: https://nwheap.com/events/board-meeting-13/ and here's a past meeting page: https://nwheap.com/events/board-meeting-16/, is there a page I'm not looking at?

mingchan96:
We thought we needed to parse the Meet Minutes on (https://nwheap.com/category/meet-minutes-and-agendas/) to get past meetings, since the past meetings sidebar only contain certain number of meetings. 
The past meetings sidebar, not only contain less meetings than the listed Meet Minutes, but also don't contain links to Meet Minutes pdf. For example, the Past Meeting link (https://nwheap.com/events/board-meeting-16/) doesn't contain Meet Minutes pdf. While "Meet Minute" link (https://nwheap.com/2019/10/09/october-10-2019-meeting-minutes-and-agenda/) does contain the pdf link, but doesn't contain meeting location or time.

pjsier:
Right, that's why you start parsing the links and associating them by date which seems to be there, then moving separately to the meetings. You can then use the meeting details to get the related links

mingchan96:
So you want use to start using this the All Events link (https://nwheap.com/events/) and then use Meet Minutes link (https://nwheap.com/category/meet-minutes-and-agendas/) to associate the pdf links? 
Also there are "Special meetings" that are not included in the "All Events" page. Do you want use to create meetings for those as well, the best way possible since there is no direct address associated?

pjsier:
Yes, that's it. It looks like all meetings are at the same location, so it's fine to default there if it's not available. We should try to create a meeting for the special meetings if they aren't listed on the events page

pjsier:

PR #930, opened 2019-11-21
WIP : 0728 Northeastern Illinois University 

bugsz:
I have finished parsing the first website, which include basic description of meetings.
Now the problem is that I don't know how to parse two sites in different structure.
Besides, tests are also unavailable.

I'd like to receive advice from everyone : )


city_scrapers/spiders/chi_housing.py
```diff
@@ -0,0 +1,79 @@
+from city_scrapers_core.constants import NOT_CLASSIFIED
```
pjsier: I'm guessing this was a from going through the documentation, but it should be removed, along with `Thursday`, `scrapy` and the other `chi_housing` files

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
+from city_scrapers_core.constants import BOARD, COMMITTEE, NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+import scrapy
+import datetime
+import re
+
+month_mapping = {
```
pjsier: Generally it's easier to use [`datetime.strptime`](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior) to parse date strings, so we should go that route here as well

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
...
+
+        return {
+            "address": "5500 North St. Louis Avenue, Chicago, Ill., 60625",
+            "name"   : 'Northeastern Illinois University'
+        }
+
+    def _parse_links(self):
+        """Parse or generate links."""
+
+        link_response = scrapy.Request("https://www.neiu.edu/about/board-of-trustees/board-meeting-materials")
```
pjsier: Since you need the materials to associate them to meetings, you should have this URL in `start_urls`, and then once it's parsed `yield` a request to the calendar page. You can see what this looks like in [`cook_housing`](https://github.com/City-Bureau/city-scrapers/blob/75a4d974e4e35fd3cc349c75d1cd68f451b96b9c/city_scrapers/spiders/cook_housing.py#L22-L43)

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
...
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
```
pjsier: This function might change after some refactoring, but generally if we have a location name we want to look up the address and include it here

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
...
+            return COMMITTEE
+        if "meeting" in title.lower():
+            return BOARD
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item,year):
+        """Parse start datetime as a naive datetime object."""
+        pattern = re.compile('([a-z]+)(,|\u00a0|\u0020)+([a-z]+)(,|\u00a0|\u0020)+([0-9]{1,2})',re.IGNORECASE)
+        day, ign1, month, ign2, date = re.findall(pattern,item)[0]
+        date_obj = datetime.datetime(int(year),month_mapping[month],int(date),13,00)
```
pjsier: Mentioned above, but we should use `strptime` instead of parsing the values individually

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def _parse_title(self, item):
+
+        """Parse or generate meeting title."""
+        pattern= re.compile('-(.*(Committee|Meeting))',re.IGNORECASE)
+        title_list = re.findall(pattern, item)
+        if title_list == None:
+            return 'Board of Trustees'
+        title, ign=title_list[-1]
+        return title.replace("<strong>","").replace("</strong>","").replace("\xa0"," ")
```
pjsier: We should avoid handing HTML tags as strings if we can. We have a good amount of examples of parsing out strings with CSS selectors using `::text` and XPath with `/text()`

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,134 @@
...
+
+class ChiNorthernIlUniversitySpider(CityScrapersSpider):
+    name = "chi_northern_il_university"
+    agency = "Northeastern Illinois University"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.neiu.edu/about/board-of-trustees/calendar-of-meetings"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.material_map = dict()
```
pjsier: It can be helpful to use a `defaultdict` here like in this [`cook_housing`](https://github.com/City-Bureau/city-scrapers/blob/75a4d974e4e35fd3cc349c75d1cd68f451b96b9c/city_scrapers/spiders/cook_housing.py#L22-L43) example. That way, you can already create a list of links for each date and just add links to it without having to check whether a key is already present

city_scrapers/spiders/chi_northern_il_university.py
WhatAreYouSaying: Thanks a lot. I'll improve it once I'm available.

WhatAreYouSaying:
I'm running into a problem when I try to write test code for it. I cant find the proper function to call in `parsed_items = [item for item in spider.parse(test_response)]` .Any suggestion on it? @pjsier

pjsier:
@WhatAreYouSaying since you're first parsing the documents page and then the detail page you'll need to go in that order in the tests. Something like:

```python
for item in spider.parse(test_docs_response):
  pass
parsed_items = [item for item in spider._parse_detail(test_response)]
```

Pipfile
```diff
@@ -20,4 +20,4 @@ pytest = "*"
 "flake8" = "*"
 isort = "*"
 yapf = "*"
-
+pylint = "*"
```
pjsier: We aren't planning on using `pylint` so this should be removed

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+
+class ChiNorthernIlUniversitySpider(CityScrapersSpider):
+    name = "chi_northern_il_university"
+    agency = "Northeastern Illinois University"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.neiu.edu/about/board-of-trustees/board-meeting-materials"]
+
+    def parse(self, response):
+        self.link_date_map = defaultdict(list)
+        link_list = response.xpath(
```
pjsier: We don't want to be overly specific in how we search for content in case they change one of the classes, so using CSS selectors instead here will likely be simpler.

Also, since you aren't using `link_list` outside of the loop it might make more sense to just iterate over it directly:

```python
for title in response.css(".field--name-field-generic-body h3"):
```

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+            '//div[@class="field field--name-field-generic-body field--type-text-long \
+field--label-hidden field--item"]/h3'
+        )
+        for title in link_list:
+            full_str = title.xpath(".//text()").extract()[0]
+            date = full_str.split("-")[0].split("(")[0]
```
pjsier: They may change the characters used after this, so checking for a date match with regular expressions might be safer:

```python
date_match = re.search(r"[a-z]{3,9} \d{1,2},? \d{4}", full_str, flags=re.I)
```
That would check for a case-insensitive match that still matches if the comma isn't there, and you can just remove the comma in `strptime` to be safe

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+
+            date_obj = datetime.strptime(date.strip(" "), '%B %d, %Y')
+            new_date_obj = datetime(date_obj.year, date_obj.month, date_obj.day, 13, 00)
```
pjsier: We don't know the time of the meeting, so the `strptime` default of midnight is fine and we shouldn't reassign the time

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+            for one_day_links in title.xpath('.//following-sibling::ul[1]/li'):
+                href = one_day_links.xpath('.//@href').get()
+                name = one_day_links.xpath('.//text()').extract()
+                if href is None:
```
pjsier: If there isn't an `href` attribute we should `continue` because we don't want to scrape empty links

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        items_list = response.xpath(
+            '//div[@class="field field--name-field-generic-body field--type-text-long \
+field--label-hidden field--item"]'
+        )
+        year_list = items_list.xpath('.//h2')
+        for years in year_list.extract():
```
pjsier: The same point here as above, just iterating through with a CSS selector would be simpler

```python
for year_str in response.css(".field--name-field-generic-body h2"):
```

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+            year = re.findall(r'\d{4}', years)[0]
```
pjsier: Checking if `re.search` matches and ignoring otherwise would handle errors better here

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+            items = year_list.xpath('.//following-sibling::ul[1]/li').extract()
```
pjsier: If we don't need this variable we can go straight into the loop

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+                meeting = Meeting(
+                    title=self._parse_title(item),
+                    description='',
+                    classification=self._parse_classification(item),
+                    start=start,
+                    end=None,
+                    all_day=False,
+                    time_notes="See agenda for meeting time",
+                    location=self._parse_location(item),
+                    links=self.link_date_map[start],
```
pjsier: Using `self.link_date_map.get(start, [])` would avoid errors here

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+                    source=self._parse_source(response),
+                )
+                meeting["status"] = self._get_status(meeting)
```
pjsier: We should pass the full text of the `li` item to this as a kwarg since it will set the status to `CANCELLED` if it sees any mention of cancellation in the text

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        items_list = response.xpath(
+            '//div[@class="field field--name-field-generic-body field--type-text-long \
+field--label-hidden field--item"]'
+        )
+        year_list = items_list.xpath('.//h2')
+        for years in year_list.extract():
+            year = re.findall(r'\d{4}', years)[0]
+            items = year_list.xpath('.//following-sibling::ul[1]/li').extract()
+            for item in items:
```
pjsier: We might want to create an `item_str` variable in here to reduce the amount of rework we're doing in parsing out the text. Also, we can do it a little more simply with text selectors and regex

```python
# Combines all text nodes into one string and replaces duplicate spaces
item_str = re.sub(r"\s+", " ", " ".join(item.css("*::text").extract())).strip()
```

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+                    location=self._parse_location(item),
+                    links=self.link_date_map[start],
+                    source=self._parse_source(response),
+                )
+                meeting["status"] = self._get_status(meeting)
+                meeting["id"] = self._get_id(meeting)
+                yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
```
pjsier: This should use the pre-parsed string from the loop

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+        """Parse or generate classification from allowed options."""
+        if 'committee' in title.lower():
+            return COMMITTEE
+        if "meeting" in title.lower():
+            return BOARD
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item, year):
+        """Parse start datetime as a naive datetime object."""
+        date = re.sub(r'<li>|,|\xa0|\*|\*\*|<strong>|</strong>', '', item.split("-")[0]).strip(" ")
```
pjsier: Mentioned above, but this should use the pre-parsed string from the loop and also not replace the time since we don't know it

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+        date_obj = datetime.strptime(date, '%A %B %d')
+        new_date_obj = datetime(int(year), date_obj.month, date_obj.day, 13, 00)
+        return new_date_obj
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
```
pjsier: This and any other functions we aren't using can be removed

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,129 @@
...
+                "name": "Jacob H. Carruthers Center",
+            }
+        if "*" in item:
+            return {"address": "", 'name': "El Centro"}
+        return {
+            "address": "5500 North St. Louis Avenue, Chicago, Ill., 60625",
+            "name": 'Northeastern Illinois University'
+        }
+
+    def _parse_links(self):
```
pjsier: This can be removed if we aren't using it

tests/test_chi_northern_il_university.py
```diff
@@ -0,0 +1,105 @@
...
+freezer = freeze_time("2019-11-15")
+freezer.start()
+
+for item in spider.parse(docs_response):
+    pass
+parsed_items = [item for item in spider._parse_detail(test_response)]
+
+freezer.stop()
+
+# def test_tests():
```
pjsier: The comments from the template can be removed

tests/test_chi_northern_il_university.py
```diff
@@ -0,0 +1,105 @@
...
+#    print("Please write some tests for this spider or at least disable this one.")
+#    assert False
+"""
+Uncomment below
+"""
+
+
+def test_title():
+    assert parsed_items[
```
pjsier: We should parse the reschedule text out of the title

city_scrapers/spiders/chi_northern_il_university.py
WhatAreYouSaying: > For regular board meetings, and unless noted otherwise, Board meetings begin at 1 p.m. and are held at Northeastern Illinois University, 5500 North St. Louis Avenue, Chicago, Ill., 60625. Confirmed meeting locations and times may be found on the agendas for posted meetings on the Board Meeting Materials page.

In fact, regular meeting time is displayed in the website

pjsier: Gotcha, I missed that. It looks like that only applies to board meetings though, so we should only include it there

WhatAreYouSaying:
Unfortunately, I'm not able to finish it immediately due to the final. I'll continue working on it and finish it within a month.Sorry for that. @pjsier

Pipfile
```diff
@@ -18,10 +18,16 @@ pywin32 = {version = "*",sys_platform = "== 'win32'"}
 [dev-packages]
 freezegun = "*"
 pytest = "*"
+
+flake8" = "*"
+isort = "*"
+yapf = "*"
+=======
```
pjsier: Looks like this is your issue, there shouldn't be any changes to the `Pipfile` here, and this is where the error is being thrown

WhatAreYouSaying: I tried to revert to the point when I could pass CI test, but it still didn't work.

pjsier: It looks like there's a separate issue going on with GitHub Actions, but this shouldn't revert the `master` branch's updates to the `Pipfile`

pjsier: Try and copy and paste the contents of the `Pipfile` and `Pipfile.lock` on the current `master` branch and run `pipenv sync --dev` to see if that updates things. I also re-ran the tests so the  output should be more informative

Pipfile
```diff
@@ -13,15 +13,11 @@ legistar = {git = "https://github.com/opencivicdata/python-legistar-scraper"}
 scrapy-wayback-middleware = "*"
 city-scrapers-core = {extras = ["azure"],version = "*"}
 pypdf2 = "*"
-pywin32 = {version = "*",sys_platform = "== 'win32'"}
 
 [dev-packages]
```
pjsier: The Pipfile shouldn't be changed, so `pywin32`, `pathlib2` and the scripts should be in place still and `pylint` shouldn't be added

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,111 @@
...
+                    start=start,
+                    end=None,
+                    all_day=False,
+                    time_notes="See agenda for meeting time",
+                    location=self._parse_location(item),
+                    links=self.link_date_map.get(start, []),
+                    source=response.url,
+                )
+                if "Rescheduled" in meeting["title"]:
+                    meeting["status"] = "Rescheduled"
```
pjsier: `status` should only come from the list of [valid `status` constants](https://cityscrapers.org/docs/development/#status), and `_get_status` handles the logic of making sure a meeting's status is `CANCELLED` if the word `rescheduled` is in the title

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,111 @@
...
+    agency = "Northeastern Illinois University"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.neiu.edu/about/board-of-trustees/board-meeting-materials"]
+
+    def parse(self, response):
+        self.link_date_map = defaultdict(list)
+        for title in response.css(".field--name-field-generic-body h3"):
+            full_str = title.xpath(".//text()").extract()[0]
+            date_match = re.search(r"[a-z]{3,9} \d{1,2},? \d{4}", full_str, flags=re.I).group(0)
+            date_obj = datetime.strptime(re.sub(',', '', date_match), '%B %d %Y')
```
pjsier: If you're only replacing regular strings without regex formatting you can call `date_match.replace(',', '')` instead of using `re.sub`. This goes for a couple other spots in the spider as well.

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,111 @@
...
+            return COMMITTEE
+        if "meeting" in title.lower():
+            return BOARD
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item, year):
+        """Parse start datetime as a naive datetime object."""
+        item = item.xpath(".//text()").extract()[0].split("-")[0].strip()
+        date = re.sub(r',|\xa0|\*|\*\*', '', item).strip(" ")
+        date_obj = datetime.strptime(date, '%A %B %d')
```
pjsier: If you're defaulting to a specific time, this can be simplified like this (also without setting intermediate variables that are returned without being used otherwise):

```python
return datetime.strptime(date + " 13", "%A %B %d %H")
```

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,111 @@
...
+        new_date_obj = datetime(int(year), date_obj.month, date_obj.day, 13, 00)
+        return new_date_obj
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        if "**" in item:
+            return {
+                "address": "",
```
pjsier: We always want an address included with a location, and if only a name is given we've just been looking up the address and adding it manually

city_scrapers/spiders/chi_northern_il_university.py
```diff
@@ -0,0 +1,111 @@
...
+                else:
+                    meeting["status"] = self._get_status(meeting)
+                meeting["id"] = self._get_id(meeting)
+                yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        title_obj = item.xpath(".//text()").extract()
+        title = re.sub(r'\(.+\)', '', "".join(title_obj))
+        title = title.split("-")[-1].strip(" ")
```
pjsier: The output of this line can be returned instead of assigning it to `title` again if it's returned in the next line

pjsier:

Issue #675, opened 2019-02-03
Spider: Illinois International Port District


pjsier:
URLs:

- Minutes: https://www.iipd.com/about/board-meeting-minutes
- Agendas: https://www.iipd.com/calendar/agendas
- Schedule: https://www.iipd.com/calendar/schedules

Spider Name: `il_port_district`
Agency Name: Illinois International Port District

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

paqman85:
I'll take this one if it's still available.

pjsier:
@paqman85 all yours!

pjsier:
@paqman85 are you still working on this? Otherwise I'll take a stab at it

paqman85:
Yes, im still on it. A few hiccups but I'm making progress now.

:)

radoslawkrolikowski:
This issue is unclaimed and seems that no one is active. I would like to have a go at this, if it is available?

pjsier:
@radoslawkrolikowski yep, it's been inactive long enough that it's all yours. Assigning you now

radoslawkrolikowski:
I encountered problems while calling `pytest`. The following is the one of the errors that regards the `legistar`:
![image](https://user-images.githubusercontent.com/55454254/69454810-5cec3d00-0d67-11ea-93f8-e0241ecba35c.png)
My local fork is up-to-date with the upstream, I also installed requirements one more time in my virtual env, in case something has changed. I would like to emphasize that I call `pytest` on the repository that is identical to the upstream, I didn't make any changes. Do you have an idea how to deal with this issue?

I have also other question, for the sake of clarification: if I my `start_urls` list consist of three elements, then I should supply three different HTML files to the `tests/files` directory? And how to include this in the test file? 

Have you got any example that shows how to change following lines, so that it will consecutively call all my start urls?
```
test_response = file_response(
    join(dirname(__file__), "files", "chi_boardofethics.html"),
    url='https://www.chicago.gov/city/en/depts/ethics/supp_info/minutes.html'
) 
```
While my parse function looks like this:
```
    def parse(self, response):
        if response.url == self.start_urls[0]:
            return self.parse_agendas(response)
        if response.url == self.start_urls[1]:
            return self.parse_minutes(response)
        if response.url == self.start_urls[2]:
            return self.parse_schedules(response)
```

pjsier:
[...]

PR #931, opened 2019-11-23
#675 Spider: Illinois International Port District

radoslawkrolikowski:
This pull request adds the Illinois International Port District Spider, the test file of that spider and target web pages as HTML files.

Spider reads the following meeting details from the http://www.iipd.com/calendar/schedules page:
- type of meeting (Board / Committee)
- date (month, day, year)
- location,
- it assumes default start time - 9:00am,
- includes all starting time changes (Start time 8:30am CST)
- Assigns "Special" label to special meetings.

In Agendas tab (http://www.iipd.com/calendar/agendas), we have got the following files that are scraped:
- Board Agenda  - added as link to BOARD type meetings,
- Leases and Agreement Committee Agenda  - added as link to COMMITTEE type meetings,
- Finance and Personnel Committee Agenda  -  added as link to COMMITTEE type meetings,

Board Minutes tab (http://www.iipd.com/about/board-meeting-minutes) contains the Minutes files of meetings that happened in the years 2019 - 2012. All the files are scraped, then corresponding files are added as links to the all meetings listed in the schedule (year 2019).

In test file, while testing the parsed links, the spider returns links as the https urls, but for unkown reasons test files sees them as http, thus target links in assert function were written as http urls.


city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class IlPortDistrictSpider(CityScrapersSpider):
+    name = "il_port_district"
+    agency = "Illinois International Port District"
+    timezone = "America/Chicago"
+    allowed_domains = ["https://www.iipd.com"]
+    start_urls = [
```
pjsier: You can remove `start_urls` if `start_requests` is used

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+        dt = " ".join([year, date, meeting_time])
+        return datetime.datetime.strptime(dt, "%Y %B %d %I:%M%p")
+
+    def _parse_agendas_links(self, classification):
+        if classification == 'Board':
+            link_list = [d for d in self.files_list if 'Board' in d['title']]
+        elif classification == 'Committee':
+            link_list = [d for d in self.files_list if 'Committee' in d['title']]
+        else:
+            link_list = [{}]
```
pjsier: An empty link list should just have `[]` instead of an empty dictionary

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+        "https://www.iipd.com/calendar/agendas", "https://www.iipd.com/about/board-meeting-minutes",
+        "https://www.iipd.com/calendar/schedules"
+    ]
+    location = {
+        "name": "Illinois International Port District ",
+        "address": "3600 E. 95th St. Chicago, IL 60617",
+    }
+
+    def start_requests(self):
+        start_urls = [
```
pjsier: Since you're doing something different in each section and the sections each depend on the results of the previous page you might need to only parse one of the document pages and then yield a request to the next one so that you can guarantee the schedules are parsed last. You could also implement the `spider_idle` signal to yield a request to the schedule page like in [`il_pollution_control`](https://github.com/City-Bureau/city-scrapers/blob/a11b7c4dda9018efcd3e3f25b7678546a1351cbe/city_scrapers/spiders/il_pollution_control.py)

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+            link_list = [{}]
+
+        return link_list
+
+    def _parse_minutes_links(self, start):
+        start = start.replace(hour=0, minute=0, second=0, microsecond=0)
+        link = self.minutes_dict.get(start, None)
+        if link:
+            name = "Board Meeting " + datetime.datetime.strftime(start, "%B %d, %Y")
+            print('NAME: LINK MINUTES: ', {name: link})
```
pjsier: This can be removed

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+            .extract()
+        file_prefix = [x.strip("\n ") for x in file_prefix]
+        file_links = response.xpath("//tr/td/a[@class='file-download']/@href").extract()
+        name_pattern = r"(?<=documents\/)(.*?)(?=.pdf)"
+        file_names = [re.findall(name_pattern, x)[0].replace('%', ' ') for x in file_links]
+
+        self.files_list = []
+
+        for link, prefix, name in list(zip(file_links, file_prefix, file_names)):
+            self.files_list.append({'title': prefix + " " + name, 'href': link})
```
pjsier: If we know these files are always agendas the title can just be "Agenda"

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def parse_minutes(self, response):
+        rows = response.xpath("//tr")
+        self.minutes_dict = {}
+        for row in rows:
+            file_name = row.xpath(".//td[@class='views-field views-field-title']/text()")\
```
pjsier: Similar to the agendas, if we know these are all minutes we don't need the meeting date associate since they're already included in the meeting details

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,181 @@
...
+        elif classification == 'Committee':
+            link_list = [d for d in self.files_list if 'Committee' in d['title']]
+        else:
+            link_list = [{}]
+
+        return link_list
+
+    def _parse_minutes_links(self, start):
+        start = start.replace(hour=0, minute=0, second=0, microsecond=0)
+        link = self.minutes_dict.get(start, None)
```
pjsier: Instead of replacing the details, it's been easier to use a `defaultdict` object keyed by `date` rather than `datetime` objects so that you can call `minutes_dict[start.date()]`

tests/test_il_port_district.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def test_id():
+    assert parsed_items[0]["id"] == "il_port_district/201903010900/x/special_committee_meeting"
+    assert parsed_items[1]["id"] == "il_port_district/201903010900/x/special_board_meeting"
+    assert parsed_items[3]["id"] == "il_port_district/201901180900/x/board_meeting"
+
+
+def test_status():
+    assert parsed_items[0]["status"] == "passed"
```
pjsier: We should use the constant `PASSED` here

city_scrapers/spiders/il_port_district.py
radoslawkrolikowski: I should use "Agenda" title for each file, or distinguish between different Agenda types (Board, Committee) an associate them to corresponding meetings?

Because if I use "Agenda" for each file name then this function will not work properly, as it looks for the "Board" or "Committee" phrases in the file names:
```
    def _parse_agendas_links(self, classification):
        if classification == 'Board':
            link_list = [d for d in self.files_list if 'Board' in d['title']]
        elif classification == 'Committee':
            link_list = [d for d in self.files_list if 'Committee' in d['title']]
        else:
            link_list = []

        return link_list
```
How about if I get rid of the file name part that was derived from the url, so instead of the following:
- Leases and Agreement Committee Agenda L 26A 20Agenda 20November 202019
We will have:
- Leases and Agreement Committee Agenda ?

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,186 @@
...
+    location = {
+        "name": "Illinois International Port District ",
+        "address": "3600 E. 95th St. Chicago, IL 60617",
+    }
+
+    schedules_url = "https://www.iipd.com/calendar/schedules"
+
+    def start_requests(self):
+        start_urls = [
+            "https://www.iipd.com/calendar/agendas",
```
pjsier: Now that it's just two URLs it might be simpler to just do this:

```python
yield scrapy.Request(url="https://www.iipd.com/calendar/agendas", callback=self.parse_agendas)
yield scrapy.Request(url="https://www.iipd.com/about/board-meeting-minutes", callback=self.parse_minutes)
```

city_scrapers/spiders/il_port_district.py
pjsier: Gotcha, yeah that should work and at least leave the values a bit cleaner

city_scrapers/spiders/il_port_district.py
radoslawkrolikowski: Yeah, it looks much simpler and cleaner now.

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,183 @@
...
+
+        return link_list
+
+    def _parse_minutes_links(self, start):
+        link = self.minutes_dict.get(start.date(), None)
+        if link:
+            name = "Board Meeting Minutes"
+            return [{'title': name, 'href': link}]
+        else:
+            return None
```
pjsier: This should return an empty list if not available, otherwise this might cause issues

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,183 @@
...
+
+        for i, row in enumerate(strong_meetings + rows[1:]):
+            if i >= len(strong_meetings):
+                meetings_dates = row.xpath(".//div/text()").extract()
+                if not meetings_dates:
+                    continue
+            else:
+                meetings_dates = row
+
+            for i, date in enumerate(meetings_dates):
```
pjsier: It might make sense to put more of this date parsing logic in `_parse_start`. Also, the July 19th meeting is set as 8:30am and it looks like it shouldn't be.

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,183 @@
...
+            name = "Board Meeting Minutes"
+            return [{'title': name, 'href': link}]
+        else:
+            return None
+
+    def _validate_location(self, response):
+        loc = response.xpath("//strong")[-1].xpath(".//text()").extract()
+        loc = [x.strip("\n ") for x in loc]
+        loc = " ".join(loc[-2:])
+        if '3600 E. 95th St. Chicago, IL 60617' != loc:
```
pjsier: To make sure we're not raising an error if they remove a period, we could just check for `3600` here

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,183 @@
...
+        elif "committee" in meeting_types.lower():
+            return COMMITTEE
+        else:
+            return NOT_CLASSIFIED
+
+    def _parse_start(self, year, date, meeting_time):
+        dt = " ".join([year, date, meeting_time])
+        return datetime.datetime.strptime(dt, "%Y %B %d %I:%M%p")
+
+    def _parse_agendas_links(self, classification):
```
pjsier: Looking at this more closely, it should only associate the agenda with the meeting that it corresponds to, but right now it's attaching the November agenda to every meeting. We should check the month here

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,193 @@
...
+            link_list = [
+                d for d in self.files_list if 'Committee' in d['title'] and date in d['title']
+            ]
+        else:
+            link_list = []
+
+        return link_list
+
+    def _parse_minutes_links(self, start):
+        link = self.minutes_dict.get(start.date(), None)
```
pjsier: This could be simplified by making the values in `minutes_dict` lists with the title already applied so that this method wouldn't be needed and you could just call `self.minutes_dict.get(start.date(), [])`

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,193 @@
...
+    def _parse_classification(self, i, meeting_types):
+        if "board" in meeting_types.lower():
+            return BOARD
+        elif "committee" in meeting_types.lower():
+            return COMMITTEE
+        else:
+            return NOT_CLASSIFIED
+
+    def _parse_start(self, date, year, meeting_time):
+
```
pjsier: `self.changed_meeting_time` and `self.special_meeting` are only used in this function, so these and any other variables that don't need to be used across different requests should be passed as arguments to the function or handled separately.

`special_meeting` could be handled in a `_parse_title` function that checks the date string, because right now logic for parsing the titles and times is combined in this method and it makes it harder to read

city_scrapers/spiders/il_port_district.py
```diff
@@ -0,0 +1,193 @@
...
+            new_date = re.findall(r"(?<=\()(.*?)(?=NEW)", date)
+            date = new_date[0][:-3] if new_date else date
+
+        date = date.strip(" *")
+        dt = " ".join([year, date, meeting_time])
+        dt = datetime.datetime.strptime(dt, "%Y %B %d %I:%M%p")
+
+        return dt
+
+    def _parse_agendas_links(self, classification, start):
```
pjsier: This could be simplified by making `files_list` into a `collections.defaultdict` object keyed by a tuple of classification and start date string. That way you could get the agenda links for a meeting with `self.agenda_dict[(classification, start.strftime("%B %Y"))]`.

In general, we should be using the constants instead of "Board" and "Committee" strings here

Also, `strftime` can be called directly on a `datetime` object and doesn't need to be called from the module

Issue #808, opened 2019-07-21
Spider: Illinois State Board of Elections


pjsier:
URL: https://www.elections.il.gov/AboutTheBoard/Agenda.aspx
Doc URL: https://www.elections.il.gov/AboutTheBoard/MeetingMinutesAll.aspx
Spider Name: `il_elections`
Agency Name: Illinois State Board of Elections

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

raufabr:
Is this issue still open? I would like to have a go at this.

pjsier:
@raufabr yep, all yours if you're interested!

akaysh:
@raufabr are you still working on this? If not, I would like to work on this.

pjsier:
@akaysh I think you're good to work on this, thanks!

AmiMappy:
Can I give it a go?

pjsier:
Hi @AmiMappy, thanks for the interest in contributing! It looks like someone is working on this one, but you’re welcome to take a look at any of the issues that haven’t had any comments in a month or so. Let me know if you have any questions

AmiMappy:
Thanks a lot.

luchiago:
Is this issue still need help?

pjsier:
@luchiago this and some of the other issues you commented on are open, but we ask that contributors only work on one scraper at a time. Is there one you're most interested in starting on?

luchiago:
@pjsier I would like to work on #680.

pjsier:
Gotcha, assigned here

luchiago:
Thanks man.

luchiago:
@pjsier The link seems broken. What should I do?

pjsier:
Both links look fine to me, not sure where the error would be coming from

luchiago:
@pjsier Should I crawl the docs from the `Doc URL`?

luchiago:
@pjsier I'm facing trouble to crawl this site. Can I work in another spider?

pjsier:
Sure, just comment on a spider you'd want to work on next

BurakcanK:
Hi, I am fairly new to open-source projects but curious nevertheless, could I have a go if that's possible ?

pjsier:
@BurakcanK all yours! Assigning you now

pjsier:
Closed by #938

PR #938, opened 2019-12-09
0808-spider-il_elections

BurakcanK:
## Summary

**Issue:** #808 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

PR for counseling.


pjsier:
Is there anything specific you'd like me to look at? In general we want the PR template checklist to stay in the description even if it's not all checked yet, and for the CI checks to pass

BurakcanK:
PR template added to the description, CI checks pass for now.

Help needed on handling multiple URLs :)

pjsier:
For parsing multiple URLs, you're going to want to use the page of meeting minutes as your start URL. Then in `parse` you can parse the meeting minutes into a dictionary that stores each date as a key and the link as a value. From this `parse` method you can yield a request to the agenda page, and while you're parsing each meeting you can pull the minutes link from that dictionary if it's present. Here's an example https://github.com/City-Bureau/city-scrapers/blob/00bc430771c2a7fc5fee456bdde83bbf65c41164/city_scrapers/spiders/chi_teacherpension.py

BurakcanK:
Waiting for feedback if I missed some minor details or something, other than that it looks completed I reckon.

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,148 @@
...
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        # parse meeting minutes and addresses first to associate them with the agenda
+        self._parse_minutes(response)
+        self._parse_addresses(response)
```
pjsier: I don't think this needs to be parsed in the minutes response, since the same footer is on the agenda page and moving it there could simplify this a bit

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,148 @@
...
+        """Parse or generate links."""
+        links = []
+        href = item.css("td")[0].css("::attr(href)").extract_first()
+        if href:
+            href = response.urljoin(href)
+            title = "Agenda"
+        else:
+            href = ""
+            title = ""
+        links.append({"href": href, "title": title})
```
pjsier: We shouldn't have empty  link dicts in the list, so any empty values should be ignored

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+    def _parse_minutes(self, response):
+        """Parse all past board meeting minutes, store for association to meetings."""
+        minutes = response.css("#ContentPlaceHolder1_gvMeetingMinutes td > a")
+        for minute in minutes:
+            date = minute.css("::text").extract_first().strip()
+            parsed_date = datetime.strptime(date, "%a, %B %d, %Y")
+            link = minute.css("::attr(href)").extract_first()
+            self.meeting_minutes[parsed_date] = link
```
pjsier: The key for this should be something like `parsed_date.date()` so that we don't have to worry about checking the time

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,148 @@
...
+        href = item.css("td")[0].css("::attr(href)").extract_first()
+        if href:
+            href = response.urljoin(href)
+            title = "Agenda"
+        else:
+            href = ""
+            title = ""
+        links.append({"href": href, "title": title})
+
+        if start.date() in [i.date() for i in self.meeting_minutes]:
```
pjsier: A simpler way of checking for this (when the comment above is applied) would be:

```python
if start.date() in self.meeting_minutes:
```

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,148 @@
...
+            dont_filter=True,
+        )
+
+    def _parse_minutes(self, response):
+        """Parse all past board meeting minutes, store for association to meetings."""
+        minutes = response.css("#ContentPlaceHolder1_gvMeetingMinutes td > a")
+        for minute in minutes:
+            date = minute.css("::text").extract_first().strip()
+            parsed_date = datetime.strptime(date, "%a, %B %d, %Y")
+            link = minute.css("::attr(href)").extract_first()
```
pjsier: It looks like the `href` attribute opens up to a viewer page. We want to parse out everything that comes after the query parameter starting with `/Downloads` so that we can get the PDF link directly

tests/test_il_elections.py
```diff
@@ -0,0 +1,107 @@
...
+
+freezer = freeze_time("2019-12-04")
+freezer.start()
+
+test_response = file_response(
+    join(dirname(__file__), "files", "il_elections_minutes.html"),
+    url="https://www.elections.il.gov/AboutTheBoard/MeetingMinutesAll.aspx",
+)
+
+with open(join(dirname(__file__), "files", "il_elections_agenda.html"), "r") as f:
```
pjsier: You can copy the `test_response` code above and apply it here since it does a lot of what the next lines do for you already

tests/test_il_elections.py
```diff
@@ -0,0 +1,107 @@
...
+def test_status():
+    assert parsed_items[0]["status"] == PASSED
+    assert parsed_items[5]["status"] == TENTATIVE
+
+
+def test_location():
+    assert parsed_items[0]["location"] == {
+        "name": "Springfield",
+        "address":
+            "2329 S. MacArthur Blvd. Springfield, IL 62704 Phone: 217-782-4141 Fax: 217-782-5959"
```
pjsier: We don't want the Phone and Fax in the location

tests/test_il_elections.py
```diff
@@ -0,0 +1,107 @@
...
+    }
+    assert parsed_items[5]["location"] == {
+        "address":
+            "100 W. Randolph, Suite 14-100 Chicago, IL 60601 Phone: 312-814-6440 Fax: 312-814-6485",
+        "name": "Chicago"
```
pjsier: If the name is just the location we can leave the `name` key empty

tests/test_il_elections.py
```diff
@@ -0,0 +1,107 @@
...
+
+def test_source():
+    assert parsed_items[0]["source"] == "https://www.elections.il.gov/AboutTheBoard/Agenda.aspx"
+
+
+def test_links():
+    assert parsed_items[0]["links"] == [{
+        "href":
+            "https://www.elections.il.gov/DocDisplay.aspx" +
+            "?Doc=/Downloads/AboutTheBoard/PDF/07_01_19Agenda.pdf",
```
pjsier: To elaborate on my comment above, we would want this to return "https://www.elections.il.gov/Downloads/AboutTheBoard/PDF/07_01_19Agenda.pdf". So we can use the `urllib.parse` module to get the `Doc` query parameter

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,143 @@
...
+class IlElectionsSpider(CityScrapersSpider):
+    name = "il_elections"
+    agency = "Illinois State Board of Elections"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.elections.il.gov/AboutTheBoard/MeetingMinutesAll.aspx"]
+    custom_settings = {"ROBOTSTXT_OBEY": False}
+
+    def __init__(self):
+        self.meeting_minutes = dict()
+        self.addresses = dict()
```
pjsier: This still doesn't need to be in the main `__init__` method and can just be passed as a regular dictionary argument to `_parse_location`

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,143 @@
...
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Illinois State Board of Elections"
```
pjsier: This can just be "Board of Elections"

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        desc = item.css("td")[3].css("::text").extract_first().strip()
```
pjsier: If you don't need an intermediate variable you can return the output directly without assigning it to `desc`

city_scrapers/spiders/il_elections.py
```diff
@@ -0,0 +1,143 @@
...
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return BOARD
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        date = item.css("td")[0].css("::text").extract_first().strip()
+        time = item.css("td")[1].css("::text").extract_first().strip().replace(".", "")
+        dt = date + " " + time
+        parsed_dt = datetime.strptime(dt, "%a, %B %d, %Y %I:%M %p")
```
pjsier: This can be returned directly as well

BurakcanK:
Done :+1:

Issue #905, opened 2019-10-24
Spider: Chicago Special Service Area #4 95th Street


pjsier:
URL: https://95thstreetba.org/events/category/board-meeting/
Spider Name: chi_ssa_4
Agency Name: Chicago Special Service Area #4 South Western Avenue

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

julytn:
I can take this one

pjsier:
Sounds good!

JordanLozinski:
is this still being worked on?

pjsier:
@JordanLozinski since there haven't been any updates in over a month it's all yours if you're interested

JordanLozinski:
Sounds good. I'll start soon.

JordanLozinski:
I could use some advice from somebody who's better at Scrapy than I am (this is my first time using it)

From [chi_ssa_4.py](https://github.com/JordanLozinski/city-scrapers/blob/0905-spider-chi_ssa_4/city_scrapers/spiders/chi_ssa_4.py)
```
start_urls = [
    "https://95thstreetba.org/events/category/board-meeting/?"
    "tribe_paged=1&tribe_event_display=list&tribe-bar-date=2017-10-01"
]

def parse(self, response):
        """
        `parse` should always `yield` Meeting items.
        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
        needs.
        """
        obj = response.css(".tribe-events-loop .tribe-events-category-board-meeting")
        for item in obj:
            # Only grab things of class tribe-events-category-board-meeting
            start, end = self._parse_time(item)
            meeting = Meeting(
              # ... Removed 
            )

            meeting["status"] = self._get_status(meeting)
            meeting["id"] = self._get_id(meeting)

            yield meeting

        # Look for next page button
        next_page = response.css(
            "#tribe-events-footer .tribe-events-nav-pagination "
            ".tribe-events-sub-nav .tribe-events-nav-next a::attr(href)"
        ).extract_first()
        if next_page:
            yield scrapy.Request(next_page, callback=self.parse)
```
[...]

PR #941, opened 2019-12-26
#905 Chicago Special Service Area #4 95th Street

JordanLozinski:
## Summary

**Issue:** #905 

## Checklist

- [ ] Tests are implemented
- [x] All tests are passing

## Questions

From the way I had to implement my spider, I'm unsure how to test it with the given pytest template. (GETs on the target URL produce strange results - there's no way to get all of the meeting data with GETs, you have to POST a specific URL on the website to get all the meeting data) Given that, I think the best way to test it would be to simply run `scrapy crawl chi_ssa_4` (as in, a crawl on the real website and not a file response) and do checks on the items it returns. I started working on that (at the top of `tests/test_chi_ssa_4.py`) but I feel like I'm going about it the wrong way. Scrapy contracts might be a good way to go about this but I don't know if they play well with pytest. If we find a way to mock the POST responses that would also make it testable. Does anyone have any suggestions on how to move forward with testing this?


city_scrapers/spiders/chi_ssa_4.py
```diff
@@ -0,0 +1,172 @@
...
+from city_scrapers_core.constants import NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa4Spider(CityScrapersSpider):
+    name = "chi_ssa_4"
+    agency = "Chicago Special Service Area #4 South Western Avenue"
+    timezone = "America/Chicago"
+    start_urls = [
```
pjsier: You'll want to dynamically generate this so that the time range will pull the few months around the time it's being run. Here's an example of something like that:

https://github.com/City-Bureau/city-scrapers/blob/9b90ad7ac653c608ccb84b0d0b7f5dc183405b35/city_scrapers/spiders/il_public_health.py#L17-L28

city_scrapers/spiders/chi_ssa_4.py
```diff
@@ -0,0 +1,172 @@
...
+        "https://95thstreetba.org/events/category/board-meeting/?"
+        "tribe_paged=1&tribe_event_display=list&tribe-bar-date=2018-10-01"
+    ]
+
+    def parse(self, response):
```
pjsier: This might be better in a `start_requests` method actually (as opposed to my comment above). Here's an example of that:

https://github.com/City-Bureau/city-scrapers/blob/20a12ba5d76186cba65b45f7f764f02393d4a991/city_scrapers/spiders/cook_county.py#L17-L27

tests/test_chi_ssa_4.py
```diff
@@ -0,0 +1,134 @@
+import http.client
+
+import pytest  # noqa
+from scrapy import FormRequest
+from scrapy.http import Request, TextResponse
+
+from city_scrapers.spiders.chi_ssa_4 import ChiSsa4Spider
+
+"""
+ChiSsa4Spider can't be tested in the typical way because it does POST
```
pjsier: Here's an example of how to test when parsing from a POST request:

https://github.com/City-Bureau/city-scrapers/blob/9b90ad7ac653c608ccb84b0d0b7f5dc183405b35/tests/test_cook_landbank.py#L11-L37

It shouldn't be overall that different as long as you pull the parsing logic out of functions that perform POST requests. Let me know if the example isn't clear

city_scrapers/spiders/chi_ssa_4.py
JordanLozinski: I think the start_urls doesn't need to be dynamically generated actually. The start URL doesn't actually affect our parsing at all (we should only be parsing the HTML we get from POST responses).

What we can do instead is dynamically determine the start date for our POST request (it'll only respond with meetings after that date) and only parse X months into the future. I'll do that

city_scrapers/spiders/chi_ssa_4.py
JordanLozinski: See my above comment.

I think I'll have it look just one month in the future then.

I'll resolve both these comments when it's implemented and request another review

tests/test_chi_ssa_4.py
JordanLozinski: >It shouldn't be overall that different as long as you pull the parsing logic out of functions that perform POST requests. Let me know if the example isn't clear

It seems I should just pull the POST request stuff out of _parse_meetings and we'll be good.

city_scrapers/spiders/chi_ssa_4.py
alexanderwoo: Follow up to the template of looping through the dates. I tested this out and this only returns two dates (today and next_month) instead of all dates in between. I'm not sure if that is expected behavior. The implementation below iterates through a start and end date. This is for reference in case someone is looking for this.

```suggestion
   from datetime import datetime, timedelta
   
   '''Next 90 days'''
    today = datetime.now()
    tomorrow = today.replace(day=today.day+1)
    next_months = tomorrow + timedelta(days=90)
    delta = timedelta(days=1)
    while today <= next_months:
       print(today.strftime("%Y-%m-%d"))
       today += delta    
```

pjsier:

Issue #680, opened 2019-02-03
Spider: Illinois Aging Advisory Council


pjsier:
URL: https://www2.illinois.gov/aging/PartnersProviders/OlderAdult/Pages/acmeetings.aspx
Spider Name: `il_aging_advisory_council`
Agency Name: Illinois Aging Advisory Council

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

allenjhuang:
On it!

ElijahLawson:
If this is currently inactive, I can get working on it.

pjsier:
@ElijahLawson sounds good!

shafique18:
can I work on it
please let me know what things need to scrape from the given site

pjsier:
@shafique18 sounds good! Everything you need to get started should be in the development documentation https://cityscrapers.org/docs/development/

luchiago:
Is this issue still need help?

pjsier:
@luchiago assigning you here, @shafique18 let me know if you're still working on it but according to our contributing guidelines we sometimes need to reassign issues if they've been inactive for more than 30 days

luchiago:
@pjsier I made a mistake, I wrote the wrong issue. Is the #808, because the link of this issue seems broken. Check my comment at #808.

erik-kristofer-anderson:
This looks available, so I'd like to claim it.
Within a week, I'll come back here and comment on my progress. (If it turns out I'm in over my head I'll bow out at that point.)

pjsier:
@erik-kristofer-anderson sounds good!

erik-kristofer-anderson:
@pjsier I'm going to bow out at this point. Please mark this as un-claimed.

pjsier:
thanks for the heads up!

nick-symon:
@pjsier Hi! I'd like to claim this issue and begin work on it!

pjsier:
@nick-symon sounds good!

pjsier:
Closed by #944, thanks @nick-symon!

PR #944, opened 2020-01-14
0680 spider il aging advisory council

nick-symon:
## Summary

**Issue:** #680 

## Checklist

- [X ] Tests are implemented
- [ X] All tests are passing
- [ X] Style checks run (see [documentation](https://cityscrapers.org/docs/development/) for more details)
- [X ] Style checks are passing
- [ X] Code comments from template removed

## Questions

### DIFF
--- a/city_scrapers/spiders/il_aging_advisory_council.py
+++ b/city_scrapers/spiders/il_aging_advisory_council.py
@@ -3,11 +3,24 @@ from city_scrapers_core.items import Meeting
 from city_scrapers_core.spiders import CityScrapersSpider
 from datetime import datetime
 
+
 class IlAgingAdvisoryCouncilSpider(CityScrapersSpider):
     name = "il_aging_advisory_council"
     agency = "Illinois Aging Advisory Council"
     timezone = "America/Chicago"
-    start_urls = ["https://www2.illinois.gov/aging/PartnersProviders/OlderAdult/Pages/acmeetings.aspx"]
+    start_urls = [
+        "https://www2.illinois.gov/aging/PartnersProviders/OlderAdult/Pages/acmeetings.aspx"
+    ]
+    location = {
+        'address': {
+            'Springfield': 'One Natural Resources Way #100',
+            'Chicago': '160 N. LaSalle Street, 7th Floor'
+        },
+        'name': {
+            'Springfield': 'Illinois Department on Aging',
+            'Chicago': 'Michael A. Bilandic Building'
+        }
+    },
 
     def parse(self, response):
         """
@@ -16,6 +29,8 @@ class IlAgingAdvisoryCouncilSpider(CityScrapersSpider):
         Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
         needs.
         """
+        self._validate_location(response)
+
         for item in response.css('div.ms-rtestate-field strong::text').getall():
             try:
                 datetime.strptime(item, "%B %d, %Y")
@@ -26,14 +41,14 @@ class IlAgingAdvisoryCouncilSpider(CityScrapersSpider):
                     continue
 
             meeting = Meeting(
-                title=self._parse_title(item),
+                title='Illinois Department on Aging Advisory Committee Meetings',
                 description=self._parse_description(item),
                 classification=self._parse_classification(item),
                 start=self._parse_start(item),
                 end=self._parse_end(item),
                 all_day=self._parse_all_day(item),
-                time_notes=self._parse_time_notes(item),
-                location=self._parse_location(item),
+                time_notes=self._parse_time_notes(response),
+                location=self.location,
                 links=self._parse_links(item),
                 source=self._parse_source(response),
             )
@@ -49,7 +64,7 @@ class IlAgingAdvisoryCouncilSpider(CityScrapersSpider):
 
     def _parse_description(self, item):
         """Parse or generate meeting description."""
-        return ""
+        return "Meetings are open to the public"
 
     def _parse_classification(self, item):
         """Parse or generate classification from allowed options."""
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class IlAgingAdvisoryCouncilSpider(CityScrapersSpider):
+    name = "il_aging_advisory_council"
+    agency = "Illinois Aging Advisory Council"
+    timezone = "America/Chicago"
+    start_urls = [
+        "https://www2.illinois.gov/aging/PartnersProviders/OlderAdult/Pages/acmeetings.aspx"
+    ]
+    location = {
```
pjsier: `location` should only return a `dict` with string keys for `address` and `name`. Some of the Illinois ones have been a bit odd because they'll have multiple locations, but we've been using the Chicago one when this happens

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self._validate_meeting_times(response)
+
+        valid_date_list = self._create_valid_date_list(response)
+
+        if valid_date_list:
+            for item in valid_date_list:
+                meeting = Meeting(
+                    title='Illinois Department on Aging Advisory Committee Meetings',
+                    description='Meetings are open to the public',
```
pjsier: Generally the `description` is only filled in if there's something dynamic on the page, so this can be an empty string

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,113 @@
...
+        return start_time
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        end_time = item.replace(hour=16)
+        return end_time
+
+    def _parse_time_notes(self, response):
+        """Parse any additional notes on the timing of the meeting"""
+        time_notes = response.css('div.ms-rtestate-field p::text').get()
```
pjsier: `time_notes` is used for things like whether a meeting start time is estimated or is conditional based on the previous meeting, so in this case it should be an empty string. More info [here](https://cityscrapers.org/docs/development/#time_notes)

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,113 @@
...
+        return time_notes
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        return [{"href": "", "title": ""}]
```
pjsier: We want to parse the documents on the page here like Minutes and presentations, and if there are no documents available this should return an empty list. The `dict` values should always have a non-blank `href` key

tests/test_il_aging_advisory_council.py
```diff
@@ -0,0 +1,69 @@
...
+spider = IlAgingAdvisoryCouncilSpider()
+
+freezer = freeze_time("2019-12-23")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+freezer.stop()
+
+# def test_tests():
```
pjsier: This commented out test and the "Uncomment below" comment can be removed

tests/test_il_aging_advisory_council.py
```diff
@@ -0,0 +1,69 @@
...
+"""
+Uncomment below
+"""
+
+
+def test_title():
+    assert parsed_items[0]["title"] == "Illinois Department on Aging Advisory Committee Meetings"
+
+
+def test_start():
```
pjsier: Generally we'll want to keep all of the boilerplate test methods (like `test_end`) in here unless there's a reason for removing them. They're helpful for tracking changes when updating scrapers

city_scrapers/spiders/il_aging_advisory_council.py
nick-symon: This has been updated to return `location` with only Chicago address and name

city_scrapers/spiders/il_aging_advisory_council.py
nick-symon: Updated description to empty string

city_scrapers/spiders/il_aging_advisory_council.py
nick-symon: Updated time_notes to empty string

tests/test_il_aging_advisory_council.py
nick-symon: Updated tests so that all attributes of 'meeting' object are being tested.

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,119 @@
...
+            text()').re(r'([A-Z]\w+)\s*(\d\d?),*\s*(\d\d\d\d)')
+            if not row_date:
+                continue
+            else:
+                converted_date = self._convert_date(row_date)
+
+            row_urls = table_row.xpath('.//a')
+
+            meeting = Meeting(
+                title='Illinois Department on Aging Advisory Committee Meetings',
```
pjsier: I missed this on the first pass, but since the meeting is always the same we can avoid repeating information by returning "Advisory Committee" instead

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,119 @@
...
+            '//*[@id="ctl00_PlaceHolderMain_ctl01__ControlWrapper_RichHtmlField"]\
+            /table'
+        )
+
+        for table_row in table.xpath('.//tr'):
+            row_date = table_row.xpath('.//strong/\
+            text()').re(r'([A-Z]\w+)\s*(\d\d?),*\s*(\d\d\d\d)')
+            if not row_date:
+                continue
+            else:
```
pjsier: Since none of the code after the if condition will run if it's not true, you can remove this `else` and just have `converted_date` be assigned on the next line outside the condition

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,119 @@
...
+            yield meeting
+
+    def _validate_location(self, response):
+        location_test = response.xpath(
+            '//*[@id="ctl00_PlaceHolderMain_ctl01__ControlWrapper_RichHtmlField"]\
+        /ul/li[2]/text()'
+        ).get()
+        if '7th Floor, 160 N. LaSalle Street' not in location_test:
+            raise ValueError("Meeting location has changed")
+        else:
```
pjsier: This `else` condition isn't needed since it's what would happen anyway

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,119 @@
...
+        """Parse start datetime as a naive datetime object."""
+        month = item[0]
+        day = item[1]
+        year = item[2]
+        converted_date = datetime.strptime(month[0:3] + day + year, '%b%d%Y')
+        return converted_date
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        start_time = item.replace(hour=13)
```
pjsier: For this and `_parse_end` you can return the value directly instead of assigning it to a variable since we don't use the variable outside the return statement

city_scrapers/spiders/il_aging_advisory_council.py
```diff
@@ -0,0 +1,119 @@
...
+        end_time = item.replace(hour=15)
+        return end_time
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = []
+        for i in item:
+            href = i.xpath('@href').get()
+            full_href = 'https://www2.illinois.gov' + href
+            text = i.xpath('text()').get()
```
pjsier: We'll want to call `.strip()` to remove leading or trailing whitespace on the title and `href` here to keep the output clean. There are some extra whitespace characters in the output now: https://github.com/City-Bureau/city-scrapers/runs/392391279#step:11:79

Also, we'll need to check if the value is `None` before stripping the whitespace, since it looks like that's the case for this link: https://github.com/City-Bureau/city-scrapers/runs/392391279#step:11:144 If the value for title or `href` is `None` anywhere we should ignore that link and `continue` to the next item

tests/test_il_aging_advisory_council.py
```diff
@@ -0,0 +1,86 @@
...
+    ]
+
+
+def test_source():
+    assert parsed_items[
+        0]["source"
+           ] == "https://www2.illinois.gov/aging/PartnersProviders/OlderAdult/Pages/acmeetings.aspx"
+
+
+# @pytest.mark.parametrize("item", parsed_items)
```
pjsier: This can be removed

nick-symon: updated

Issue #559, opened 2018-11-07
Spider: Chicago Special Service Area #18 North Halsted


pjsier:
URL: https://northalsted.com/community/
Spider Name: `chi_ssa_18`
Agency Name: Chicago Special Service Area #18 North Halsted

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

AmiMappy:
Can I work on this?

pjsier:
@AmiMappy all yours!

AmiMappy:
I keep getting the 'no module named city_scrapers_core' error. Help needed.

AmiMappy:
I keep getting the 'no module named legistar' error. Help needed.

AmiMappy:
What classification should I provide the spider with?
(ADVISORY_COMMITTEE, BOARD, CITY_COUNCIL, COMMISSION, COMMITTEE, FORUM, POLICE_BEAT, NOT_CLASSIFIED). By default, it is NOT_CLASSIFIED.

pjsier:
The SSAs are generally commissions, so `COMMISSION` would be best

AmiMappy:
Thanks.

AmiMappy:
What should I put in _parse_links()?

pjsier:
You can take a look at some of the other scrapers [for examples](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_ssa_26.py#L50-L69), but in general you want to find agendas and minutes and return a list of dictionaries with `title` and `href` keys. Also feel free to [join the Slack](https://airtable.com/shrsdRcYVzp019U22) for general questions

AmiMappy:
Can you help me on how I should go about this? :

          ERROR: /home/travis/build/City-Bureau/city-scrapers/chi_ssa_18.py Imports are incorrectly sorted.

          ---/home/travis/build/City-Bureau/city-scrapers/chi_ssa_18.py:before	2019-10-05 12:20:37.663324

         +++ /home/travis/build/City-Bureau/city-scrapers/chi_ssa_18.py:after	2019-10-05 12:21:23.801026

         @@ -1,6 +1,5 @@

         +import re
          from datetime import datetime, time
         -import re
          from city_scrapers_core.constants import COMMISSION
          from city_scrapers_core.items import Meeting

pjsier:
[...]

PR #945, opened 2020-02-01
0559 spider chi ssa 18

nick-symon:
## Summary

**Issue:** #559 

## Checklist

- [X ] Tests are implemented
- [ X] All tests are passing
- [ X] Style checks run (see [documentation](https://cityscrapers.org/docs/development/) for more details)
- [ X] Style checks are passing
- [ X] Code comments from template removed

## Questions

### DIFF
--- a/city_scrapers/spiders/chi_ssa_18.py
+++ b/city_scrapers/spiders/chi_ssa_18.py
@@ -39,32 +39,30 @@ class ChiSsa18Spider(CityScrapersSpider):
                 split_items = [Selector(text=section) for section in re.split(r'<br>', item.get())]
                 for split_string in split_items:
                     meeting_date_match = split_string.re_first(r'([A-Z]\w{2,}\s\d\d?)')
-                    if meeting_date_match and meeting_year:
-                        try:
-                            converted_date = \
-                                self._convert_date(
-                                    meeting_date_match, meeting_year
-                                    )
-                        except ValueError:
-                            continue
+                    if not (meeting_date_match and meeting_year):
+                        continue
+                    converted_date = \
+                        self._convert_date(
+                            meeting_date_match, meeting_year
+                            )
 
-                        meeting = Meeting(
-                            title='Commission',
-                            description='',
-                            classification=COMMISSION,
-                            start=self._parse_start(converted_date),
-                            end=self._parse_end(converted_date),
-                            all_day=self._parse_all_day(item),
-                            time_notes='',
-                            location=self.location,
-                            links=self._parse_links(split_string),
-                            source=self._parse_source(response),
-                        )
+                    meeting = Meeting(
+                        title='Commission',
+                        description='',
+                        classification=COMMISSION,
+                        start=self._parse_start(converted_date),
+                        end=self._parse_end(converted_date),
+                        all_day=self._parse_all_day(item),
+                        time_notes='',
+                        location=self.location,
+                        links=self._parse_links(split_string),
+                        source=self._parse_source(response),
+                    )
 
-                        meeting["status"] = self._get_status(meeting)
-                        meeting["id"] = self._get_id(meeting)
+                    meeting["status"] = self._get_status(meeting)
+                    meeting["id"] = self._get_id(meeting)
 
-                        yield meeting
+                    yield meeting
 
     def _validate_location(self, item):
         location_test = item.get()
--- a/city_scrapers/spiders/chi_ssa_18.py
+++ b/city_scrapers/spiders/chi_ssa_18.py
@@ -1,4 +1,7 @@
-from city_scrapers_core.constants import NOT_CLASSIFIED
+import re
+from datetime import datetime
+
+from city_scrapers_core.constants import COMMISSION
 from city_scrapers_core.items import Meeting
 from city_scrapers_core.spiders import CityScrapersSpider
 
@@ -8,6 +11,11 @@ class ChiSsa18Spider(CityScrapersSpider):
     agency = "Chicago Special Service Area #18 North Halsted"
     timezone = "America/Chicago"
     start_urls = ["https://northalsted.com/community/"]
+    location = {
+        'name': 'Center on Halsted',
+        'address': '3656 N Halsted St, Chicago IL 60613, Conference Room 200 on the 2nd Floor'
+    }
+    meeting_year = ''  # This is used to track year in header elements & append to meetings
 
     def parse(self, response):
         """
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+                            converted_date = \
+                                self._convert_date(
+                                    meeting_date_match.group(), ChiSsa18Spider.meeting_year
+                                    )
+                        except ValueError:
+                            continue
+                        links = re.findall(r'href="(.*?)">(.*?)</a>', split_string)
+
+                        meeting = Meeting(
+                            title='SSA #18 Commission',
```
pjsier: "Commission" alone is fine here

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+class ChiSsa18Spider(CityScrapersSpider):
+    name = "chi_ssa_18"
+    agency = "Chicago Special Service Area #18 North Halsted"
+    timezone = "America/Chicago"
+    start_urls = ["https://northalsted.com/community/"]
+    location = {
+        'name': 'Center on Halsted',
+        'address': '3656 N Halsted St, Chicago IL 60613, Conference Room 200 on the 2nd Floor'
+    }
+    meeting_year = ''  # This is used to track year in header elements & append to meetings
```
pjsier: This doesn't need to be an attribute on the spider. You can set `meeting_year = None` before starting the loop and then set it the same way and that should work

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class ChiSsa18Spider(CityScrapersSpider):
+    name = "chi_ssa_18"
+    agency = "Chicago Special Service Area #18 North Halsted"
+    timezone = "America/Chicago"
+    start_urls = ["https://northalsted.com/community/"]
+    location = {
+        'name': 'Center on Halsted',
+        'address': '3656 N Halsted St, Chicago IL 60613, Conference Room 200 on the 2nd Floor'
```
pjsier: To keep things consistent this can be "3656 N Halsted St, Conference Room 200, Chicago IL 60613"

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+                for split_string in split_test:
+                    meeting_date_match = re.search(r'([A-Z]\w{2,}\s\d\d?)', split_string)
+                    if meeting_date_match and ChiSsa18Spider.meeting_year:
+                        try:
+                            converted_date = \
+                                self._convert_date(
+                                    meeting_date_match.group(), ChiSsa18Spider.meeting_year
+                                    )
+                        except ValueError:
+                            continue
```
pjsier: If `converted_date` is required to keep iterating then instead of catching `ValueError` the `else` condition in here can be `continue`

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        table = response.xpath('//*[@id="page_content_wrapper"]/div/div/div/div[1]/div[3]')
+
+        self._validate_location(table)
+        self._validate_meeting_times(table)
+
+        for item in table.xpath('.//h3 | .//p'):
+            if 'h3' in item.get():
+                ChiSsa18Spider.meeting_year = item.xpath('descendant-or-self::text()')\
+                    .re_first(r'\d\d\d\d')
```
pjsier: A simpler syntax for the regex here is `r"\d{4}"`

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_links(self, item):
```
pjsier: To keep things a bit clearer you could pass `split_string` as the argument here instead of `links` and include the parsing in this function. Based on the above comment it should be a selector and not a string though

city_scrapers/spiders/chi_ssa_18.py
```diff
@@ -0,0 +1,117 @@
...
+        self._validate_location(table)
+        self._validate_meeting_times(table)
+
+        for item in table.xpath('.//h3 | .//p'):
+            if 'h3' in item.get():
+                ChiSsa18Spider.meeting_year = item.xpath('descendant-or-self::text()')\
+                    .re_first(r'\d\d\d\d')
+                continue
+            if '<p>' in item.get():
+                split_test = item.get().split("<br>")
```
pjsier: You might want to change the variable name here, and after splitting on `<br>` you should convert the strings back into selectors so that you can query them as HTML. Here's an example:

https://github.com/City-Bureau/city-scrapers-cle/blob/1b3d7d84c073249d5cedb2eb84f2b7eac7ece527/city_scrapers/spiders/cuya_soldiers_sailors_monument.py#L24-L27

tests/test_chi_ssa_18.py
```diff
@@ -0,0 +1,76 @@
...
+)
+spider = ChiSsa18Spider()
+
+freezer = freeze_time("2020-01-28")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+freezer.stop()
+"""
```
pjsier: This can be removed

tests/test_chi_ssa_18.py
```diff
@@ -0,0 +1,76 @@
...
+def test_time_notes():
+    assert parsed_items[0]["time_notes"] == ""
+
+
+def test_id():
+    assert parsed_items[0]["id"] == 'chi_ssa_18/202002050900/x/ssa_18_commission'
+
+
+def test_status():
+    assert parsed_items[0]["status"] == 'tentative'
```
pjsier: We should use the constant `TENTATIVE` here instead in case anything changes

tests/test_chi_ssa_18.py
```diff
@@ -0,0 +1,76 @@
...
+def test_source():
+    assert parsed_items[0]["source"] == 'https://northalsted.com/community/'
+
+
+def test_links():
+    assert parsed_items[0]["links"] == []
+
+
+def test_classification():
+    assert parsed_items[0]["classification"] == 'Commission'
```
pjsier: The constant should be used here  as well

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
nick-symon: I didn't make changes for this. I think I like the control flow better here since the 'else' would have to come after the creation of the meeting object. Here, converted date is required to create a meeting in the first place. Also, there are some h3 headers before the actual meeting items that actually return a string in response to the regex, so the 'try' here ensures that any value errors from passing the string doesn't stop the spider.

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
nick-symon: updated

tests/test_chi_ssa_18.py
nick-symon: updated

tests/test_chi_ssa_18.py
nick-symon: updated

tests/test_chi_ssa_18.py
nick-symon: updated

city_scrapers/spiders/chi_ssa_18.py
pjsier: Gotcha, I see what you're saying. I wasn't clear on this, but generally we try to exit the loop as soon as possible rather than nesting more because it's easy to read, so it wouldn't necessarily be the `else` condition, but more something like this:

```python
if not (meeting_date_match and meeting_year):
  continue

meeting = Meeting(...
```
This way, it's clear what we're looking for and the nesting is a little less deep. As for the errors, typically we want the scrapers to fail rather than behave in a way we aren't expecting. Otherwise we sometimes end up with bad data that isn't as easy to track as getting an error message. Let me know what you think, thanks!

nick-symon: This makes sense and the code is much easier to understand! I updated the code as you described and everything works out!

Issue #562, opened 2018-11-07
Spider: Chicago Special Service Area #23 Clark St/Lincoln Park


pjsier:
URL: visitclarkstreet.com/clark-street-ssa-commission-meetings-pages-182.php
Spider Name: `chi_ssa_23`
Agency Name: Chicago Special Service Area #23 Clark St/Lincoln Park

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

mericleac:
Hi! I'd like to try this for my first contribution.

pjsier:
@mericleac sounds good!

mericleac:
Visiting visitclarkstreet.com redirects to lincolnparkchamber.com. The new URL for this special service area's meetings seems to be https://www.lincolnparkchamber.com/clark-street-ssa-administration/.

zuzaman:
@pjsier! Hello! I would like to give this issue a go. I can see that the previous contributor already found the live link where the next meeting schedules are shared.

zuzaman:
So I played around a bit, as web scraping is a completely new thing for me, and now I have a mock version that that creates the meeting entries for the current year. 

As this is my first commit I would like to ask two questions:
1. Are the dates between 2016-2019 also interesting for the scraper? Or only the ones of the current year?
2. Pre-2020 meetings have in this case agendas and meeting minutes available in separate files. I am not sure if documents of the current year are also uploaded or that is only done later. Should both of them get attached as a link to the meeting entry?

pjsier:
@zuzaman sounds great! For the first question, they're interesting if they follow the same format, but we don't always want to slow down the spider by pulling them every day. We've been using a `CITY_SCRAPERS_ARCHIVE` setting to control pulling older meetings where it's easy, but if it introduces complications feel free to ignore them https://github.com/City-Bureau/city-scrapers/blob/c835f0cedd36683c2ea64e28e7c8359badfd3009/city_scrapers/spiders/chi_zoning_board.py#L33

For the second point, we do want those links, but if they aren't included now you can implement some basic checks off of how they look in the past. More info in the docs https://cityscrapers.org/docs/development/#links

zuzaman:
Ok thanks for the hints. I opened a pull request.

pjsier:
Closed by #949, thanks @zuzaman!

Issue #23, opened 2017-08-08
write spider creation docs


eads:

eads:
duplicated of #6

PR #949, opened 2020-03-09
#562 Chicago Special Service Area #23

zuzaman:
## Summary

**Issue:** #562

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

First pull request and first time working with scrapy. I am interested to hear your opinion. First started with CSS selectors, the realized how much XPATH makes things easier.

I tried to keep the checks regarding the document links from getting too "nested". Currently there are no documents uploaded for the year 2020 it. I find it hard to guess with what HTML formatting will this be done (if they will leave the ordered list and add the href for a list element or if an extra H4 will be made for it still during the same year), so I just assumed that only prior years have document links.

What do you think?

### DIFF
--- a/city_scrapers/spiders/chi_ssa_23.py
+++ b/city_scrapers/spiders/chi_ssa_23.py
@@ -24,7 +24,7 @@ class ChiSsa23Spider(CityScrapersSpider):
     def parse(self, response):
         # Due to the current lack of documents for the meetings of 2020
         # no assumption is made regarding the expected HTML page format
-        # once these are uploaded. 
+        # once these are uploaded.
         h4s = response.xpath('//h4')
         meeting_years = list()
 
@@ -56,7 +56,7 @@ class ChiSsa23Spider(CityScrapersSpider):
                         'start': start,
                         'end': end,
                         # Currently no meetings from 2020 have links
-                        'links': None
+                        'links': []
                     })
 
             # Always multiply the counter with two as each year has usually two links
@@ -78,7 +78,8 @@ class ChiSsa23Spider(CityScrapersSpider):
                     except IndexError:
                         # No problem, this means that this is an older entry and
                         # there is no second link to this element
-                        links.append(None)
+                        # links.append(None)
+                        pass
                     # check for old entries
                     if start < last_year and not self.settings.getbool("CITY_SCRAPERS_ARCHIVE"):
                         continue
--- a/city_scrapers/spiders/chi_ssa_23.py
+++ b/city_scrapers/spiders/chi_ssa_23.py
@@ -1,0 +1,144 @@
+# Just debugging
+import logging
+import re
+import unicodedata
+from datetime import datetime, timedelta
+
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa23Spider(CityScrapersSpider):
+    name = "chi_ssa_23"
+    agency = "Chicago Special Service Area #23 Clark Street"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.lincolnparkchamber.com/clark-street-ssa-administration/"]
+    location = {
+        "name": "Lincoln Park Chamber of Commerce",
+        "address": "2468 N. Lincoln Chicago, IL 60614"
+    }
+    # Each meeting takes place on Wednesday 4 PM
+    meeting_day = 'Wednesday'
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/chi_ssa_23.py
```diff
@@ -0,0 +1,143 @@
...
+
+            if year_counter == 0:
+                for item in h4s.xpath('following-sibling::ol[1]//li//text()').extract():
+                    start, end = self._parse_start_end(item, meeting_years[0])
+
+                    meetings.append({
+                        'start': start,
+                        'end': end,
+                        # Currently no meetings from 2020 have links
+                        'links': None
```
pjsier: This should return an empty list if no links are available, not `None`

city_scrapers/spiders/chi_ssa_23.py
```diff
@@ -0,0 +1,143 @@
...
+        # General meeting description is mentioned just after the H4 for the current year
+        general_desc = h4s.xpath('following-sibling::p[1]//em//text()').extract_first()
+
+        # List containing all meeting dictionaries
+        meetings = list()
+
+        link_collection = []
+        for h4 in h4s[1:]:
+            # Get the links for each meeting
+            link_collection.append(h4.xpath('following-sibling::p[1]//a'))
```
pjsier: The `link_collection` setup here is a bit hard to understand. It might be easier to understand as dictionaries where the key is the meeting date and the value is a list of links?

city_scrapers/spiders/chi_ssa_23.py
```diff
@@ -0,0 +1,143 @@
...
+                    # Try to see if we have a second link
+                    try:
+                        links.append(
+                            link_collection[2 * (year_counter - 1) +
+                                            1][item_counter].xpath('@href').extract_first()
+                        )
+                    except IndexError:
+                        # No problem, this means that this is an older entry and
+                        # there is no second link to this element
+                        links.append(None)
```
pjsier: We don't want to add `None` to any list of links. If a value is invalid we can leave it out

city_scrapers/spiders/chi_ssa_23.py
```diff
@@ -0,0 +1,143 @@
...
+    name = "chi_ssa_23"
+    agency = "Chicago Special Service Area #23 Clark Street"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.lincolnparkchamber.com/clark-street-ssa-administration/"]
+    location = {
+        "name": "Lincoln Park Chamber of Commerce",
+        "address": "2468 N. Lincoln Chicago, IL 60614"
+    }
+    # Each meeting takes place on Wednesday 4 PM
+    meeting_day = 'Wednesday'
```
pjsier: It might make sense to parse this from the page manually, or check if it's changed and raise an error if it has like we've done with `_validate_location` in a couple places like this: https://github.com/City-Bureau/city-scrapers/blob/d7d1fbbe1236c4353be03e289580e1421465c5e8/city_scrapers/spiders/chi_lsc_advisory.py#L80-L83

city_scrapers/spiders/chi_ssa_23.py
zuzaman: Makes sense, thanks. I changed this part.

city_scrapers/spiders/chi_ssa_23.py
zuzaman: Right, I changed that part to pass.

city_scrapers/spiders/chi_ssa_23.py
zuzaman: It took some time, but I got back to this today and changed quite much. The current version is based on dictionaries and seems to be more general.

city_scrapers/spiders/chi_ssa_23.py
zuzaman: I added this manual check, makes sense.

zuzaman:
It seems like since making the changes with the tests, the ones on github seems to fail even though with the same file they pass on my PC. I am a bit puzzled currently why this could be the case.

zuzaman:
Hmm even after making sure I only have the raw HTML file of the website, it seems that the automated tests return, possibly, the results in a different order than the pytest on my own PC. Do you have any idea why this might happen?

zuzaman:
@pjsier thanks that did the trick. There seems to be an issue with a connection timeout in test 3.7. Even though I can open the same URL without issue at the same time.

Issue #561, opened 2018-11-07
Spider: Chicago Special Service Area #22 Clark St/Andersonville


pjsier:
URL: http://www.andersonville.org/our-organizations/andersonville-ssa-22/
Spider Name: `chi_ssa_22`
Agency Name: Chicago Special Service Area #22 Clark St/Andersonville

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

annaderengowski:
I'd like to take a crack at this issue, if that's okay!

pjsier:
@annaderengowski sounds good!

zuzaman:
I would like a claim on this please.

pjsier:
@zuzaman I'll mark it claimed now

pjsier:
Closed by #953, thanks @zuzaman!

PR #953, opened 2020-05-15
Chicago Special Service Area #22 Clark St/Andersonville

zuzaman:
## Summary

**Issue:** #561

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

In this case the start time was one of 2 possible times still last week and now by today it seems they have changed this to "Start: Times vary" for the meetings. The information is in the newsletter. No length is known. How should I deal with the start and end times in this case?

### DIFF
--- a/city_scrapers/spiders/chi_ssa_22.py
+++ b/city_scrapers/spiders/chi_ssa_22.py
@@ -1,0 +1,124 @@
+import re
+import logging
+import dateutil.parser
+import unicodedata
+from datetime import datetime, timedelta
+
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa22Spider(CityScrapersSpider):
+    name = "chi_ssa_22"
+    agency = "Chicago Special Service Area #22 Clark St/Andersonville"
+    timezone = "America/Chicago"
+    start_urls = ["http://www.andersonville.org/our-organizations/andersonville-ssa-22/"]
+    location = {
+        "name": "Andersonville Chamber of Commerce",
+        "address": "5153 N. Clark St. #228 Chicago, Illinois 60640"
+    }
+    time = '9:30am'
+
+    def parse(self, response):
+
+        #address_text = response.xpath('//div[@class = "address"][1]/text()').extract()[1]
+        #self._validate_location(address_text)
+
+        h2s = response.xpath('//h2')
+
+        # General meeting description is mentioned just after the H4 for the current year
+        #general_desc = h4s.xpath('following-sibling::p[1]//em//text()').extract_first()
+        general_desc= 'TODO'
+
+        # Dictionary containing all meeting dictionaries
+        # The dates will be the keys
+        meetings = dict()
+
+        last_year = datetime.today().replace(year=datetime.today().year - 1)
+
+        for entry in h2s:
+            entry_str = entry.xpath('./text()').extract_first()
+            if entry_str and ('Meeting Schedule' in entry_str or 'Meetings & Minutes' in entry_str):
+                year = entry_str[0:4]
+                logging.debug(year)
+
+
+                # Only consider ps between two h4s
+                for item in entry.xpath('following-sibling::ul').xpath('./li'):
+
+                    # The  non-breaking space signals the end of the meeting lists
+                    #if li.xpath('./text()') and u'\xa0' in p.xpath('./text()').extract_first():
+                    #    break
+
+                    item_str = ' '.join(item.xpath('./text()').extract_first().split(' ')[0:1])
+                    start = self._parse_start(item_str, year)
+                    date = start.date()
+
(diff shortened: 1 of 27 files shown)


city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
+# import logging
```
pjsier, line 1: This can be removed

city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class ChiSsa22Spider(CityScrapersSpider):
+    name = "chi_ssa_22"
+    agency = "Chicago Special Service Area #22 Clark St/Andersonville"
+    timezone = "America/Chicago"
+    start_urls = ["http://www.andersonville.org/our-organizations/andersonville-ssa-22/"]
+    location = {
+        "name": "Andersonville Chamber of Commerce",
+        "address": "5153 N. Clark St. #228 Chicago, Illinois 60640"
```
pjsier, line 18: Small thing, but we've been using the state abbreviation instead of the name so Illinois should be IL

city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
...
+        # The dates will be the keys
+        meetings = dict()
+
+        last_year = datetime.today().replace(year=datetime.today().year - 1)
+
+        for entry_cnt, entry in enumerate(h2s, start=1):
+            entry_str = entry.xpath('./text()').extract_first()
+            if entry_str and ('Meeting Schedule' in entry_str or 'Meetings & Minutes' in entry_str):
+                year = entry_str[0:4]
+                # logging.debug(year)
```
pjsier, line 41: This and the other commented out logging calls can be removed

city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
...
+
+            if item['start'] < last_year and not self.settings.getbool("CITY_SCRAPERS_ARCHIVE"):
+                continue
+
+            meeting = Meeting(
+                title='Commission',
+                description=unicodedata.normalize("NFKD", general_desc),
+                classification=COMMISSION,
+                start=item['start'],
+                end=item['end'],
```
pjsier, line 76: If the end is always `None` then you can ignore it in the loop above and set it to `None` here directly to keep things simple

city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
...
+                time_notes='9:30am or 3:45pm (Please check our '
```
pjsier, line 77: We should probably refer people to the info@andersonville.org email instead, since it would be easy to miss the newsletter

city_scrapers/spiders/chi_ssa_22.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @staticmethod
+    def _parse_date(raw_date, year):
+        """Parse date."""
+        if raw_date:
+            return dateutil.parser.parse(year + ' ' + raw_date)
+
+    def _parse_start(self, item_str, year):
+        """Parse start datetime."""
+        start = self._parse_date(item_str, year)
```
pjsier, line 122: If we aren't sure of the start time it's probably safer to leave it as midnight. This means we can also remove the `time` property from the class

tests/test_chi_ssa_22.py
```diff
@@ -0,0 +1,131 @@
...
+        datetime(2019, 11, 12, 9, 30),
+        datetime(2019, 9, 10, 9, 30),
+        datetime(2019, 7, 9, 9, 30),
+        datetime(2019, 5, 21, 9, 30)
+    ]
+    for i in range(len(parsed_items)):
+        assert parsed_items[i]["start"] == expected_starts[i]
+
+
+"""
```
pjsier, line 63: This can be removed

zuzaman: Right. Removed it.

city_scrapers/spiders/chi_ssa_22.py
zuzaman: OK, thanks for the hint.

city_scrapers/spiders/chi_ssa_22.py
zuzaman: That does make sense. I changed this.

city_scrapers/spiders/chi_ssa_22.py
zuzaman: OK, thanks for the hint.

city_scrapers/spiders/chi_ssa_22.py
zuzaman: Still that is of course important. thanks for pointing it out.

city_scrapers/spiders/chi_ssa_22.py
zuzaman: True. I have removed them.

Issue #910, opened 2019-10-24
Spider: Cook County Emergency Telephone System Board


pjsier:
URL: http://www.cookcounty911.com/
Spider Name: cook_emergency_telephone
Agency Name: Cook County Emergency Telephone System Board

We'll need to parse PDFs for this one, and there's an example in [`chi_human_relations`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/chi_human_relations.py)

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

dalisaydavid:
👋 I'd like to take this one @pjsier!

pjsier:
@dalisaydavid sounds good!

cedarfall:
Is this still being worked on? If not, I'd like to work on it.

pjsier:
It hasn't been active in a while, so it's all yours @cedarfall

pjsier:
Closed by #954, thanks @cedarfall!

PR #954, opened 2020-05-29
0910 spider cook emergency telephone

cedarfall:
## Summary

**Issue:** #910 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions


cedarfall:
@pjsier Please help me with this. Only a flake8 test is failing because of a string being too long, I was thinking of using concatenation but I am not sure.

Link :
https://github.com/cedarfall/city-scrapers/blob/c74230abba0310ddcf58cebb683b706263ec4f4e/tests/test_cook_emergency_telephone.py#L69

city_scrapers/spiders/il_port_district.py
```diff
@@ -79,7 +79,7 @@ def parse_schedules(self, response):
 
                 agendas_links = (
                     self.agendas_dict.get((classification, start.strftime('%B %Y')), []) +
-                    self.agendas_dict.get((classification, start.strftime('%-m-%y')), [])
+                    self.agendas_dict.get((classification, start.strftime('%m-%y')), [])
```
pjsier, line 82: This should be kept out of this PR, it may be causing issues due to platform compatibility but it's easier to review if we keep the changes limited

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+        if schedule_link and self.docs_link:
+            yield scrapy.Request(
+                response.urljoin(schedule_link), callback=self._parse_schedule, dont_filter=True
+            )
+        else:
+            raise ValueError("Required links not found")
+
+    def _parse_schedule(self, response):
+        """Parse PDF and then yield to documents page"""
+        schedule_pdf_link = 'schedule.pdf'
```
pjsier, line 55: Since we're only using this variable once I think it'll be more straightforward to remove it and just request `response.urljoin('schedule.pdf')`

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        schedule_link = ""
+        for link in response.css(".sliding_box a"):
+            link_text = link.css("*::text").get()
+
+            schedule_link = self.start_urls[0] + '/wp-content/uploads/pdfs/'
```
pjsier, line 39: Does this need to be set inside the loop? I may have missed a piece of this, but it looks like it's a static link based on the start URL

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+
+            if "Minutes" in link_text:
+                self.docs_link = link.attrib["href"]
+            elif "Agenda" in link_text:
+                self.agenda_link = link.attrib["href"]
+
+        if schedule_link and self.docs_link:
+            yield scrapy.Request(
```
pjsier, line 47: Is this request necessary? It looks like `_parse_schedule` could request the page ending in "schedule.pdf" call `_parse_schedule_pdf` directly and then yield from `_parse_documents` instead of making a separate call

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+        DATE_PATTERN = r"[a-zA-z]{3,9} [0-9]{2},\s[0-9]{4} [\d]{2}[:][\d]{2} (?:p[.]m|a[.]m)"
+
+        for date_str in re.findall(DATE_PATTERN, clean_text):
+            self.meeting_starts.append(self._parse_start(date_str))
+
+    def _parse_documents(self, response):
+        """Parse agenda and minutes page"""
+        for start in self.meeting_starts:
+            meeting = Meeting(
+                title="Cook County Emergency Telephone System Board Public Meeting",
```
pjsier, line 90: In general, we're leaving off "Meeting" from the title if we can avoid it. "Emergency Telephone System Board" works here

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+
+            yield meeting
+
+    def _parse_description(self):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self):
+        """Parse or generate classification from allowed options."""
+        return FORUM
```
pjsier, line 113: `BOARD` seems like a better fit here

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+        schedule_link = ""
+        for link in response.css(".sliding_box a"):
+            link_text = link.css("*::text").get()
+
+            schedule_link = self.start_urls[0] + '/wp-content/uploads/pdfs/'
+
+            if "Minutes" in link_text:
+                self.docs_link = link.attrib["href"]
+            elif "Agenda" in link_text:
+                self.agenda_link = link.attrib["href"]
```
pjsier, line 44: Since they're always using the same URL for agendas even though the file changes, we should include a query parameter in the URL so that there's a way of distinguishing between copies (since this will be archived to the Wayback Machine either way). We're taking this approach for the Chicago Board of Election Commissioners:

https://github.com/City-Bureau/city-scrapers/blob/db6cba7919bd6450e8e74fb4cd2cceea135f5184/city_scrapers/spiders/chi_board_elections.py#L149

city_scrapers/spiders/cook_emergency_telephone.py
```diff
@@ -0,0 +1,135 @@
...
+        needs.
+        """
+        schedule_link = ""
+        for link in response.css(".sliding_box a"):
+            link_text = link.css("*::text").get()
+
+            schedule_link = self.start_urls[0] + '/wp-content/uploads/pdfs/'
+
+            if "Minutes" in link_text:
+                self.docs_link = link.attrib["href"]
```
pjsier, line 42: It looks like the current "/minutes" page is a list of meeting minutes. We should only attach it to a meeting if we can match the meeting date with a specific minutes link

city_scrapers/spiders/il_port_district.py
cedarfall: Yes, I'll remove it

city_scrapers/spiders/cook_emergency_telephone.py
cedarfall: No there is only one link. I was using the chi_human_relations as a template. I will remove the loop.

city_scrapers/spiders/cook_emergency_telephone.py
cedarfall: I was thinking of writing a helper function that checks for the meeting date and the minutes date and only then adds it? 
In chi_human_relations they don't check the dates but directly add all minutes links, that is why I used this approach. Which way should I go?

pjsier: That helper function sounds like the way to go. I'd have to double check, but I think the reason we did that in the other spider is because the minutes links were already associated with the date, and it wasn't one page that linked to all of them

city_scrapers/spiders/cook_emergency_telephone.py
cedarfall: Ok I will add it. But if someone visits this link, they will still see the upcoming agenda only, right?

cedarfall:
@pjsier I have made the changes. Please let me know if you approve? I will rebase and clean the commits.

city_scrapers/spiders/cook_emergency_telephone.py
pjsier: In general yes, but we're archiving everything so if it's stored when it's first generated we can link either to the Wayback Machine or a saved copy

city_scrapers/spiders/cook_emergency_telephone.py
pjsier: I might be misunderstanding the changes, but is `docs_link` still needed here?

pjsier:
No worries about rebasing, it's easier for me to see the changes as separate commits, and we've been doing squash and merge either way @cedarfall. For any other changes there's no need to rebase and force push

city_scrapers/spiders/cook_emergency_telephone.py
cedarfall: The `docs_link` refers to this page https://cookcounty911.com/minutes/. This page has links to the minutes. We then use this page to parse the links given on the page, if the date for these links match to our meeting we add to the `links_map` as `minutes_href` otherwise these are not added.

pjsier: Got it, thanks for clarifying! That works

PR #956, opened 2020-06-10
Update Pipenv

pjsier:
## Summary

Updates Pipenv now that it incorporates changes so that `flake8` doesn't break


pjsier:
Closing in favor of a PR that drops Python 3.5 support since that's causing issues here

Issue #911, opened 2019-10-24
Spider: Cook County Justice Advisory Council


pjsier:
URL: https://www.cookcountyil.gov/service/justice-advisory-council-meetings
Spider Name: cook_justice_advisory_council
Agency Name: Cook County Justice Advisory Council

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

haworku:
Hey!  I'm going to try to work on this. Just forked the repo and got things up and running.

pjsier:
@haworku sounds great! This one should be pretty similar to [`cook_board_ethics`](https://github.com/City-Bureau/city-scrapers/blob/master/city_scrapers/spiders/cook_board_ethics.py), but we'll also need to pull documents from the main link in the description

RebeccaCurry:
Hi, I'm new to contributing and there's a lot of libraries in this I'm not super familiar with, but I'd like to give this a shot. I know your project is listed as friendly to beginners. Hopefully I won't be putting that too much up to the test!

pjsier:
@RebeccaCurry all yours! Feel free to ask any questions that come up

bchao99:
Hi! It looks like there hasn't been any activity on this issue for the past 30 days so I was wondering if I could work on this.

pjsier:
@bchao99 sounds good!

pjsier:
Closed by #961

PR #961, opened 2020-06-16
0911 spider cook justice advisory

ledaliang:
## Summary

**Issue:** #911

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

### DIFF
--- a/city_scrapers/spiders/cook_justice_advisory.py
+++ b/city_scrapers/spiders/cook_justice_advisory.py
@@ -163,6 +163,16 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
     def _parse_links(self, response):
         """Parse links"""
         agenda_map = self.agenda_map
+
+        links = response.css('span.file a')
+        links = links[2:]
+        for index, link in enumerate(links):
+            link_name = link.xpath('text()').extract_first()
+            link_name = link_name.replace('\xa0', ' ')
+            link_path = link.xpath('./@href').extract_first()
+            
+
+        """
         files = response.css("span.file a")
         files = files[2:]
         for f in files:
@@ -175,3 +185,4 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
                 y = int(regex.group("y")) % 2000
                 m = int(regex.group("m"))
                 agenda_map[str(y) + "-" + str(m)] = [{"href": link, "title": "Agenda"}]
+        """
--- a/city_scrapers/spiders/cook_justice_advisory.py
+++ b/city_scrapers/spiders/cook_justice_advisory.py
@@ -18,7 +18,7 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
 
     def start_requests(self):
         toay = datetime.now()
-        url = 'https://www.cookcountyil.gov/service/justice-advisory-council-meetings'
+        url = 'https://www.cookcountyil.gov/calendar-node-field-date/month/{}'.format(mo_str)
         yield scrapy.Request(url=url, method='GET', callback=self.parse)
         
 
@@ -52,8 +52,13 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
 
     def _get_event_urls(self, response):
         """
-        Get urls for all board of ethics meetings on the page
+        Get urls for all justice advisory council (JAC in calendar) meetings on the page
         """
+        return [
+            response.urljoin(href)
+            for href in response.xpath('//a[contains(text(), "JAC")]'
+                                       ).css('a::attr(href)').extract()
+        ]
         
     def _parse_location(self, response):
         """
--- a/city_scrapers/spiders/cook_justice_advisory.py
+++ b/city_scrapers/spiders/cook_justice_advisory.py
@@ -48,9 +48,12 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
             time_notes="",
             all_day=self._parse_all_day(response),
             location=self._parse_location(response),
-            links=self._parse_links(response),
+            # links=self._parse_links(response),
             source=response.url,
         )
+
+        meeting["links"] = self._parse_links(meeting)
+
         meeting["id"] = self._get_id(meeting) # need a get_id and get_status method? these lines cause
         meeting["status"] = self._get_status(meeting)
         return meeting
@@ -142,20 +145,40 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
         date = start_end[0][: start_end[0].rindex(" ")]
         return datetime.strptime("{} {}".format(date, end_time), "%B %d, %Y %I:%M%p")
 
-    def _parse_links(self, response):
+    def _parse_links(self, meeting):
+        date = meeting['start']
+
         url = "https://www.cookcountyil.gov/service/justice-advisory-council-meetings"
 
-        link_response = scrapy.Request(url=url, method='GET', callback=self.parse)
+        yield scrapy.Request(url=url, method='GET', callback=self._parse_agenda, meta={'meeting_date': date})
 
-        files = link_response.css("span.file a::attr(href)").extract()
+
+        # links = link_response.css("span.file a::attr(href)").extract()
         # files = response.css("span.file a")
 
+        # dates = link_response.css("span.file a::attr(href)").re(r"((?:\d{1,2})(?:\.|_|-)(?:\d{1,2})(?:\.|_|-)(?:\d{2,4}))")
         # ((?:\d{1,2})(?:\.|_|-)(?:\d{1,2})(?:\.|_|-)(?:\d{2,4})) some regex fun :)
 
-        return [
-            {
-                "href": f.xpath("./@href").extract_first(),
-                "title": f.xpath("./text()").extract_first(),
-            }
-            for f in files
-        ]
+        # return [
+        #     {
+        #         "href": f.xpath("./@href").extract_first(),
+        #         "title": f.xpath("./text()").extract_first(),
+        #     }
+        #     for f in files
+        # ]
+       
+        # date = meeting['start'].strftime("%m/%d/%Y")
+
+    def _parse_agenda(self, response):
+        files = response.css("span.file a").extract()
+        meeting_date = response.meta.get('meeting_date')
+
+        for f in files:
+            link = f.xpath("./@href").extract_first()
+            title = f.xpath("./text()").extract_first()
+            date = datetime.strftime(link.re(r"((?:\d{1,2})(?:\.|_|-)(?:\d{1,2})(?:\.|_|-)(?:\d{2,4}))")) 
+
+            if (date == meeting_date.date()):
+                return [{ "href": link, "title": title, }]
+        return []
+
--- a/city_scrapers/spiders/cook_justice_advisory.py
+++ b/city_scrapers/spiders/cook_justice_advisory.py
@@ -15,15 +15,11 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
     allowed_domains = ["www.cookcountyil.gov"]
 
     def start_requests(self):
-        # old stuff:
-        # url = "https://www.cookcountyil.gov/service/justice-advisory-council-meetings"
-        # yield scrapy.Request(url=url, method="GET", callback=self.parse)
 
         today = datetime.now()
         for month_delta in range(-3, 0):
             mo_str = (today + relativedelta(months=month_delta)).strftime("%Y-%m")
             url = 'https://www.cookcountyil.gov/calendar-node-field-date/month/{}'.format(mo_str)
-            # print(url)
             yield scrapy.Request(url=url, method='GET', callback=self.parse)
 
     def parse(self, response):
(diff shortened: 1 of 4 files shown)


city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,143 @@
...
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class CookJusticeAdvisorySpider(CityScrapersSpider):
+    name = "cook_justice_advisory"
+    agency = "Cook County Justice Advisory"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cookcountyil.gov"]
+
+    def start_requests(self):
```
pjsier, line 15: This is the default behavior for `start_requests` when `start_urls` is supplied, so the URL can be moved into `start_urls` and this method removed

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if len(start_end) < 2:
+            return
+
+        end_time = start_end[1]
+        date = start_end[0][: start_end[0].rindex(" ")]
+        return datetime.strptime("{} {}".format(date, end_time), "%B %d, %Y %I:%M%p")
+
+    def _parse_links(self, response):
+        files = response.css("span.file a")
```
pjsier, line 136: This is only going to get documents on the event page itself, but we'll want the minutes and agendas if they match up to an event that are on https://www.cookcountyil.gov/service/justice-advisory-council-meetings

We'll want to create a mapping of months and years to links for those meetings and then reference that when pulling from events pages so that the agendas and minutes can be associated. Here's one example of something similar https://github.com/City-Bureau/city-scrapers/blob/19fd52d76aa314734f620c4dafa9be0a0228ba81/city_scrapers/spiders/il_labor.py#L28

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,143 @@
...
+            all_day=self._parse_all_day(response),
+            location=self._parse_location(response),
+            links=self._parse_links(response),
+            source=response.url,
+        )
+        meeting["id"] = self._get_id(meeting)
+        meeting["status"] = self._get_status(meeting)
+        return meeting
+
+    def _get_event_urls(self, response):
```
pjsier, line 48: The wording for the meetings is sometimes inconsistent, so searching for "JAC Council" might not always work. We'll also want to get some recent past meetings as well as more than the next two upcoming meetings if present, so we should probably iterate through the calendar page here

ledaliang: Since "JAC Council" might not always work, what would you suggest searching for when iterating through the calendar page?

pjsier: Checking "JAC" or "Justice Advisory" after extracting the text from a link and cleaning up extra spaces is probably safest. Using the selector for text might not work in some cases where there's extra text inside accidentally

city_scrapers/spiders/cook_justice_advisory.py
ledaliang: Should cancelled meetings also get scraped? There is an agenda link associated with the cancelled meeting in March.

ledaliang: Unrelated to the last comment, but is there any special way virtual meeting addresses should be handled? Right now the spider returns it as 'Microsoft Teams Chicago IL 60602'.

pjsier: Yep, we'll want to know when meetings are cancelled. And if it's a virtual location I think that probably goes better in the description and can be left blank in the location

ledaliang: I believe we have implemented all of the comments that you left, could you take another look at our code? Thanks!

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,178 @@
...
+from city_scrapers_core.spiders import CityScrapersSpider
+from dateutil.relativedelta import relativedelta
+
+
+class CookJusticeAdvisorySpider(CityScrapersSpider):
+    name = "cook_justice_advisory"
+    agency = "Cook County Justice Advisory"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cookcountyil.gov"]
+    agenda_map = defaultdict(list)
```
pjsier, line 17: Could you explain this a bit more? We've usually been initializing `defaultdict` objects in the `__init__` method to avoid any complications of when specifically they spider object is created. It also looks like there's a getter method in `_get_agenda` that could be avoided unless I'm missing something in the implementation

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,178 @@
...
+        """
+        for url in self._get_event_urls(response):
+            yield scrapy.Request(url, callback=self._parse_event, dont_filter=True)
+
+    def _parse_event(self, response):
+        """Parse the event page."""
+        agenda_map = self._get_agenda()
+        title = self._parse_title(response)
+        start = self._parse_start(response)
+        meeting_year = int(start.strftime("%y"))
```
pjsier, line 48: Instead of making two calls to `strftime` here, it might be quicker to just call once with a matching format like this

```python
links_key = start.strftime("%y-%m")
```

but that might require changing the implementation in `_parse_links`. Will comment there as well

city_scrapers/spiders/cook_justice_advisory.py
pjsier: I still think it would be safer to pull the string out of the selector and do multiple comparisons for "JAC" as well as "Justice Advisory" in case they change this up since using the text contains selector is pretty strict

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,178 @@
...
+    def _get_agenda(self):
+        return CookJusticeAdvisorySpider.agenda_map
+
+    def _parse_links(self, response):
+        """Parse links"""
+        agenda_map = self._get_agenda()
+        files = response.css("span.file a")
+        files = files[2:]
+        for f in files:
+            link = f.xpath("./@href").extract_first()
```
pjsier, line 170: It looks like there are some files that don't follow this pattern like the minutes for July 2017: https://www.cookcountyil.gov/sites/default/files/jac_council_minutes_finaldraft.pdf

It might be easier to check for a month in either "%b" or "%B" format and then the year in the text of the link rather than the `href` attribute. That way, it can be parsed to a datetime object and then converted back to a string of "%m-%Y" for easier matching in `_parse_event`

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -14,8 +14,10 @@ class CookJusticeAdvisorySpider(CityScrapersSpider):
     agency = "Cook County Justice Advisory"
     timezone = "America/Chicago"
     allowed_domains = ["www.cookcountyil.gov"]
-    agenda_map = defaultdict(list)
-
+    
+    def __init__(self, agenda_map=None):
+        self.agenda_map = defaultdict(list)
```
pjsier, line 19: We'll need to call the superclass method here too like this https://github.com/City-Bureau/city-scrapers/blob/19fd52d76aa314734f620c4dafa9be0a0228ba81/city_scrapers/spiders/chi_teacherpension.py#L22-L24

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -42,7 +44,7 @@ def parse(self, response):
 
     def _parse_event(self, response):
         """Parse the event page."""
-        agenda_map = self._get_agenda()
+        agenda_map = self.agenda_map
```
pjsier, line 47: Is the `agenda_map` variable necessary here when we can call `self.agenda_map` directly?

city_scrapers/spiders/cook_justice_advisory.py
mikejoo: I got the event title out of the selector and put some conditions/comparisons in there. Are there any more comparisons that might be useful to add? I wrote the code so that adding more comparisons is as simple as possible.

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+            if (
+                "jac" in event_title
+                or "justice advisory" in event_title
+                or ("justice" in event_title and "advisory" in event_title)
+            ):
+                href = event.xpath("./@href").extract_first()
+                responses.append(response.urljoin(href))
+        return responses
+
+        """
```
pjsier, line 87: Looks like this might have been left in by mistake

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+import scrapy
+from city_scrapers_core.constants import ADVISORY_COMMITTEE
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from dateutil.relativedelta import relativedelta
+
+
+class CookJusticeAdvisorySpider(CityScrapersSpider):
+    name = "cook_justice_advisory"
+    agency = "Cook County Justice Advisory"
```
pjsier, line 14: Sorry I missed this earlier, but this should be "Cook County Justice Advisory Council"

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+            '//span[@class="date-display-single"]/descendant-or-self::*/text()'
+        ).extract()
+        date = "".join(date).upper()
+        return "ALL DAY" in date
+
+    def _parse_title(self, response):
+        """Parse or generate event"""
+        title = "".join(response.css("h1::text").extract())
+        if "JAC Council Meeting" in title:
+            return "JAC Council Meeting"
```
pjsier, line 133: I should have caught this before, but this can be "Justice Advisory Council" and we can drop "Meeting"

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from dateutil.relativedelta import relativedelta
+
+
+class CookJusticeAdvisorySpider(CityScrapersSpider):
+    name = "cook_justice_advisory"
+    agency = "Cook County Justice Advisory"
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cookcountyil.gov"]
```
pjsier, line 16: We won't want `allowed_domains` here because we're using a separate middleware to submit requests to the Wayback Machine for archiving, and if it's included those will be blocked

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.agenda_map = defaultdict(list)
```
pjsier, line 19: Looking at this again, it looks like there are minutes as well as agendas, and we'll want to reflect that if we can here.

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+            pattern = r"(?P<m>[a-zA-Z]+)( *)(?P<y>\d{4})"
+            regex = re.search(pattern, link_name)
+            if regex is not None:
+                raw_monthyear = regex.group("m") + " " + regex.group("y")
+                if len(regex.group("m")) < 4:
+                    date_obj = datetime.strptime(raw_monthyear, "%b %Y")
+                else:
+                    date_obj = datetime.strptime(raw_monthyear, "%B %Y")
+                formatted_date = datetime.strftime(date_obj, "%y-%m")
+                self.agenda_map[formatted_date] = [
```
pjsier, line 194: Mentioned above, but looking again I'm realizing there are meeting minutes here as well as agendas that we'll want to get, and I think all that would be needed is to check if "Agenda" or "Minutes" is in the string text

city_scrapers/spiders/cook_justice_advisory.py
```diff
@@ -0,0 +1,196 @@
...
+    timezone = "America/Chicago"
+    allowed_domains = ["www.cookcountyil.gov"]
+
+    def __init__(self, *args, **kwargs):
+        self.agenda_map = defaultdict(list)
+        super().__init__(*args, **kwargs)
+
+    def start_requests(self):
+        url = "https://www.cookcountyil.gov/service/justice-advisory-council-meetings"
+        yield scrapy.Request(
```
pjsier, line 24: I had missed this before, but to make sure that this is populated before getting to the detail page we'll need to chain it in separate methods. So this could have a new callback `_parse_meetings_page` that looks something like this:

```python
def _parse_meetings_page(self, response):
  self._parse_links(response)
  ... # The rest of it would be everything below in `start_requests`
```

Looking at it again, it could be even simpler to make the piece above the `parse` method, set `start_urls` to `["https://www.cookcountyil.gov/service/justice-advisory-council-meetings"]`, and change what is currently `parse` to something like `_parse_events_page`.

The reason this is important is currently there's the potential for one of the results pages to yield before the links are populated, and by chaining them this way we can be sure that the link map is filled before we try to pull from it

ledaliang:
We made some changes based off of you comments. We changed the agenda map to hold both agendas and minutes as well as added `_parse_events_page` and `_parse_meetings_page`.

pjsier:
I fixed one small thing and now it should be good to go, thanks so much!

Issue #572, opened 2018-11-07
Spider: Chicago Special Service Area #38 Northcenter


pjsier:
URL: http://www.northcenterchamber.com/pages/MeetingsTransparency1
Spider Name: `chi_ssa_38`
Agency Name: Chicago Special Service Area #38 Northcenter

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

zarifmahmud:
Could I try this one?

pjsier:
It's all yours!

rosemarymomoh:
hi @pjsier, looks like there hasn't been an update on this for a while now, mind if I take a stab at it?

pjsier:
@rosemarymomoh go for it!

jaspsingh:
Hey @pjsier, I'd be interested in helping with this issue. Could I claim it?

pjsier:
Hi @jaspsingh, thanks for your interest! It's our policy right now to limit contributors to [one issue at a time](https://github.com/City-Bureau/city-scrapers/blob/master/CONTRIBUTING.md#1-find-an-open-issue-to-work-on), so let me know which of the issues you commented on you'd like to start with.

jaspsingh:
Hey @pjsier, I'd like to start with this one.

PR #962, opened 2020-06-24
0572 spider chi ssa 38

jaspsingh:
## Summary

**Issue:** #[572](https://github.com/City-Bureau/city-scrapers/issues/572) 

## Checklist

- [ ] Tests are implemented
- [ ] All tests are passing

## Questions

### DIFF
--- a/city_scrapers/spiders/chi_ssa_38.py
+++ b/city_scrapers/spiders/chi_ssa_38.py
@@ -1,5 +1,5 @@
-from datetime import datetime
 import re
+from datetime import datetime
 
 from city_scrapers_core.constants import NOT_CLASSIFIED
 from city_scrapers_core.items import Meeting
@@ -60,7 +60,7 @@ class ChiSsa38Spider(CityScrapersSpider):
         date_words = date_str.split()
         month_name = date_words[0]
         datetime_obj = datetime.strptime(month_name, "%B")
-        meeting_day = int(re.findall('[0-9]+', date_words[1])[0])
+        meeting_day = int(re.findall("[0-9]+", date_words[1])[0])
 
         # Meeting time defaults to 12:00 AM
         meeting_hour = 0
@@ -69,18 +69,21 @@ class ChiSsa38Spider(CityScrapersSpider):
 
         # Handle irregularly formatted parts of date
         for i in range(len(date_words)):
-            # Meeting time appears in different formats, but always one index before "a.m." or "p.m."
+            # Meeting time appears in different formats, but always one
+            # index before "a.m." or "p.m."
             if "a.m." in date_words[i]:
                 meeting_hour = int(date_words[i - 1])
             elif "p.m." in date_words[i]:
                 meeting_hour = int(date_words[i - 1]) + 12
 
             # Validate reasonable year exists if it starts with "20"
-            if date_words[i][0:2] is "20":
-                meeting_year = int(date_words[i])
+            if date_words[i][0:2] == "20":
+                meeting_year = int(re.findall("[0-9]+", date_words[i])[0])
 
         # Put time, day, and year into datetime object
-        start_time = datetime_obj.replace(hour = meeting_hour, day = meeting_day, year = meeting_year)
+        start_time = datetime_obj.replace(
+            hour=meeting_hour, day=meeting_day, year=meeting_year
+        )
 
         return start_time
 
--- a/city_scrapers/spiders/chi_ssa_38.py
+++ b/city_scrapers/spiders/chi_ssa_38.py
@@ -1,0 +1,113 @@
+from datetime import datetime
+import re
+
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+        return title
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        description = ""
+        return description
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
```
pjsier, line 53: This should be `COMMISSION` for all meetings on this spider

city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        title = "Chamber of Commerce"
+        return title
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        description = ""
+        return description
```
pjsier, line 49: It's fine to just return "" instead of setting a variable first

city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        title = "Chamber of Commerce"
+        return title
```
pjsier, line 44: Mentioned in `_parse_description`, but it's fine to just return the string without assigning to a variable first. It's a bit odd for SSAs, but this one should be "Commission". They're technically separate entities managed by a nonprofit

city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _parse_location(self, item):
+        # Meetings seemingly ocurred at
+
+        return {
+            "address": "4054 N Lincoln Ave, Chicago, IL 60618",
+            "name": "Northcenter Chamber of Commerce",
+        }
+
+    def _parse_links(self, item):
```
pjsier, line 110: We'll need to parse a mapping of dates to relevant links from the page so that things like meeting minutes can be associated with the meetings listed. Here's an example of that:

https://github.com/City-Bureau/city-scrapers/blob/c6771d58209fa9ade4aa3768fa3ddb23b45c09fc/city_scrapers/spiders/chi_il_medical_district.py#L109-L140

city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        # Meetings seemingly ocurred at
```
pjsier, line 103: It looks like an abbreviated version of this is in the meeting item, since we don't have examples of another format we can do something like this to return a default location if "4054" is present and raise an exception otherwise:

```python
if "4054" not in item.extract():
  raise ValueError("Meeting location has changed")
return {
  "address": "4054 N Lincoln Ave, Chicago, IL 60618",
  "name": "Northcenter Chamber of Commerce",
}
```

city_scrapers/spiders/chi_ssa_38.py
```diff
@@ -0,0 +1,116 @@
...
+        description = ""
+        return description
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        date_str = item.extract()
```
pjsier, line 57: It looks like we might be able to simplify this a bit, and we'll also need to handle situations where minutes may be supplied for the time. Haven't tested this, but something like this snippet could work:


```python
item_str = item.extract()
month_day_str = re.search(r"[A-Z][a-z]{2,9} \d{1,2}", item_str).group()

year_str = re.search(r"\d{4}", item_str).group()
if not year_str[:2] == "20":
  year_str = str(datetime.today().year)  # Default to current year

time_match = re.search(r"\d{1,2}(\:\d\d) [apm\.]{2,4}", item_str)  # We want to check for a minutes portion here
time_str = "12 am"
if time_match:
  time_str = time_match.group().replace(".", "")

time_fmt = "%I %p"
if ":" in time_str:
  time_fmt = "%I:%M %p"

return datetime.strptime(f"{month_day_str} {year_str} {time_str}", f"%B %d %Y {time_str}")
```

PR #964, opened 2020-06-27
0905 spider chi ssa 4

ledaliang:
## Summary

**Issue:** #905 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing


city_scrapers/spiders/chi_ssa_4.py
```diff
@@ -0,0 +1,137 @@
...
+                urls.append(link)
+        return urls
+
+    def _parse_title(self, response):
+        """Parse or generate meeting title."""
+        return "".join(response.css("div.tribe-events-single h1::text").get())
+
+    def _parse_description(self, response):
+        """Parse or generate meeting description."""
+        selector = "div.tribe-events-single-event-description p::text"
```
pjsier, line 76: It looks like this is always the same so it should be fine to ignore and return an empty string instead

city_scrapers/spiders/chi_ssa_4.py
```diff
@@ -0,0 +1,137 @@
...
+        ).extract()
+
+        for link, description in zip(links, descriptions):
+            if "meeting" in description.lower():
+                urls.append(link)
+        return urls
+
+    def _parse_title(self, response):
+        """Parse or generate meeting title."""
+        return "".join(response.css("div.tribe-events-single h1::text").get())
```
pjsier, line 72: We can remove "Meeting" from the end of this string

city_scrapers/spiders/chi_ssa_4.py
ledaliang: In this case, do you think it would be more efficient to get rid of the `_parse_description` method entirely and put `description=""` in the `_parse_event` method?

pjsier: That works! That's usually what I end up doing, but it's a preference thing

PR #966, opened 2020-07-17
0672 spider chi northwest home equity

SubtleHyperbole:
## Summary

**Issue:** #672  

## Checklist

- [X] Tests are implemented
- [X**] All tests are passing
- [X] Code comments from template removed

## Questions

I am such a newb when it comes to using github. After royally screwing things up not once but twice, and having to literally delete the entire repository and start over, then re-adding my files, I think I got it down this time.

One thing though is odd, and its why i included the asterisk * next to "All tests are passing" -- for some reason this final time that I re-cloned and forked the repository onto my machine, it started throwing errors concerning three spiders (unrelated to the one i'm working on) which kept throwing exceptions up for trying to import a library called pdfminer.  All spiders which used these two import lines:

>from pdfminer.high_level import extract_text_to_fp
>from pdfminer.layout import LAParams

These lines threw exceptions about there being no existing library, even though I do have pdfminer installed on my main python as well as the pipenv shell.  For some reason this was suddenly an issue.  I had to actually comment out the two lines on the three spiders in order to get the terminal to finish setting up my fork.  I then returned the lines to normal, but later when it came time to run pytests, it threw up errors there too for those three spiders again.  

Anyhoo, long winded way of saying that what I mean by the ** in the checklist is that, yes, for all MY tests for the new spider are passing.  Someone else's spiders that use pdfminer, however, are having trouble all of a sudden. The three spiders are:

il_pollution_control
chi_human_relations
cook_emergency_telephone

(none of those three spider files or tests got added to my commit/push anyway, I'm just bringing it up here because I thought it was an odd error, especially since it didn't happen when i did this a day or so ago, however it seems to have nothing to do with what I am working on)

### DIFF
--- a/city_scrapers/spiders/chi_northwest_home_equity.py
+++ b/city_scrapers/spiders/chi_northwest_home_equity.py
@@ -29,9 +29,8 @@ class ChiNorthwestHomeEquitySpider(CityScrapersSpider):
             "3234 N. Central Ave.": "Northwest Home Equity Assurance Program",
             "5363 W. Lawrence Ave": "Jefferson Park Library",
         }
-        # Before we begin, need to collect meeting minutes data -- which for some reason is housed
-        # on entirely different pages than the list of meetings itself. This list will be held
-        #  in variable article_all_pages
+        # Before we begin, need to collect meeting minutes data -- which is
+        #  contained on multiple iterative pages (/page/2 /page/3 etc)
         r = requests.get("https://nwheap.com/category/meet-minutes-and-agendas/")
         article_response = scrapy.http.HtmlResponse(r.url, body=r.content).xpath(
             "//article"
--- a/city_scrapers/spiders/chi_northwest_home_equity.py
+++ b/city_scrapers/spiders/chi_northwest_home_equity.py
@@ -1,0 +1,164 @@
+import re
+from datetime import datetime
+
+import requests
+import scrapy
+from city_scrapers_core.constants import NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiNorthwestHomeEquitySpider(CityScrapersSpider):
+    name = "chi_northwest_home_equity"
+    agency = "Chicago Northwest Home Equity Assurance Program"
+    timezone = "America/Chicago"
+    start_urls = ["https://nwheap.com/category/meet-minutes-and-agendas/"]
+
+    def parse(self, response):
+        self.location_dict = {
+            "3022 N. Harlem": "3022 N. Harlem Ave #2S, Chicago, IL 60634",
+            "3022 N. Harlem Ave. #2S": "3022 N. Harlem Ave #2S, Chicago, IL 60634",
+            "3022 Harlem Ave. #2S": "3022 Harlem Ave #2S, Chicago, IL 60634",
+            "3234 N. Central Ave.": "3234 N. Central Ave, Chicago, IL 60634",
+            "5363 W. Lawrence Ave.": "5363 W. Lawrence Ave, Chicago, IL 60630",
+        }
+        self.location_name_dict = {
+            "3022 N. Harlem": "Northwest Home Equity Main Office",
+            "3022 N. Harlem Ave #2S": "Northwest Home Equity Main Office",
+            "3022 Harlem Ave #2S": "Northwest Home Equity Main Office",
+            "3234 N. Central Ave.": "Northwest Home Equity Assurance Program",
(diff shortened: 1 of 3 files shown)


SubtleHyperbole:
I am confused about the "status" and "id" tests in the template -- no parse functions for these exist in the template like they do for every other test.  Are they necesarry? And where would they even come from? "status" sounds like the same information that's contained in "description" -- and I this particular site doesn't give each meeting a unique ID other than the date and time and title...

city_scrapers/spiders/chi_northwest_home_equity.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class ChiNorthwestHomeEquitySpider(CityScrapersSpider):
+    name = "chi_northwest_home_equity"
+    agency = "Chicago Northwest Home Equity Assurance Program"
+    timezone = "America/Chicago"
+    start_urls = ["https://nwheap.com/category/meet-minutes-and-agendas/"]
+
+    def parse(self, response):
+        self.location_dict = {
```
pjsier, line 18: These can be set above by `start_urls` instead of each time `parse` is called

city_scrapers/spiders/chi_northwest_home_equity.py
```diff
@@ -0,0 +1,163 @@
...
+        self.location_name_dict = {
+            "3022 N. Harlem": "Northwest Home Equity Main Office",
+            "3022 N. Harlem Ave #2S": "Northwest Home Equity Main Office",
+            "3022 Harlem Ave #2S": "Northwest Home Equity Main Office",
+            "3234 N. Central Ave.": "Northwest Home Equity Assurance Program",
+            "5363 W. Lawrence Ave": "Jefferson Park Library",
+        }
+        # Before we begin, need to collect meeting minutes data -- which is
+        #  contained on multiple iterative pages (/page/2 /page/3 etc)
+        r = requests.get("https://nwheap.com/category/meet-minutes-and-agendas/")
```
pjsier, line 34: We don't need to use `requests` here since the response from this page should be the first thing Scrapy loads in `start_urls`

pjsier:
@SubtleHyperbole to your question about ID and status, we've got more information on those [in our docs](https://cityscrapers.org/docs/development/#id), but in general they're a part of the [Open Civic Data Event specification](https://opencivicdata.readthedocs.io/en/latest/data/event.html) that we're following, and status in particular is a more structured way of tracking whether meetings are cancelled, upcoming, or in the past

SubtleHyperbole:
So the successive requests are actually only needed to get links for files posted AFTER meetings which have already happened (like meetings minutes).  

Following the restructuring of the website itself (which my email to their admin caused unintended), the only location where it lists future meetings dates is on the side bar to the right which exists on several different pages.

This side bar contains (at least right now) the next three future meetings, and 5-6 past meetings.  While each meeting on the sidebar contains a link to its own page, that page actually contains no new information that isn't already in the side bar.  

The only information that I figured would be of interest to scrape is the meetings minutes and other uploaded files for those 5-6 PAST meetings on the side bar.  For some reason, the link for each meeting on the side bar doesn't actually lead to those files.  Instead, separate pages for each meeting is listed on that nwheap/meeting-minutes-and-agenda main page, along with multiple additional pages in groups of ten.

So here is the thing, since theres only 5-6 meetings listed on that right sidebar now, all of their meeting-minutes-side-pages are listed on the first page of results.  As it is right now, while I iterated through all the existing back pages, those include files for past meetings that go way way further back.  I thought maybe I should use that list to generate the list of meetings, but unfortunately other than the date, it doesn't give any other information except the associated links. 

So should I just ditch the iterative page scroll, given that, at least for right now, nothing except the first page holds any data that ends up in a yielded meeting?

SubtleHyperbole:
Also i'm still not sure about where the _id and _status comes from, but really as long as it's not something I'm supposed to be scraping from the website, that's good by me.

SubtleHyperbole:
Also, I think my first comment might be confusing

        # Now moving onto the main parse of the meetings list

should really say something more like

        # First create a list of all potential meetings minutes/associated files pages. Later in the main parse, we will cycle through this list to see if there is an associated files page containing stuff like meeting minutes files.  Only past meetings have these, and even then not every time.

pjsier:
@SubtleHyperbole thanks for the clarification, but I think we'll still want to use Scrapy here, even if we're chaining together requests that aren't pulling links directly from a page. Here's an example of where we're doing this on [`il_elections`](https://github.com/City-Bureau/city-scrapers/blob/b1c9c8a69aea9be3cb05a5a6fd6d5622c1bcba0e/city_scrapers/spiders/il_elections.py#L15-L37)

PR #967, opened 2020-07-20
0770 Spider: Cook County Electoral Board (Suburban Cook)

turnout-coding:
## Summary

**Issue:** 0770

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

* I'm not sure how far back or forward to be fetching meetings when not in archive mode. This is an .aspx page, so each meeting is an additional request. I've implemented it to fetch the previous year if we're still in March or earlier, and the next year if it's available, but can change those if there's a standard approach.

* I've added some tests, but don't have any chained together from the original request to the subsequent ones. Is there a need for that kind of test?

### DIFF
--- a/city_scrapers/spiders/cook_electoral.py
+++ b/city_scrapers/spiders/cook_electoral.py
@@ -1,0 +1,150 @@
+from city_scrapers_core.constants import BOARD
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from scrapy import FormRequest
+from scrapy.http.cookies import CookieJar
+from datetime import datetime
+import re
+import scrapelib
+
+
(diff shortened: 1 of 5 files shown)


pjsier:
@turnout-coding thanks so much for the PR! I'm really sorry about this, but I think we got our wires crossed posting the issue initially because looking at it again it's an alternate page for the Board of Commissioners. It also looks like the link for the Electoral Board is only intermittently updated and probably not best for a scraper right now https://www.cookcountyclerk.com/service/electoral-board

That's our bad, and you're welcome to take another scraper! This was really well done though, I haven't used the `CookieJar` class before and it looks interesting here

turnout-coding:
@pjsier No worries, I learned a bunch working on this one regardless! I think I'll find another scraper to do instead and comment on whatever issue I find -- thanks!

Issue #673, opened 2019-02-03
Spider: Chicago O'Hare Noise Compatibility Commission


pjsier:
URL: https://www.oharenoise.org/about-oncc/oncc-meetings/month.calendar/2019/02/03/-
Spider Name: `chi_ohare_noise`
Agency Name: Chicago O'Hare Noise Compatibility Commission

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

GeorgeDubuque:
Hello! I'm a first time open-source contributor. I have some experience scraping, can I try this one?

pjsier:
@GeorgeDubuque go for it! Just saw that you filled out the Slack form so I'll add you there too

GeorgeDubuque:
Awesome! Sounds good to me!

GeorgeDubuque:
@pjsier I am having trouble installing requirments.txt with pip I get the error ```ERROR: Could not find a version that satisfies the requirement city-scrapers-core[azure]==0.4.2 (from -r requirements.txt (line 11)) (from versions: none)
ERROR: No matching distribution found for city-scrapers-core[azure]==0.4.2 (from -r requirements.txt (line 11))```

stmfunk:
Hi there, I'm a first timer to open source contributions. Looking for a good place to start and this seems fairly well established so I'd love to get involved!

pjsier:
@stmfunk welcome! If you're interested in this one I'll mark it claimed

stmfunk:
I've created the pull request. The spider returns data from the two sources sequentially. I tried merging them on the fly but the data between them is inconsistent so it was impossible to match them on pass. I think it might be possible with  a database operation.

stefovego:
I would like to write this scrapper.

alexfoss8:
Hello! I'm a first time open-source contributor. I have some experience scraping, can I try this one?

creepymarshmallow117:
Hey!! I'm a first time contributor and would like to work on this, I checked out the links and the https://www.oharenoise.org/about-oncc/agendas-and-minutes
page seems to have been removed from the site.

PR #972, opened 2020-09-09
0673 spider chi ohare noise

stmfunk:
## Summary

**Issue:** #673 

Finished creating spider for scraping data from the Chicago O'Hare Noise Compatibility Commission.

## Checklist

- [ ] Tests are implemented
- [ ] All tests are passing


city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+from city_scrapers_core.spiders import CityScrapersSpider
+from scrapy import Request
+
+
+class ChiOhareNoiseSpider(CityScrapersSpider):
+    name = "chi_ohare_noise"
+    agency = "Chicago O'Hare Noise Compatibility Commission"
+    timezone = "America/Chicago"
+
+    class ChiOhareNoiseSubSpider1:
```
pjsier, line 22: Thanks for picking this one up! Looks like it was one of the more difficult ones. I think we'll want to keep it limited to one class here though for readability and split out methods for each (something like `_parse_events_start`, `_parse_docs_start`).

I could be missing some complexity here though, is there a reason for not doing that?

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+        def parse(self, response):
+            """
+            `parse` should always `yield` Meeting items.
+
+            Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+            needs.
+            """
+            for item in response.css(".ev_td_li"):
+                surl = self._parse_url(item)
+                yield Request(url=response.urljoin(surl), callback=self._parse_details)
```
pjsier, line 32: Might be simpler to use the `response.follow` shortcut method here like you did below

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+
+            next_page = response.xpath("//div[@class='previousmonth']/a/@href").get()
+            if next_page is not None:
```
pjsier, line 35: We might want to limit this to only calling the next page once so that we don't hammer the site with a ton of requests every time.

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+                description=self._parse_description(response),
+                start=stime,
+                end=stime + timedelta(hours=1),
+                all_day=False,
+                location=self._parse_location(response),
+                links=self._parse_links(response),
+                source=response.url,
+            )
+
+            meeting = self._get_status(meeting)
```
pjsier, line 51: We'll want to pass the `text` kwarg with the raw title here so that we can check whether or not it contains "cancelled" which is handled for you in `get_status`.

Also, we should make sure that `get_status` is assigning `meeting['status']` and not overwriting the meeting variable entirely. That goes for `_parse_classification` as well

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+            if next_page is not None:
+                yield response.follow(response.urljoin(next_page), callback=self.parse)
+
+        def _parse_details(self, response):
+            stime = self._parse_start(response)
+            meeting = Meeting(
+                title=self._parse_title(response).replace("CANCELLED ", "").strip("- "),
+                description=self._parse_description(response),
+                start=stime,
+                end=stime + timedelta(hours=1),
```
pjsier, line 44: We can leave this as `None` if we aren't parsing it and potentially pass a string to the `time_notes` kwarg with any details about scheduling

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+            return meeting
+
+        def _parse_title(self, response):
+            """Parse or generate meeting title."""
+            return response.xpath(
+                "//div[@class='jev_evdt_header']/div/h2/text()"
+            ).extract()[0]
+
+        def _parse_url(self, response):
+            """Parse or generate meeting title."""
```
pjsier, line 62: The docstring can be deleted here if it doesn't apply

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+            return {
+                # Using reverse indexing for the cases
+                # where there is no building name or no location
+                "address": addr[-1].strip(),
+                "name": addr[0].strip(),
+            }
+
+        def _parse_links(self, response):
+            """Parse or generate links."""
+            return [
```
pjsier, line 115: This might not always be a Zoom link, so we'll want to have checks for that here

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+                for item in response.xpath(
+                    "//div[@class='jev_evdt_desc']/p/a/@href"
+                ).extract()[:1]
+            ]
+
+        def _parse_source(self, response):
+            """Parse or generate source."""
+            return response.url
+
+        def _get_status(self, meeting):
```
pjsier, line 126: Missed this when looking at the implementation above, but this functionality is all contained in the `CityScrapersSpider` `get_status` method if you pass the title text to check for cancellation info to the `text` kwarg

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+                    start=self._parse_start(item),
+                    links=self._parse_links(item, response),
+                    source=self._parse_source(response),
+                )
+
+                meeting = self._get_status(meeting)
+                meeting = self._parse_classification(meeting)
+                yield meeting
+            next_page = response.xpath("//li[@class='pagination-next']/a/@href").get()
+            if next_page is not None:
```
pjsier, line 156: Same comment as above, we can limit this to only getting the next page once if there's a `_parse_page` method that is called here instead of running all of the logic in `parse`

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+            return datetime.strptime(
+                item.css(".djc_producer span").xpath("text()").extract()[0], "%B %d, %Y"
+            )
+
+        def _parse_links(self, item, response):
+            """Parse or generate links."""
+            links = item.xpath('td[@class="djc_price"]/div/ul/li/a')
+            return [
+                {
+                    "href": response.url
```
pjsier, line 174: Is there a reason for recreating the query parameters here? It looks like `response.urljoin(link.attr['href'])` could work

city_scrapers/spiders/chi_ohare_noise.py
```diff
@@ -0,0 +1,212 @@
...
+            return response.xpath("//div[@class='jev_evdt_desc']/p/text()").get() or ""
+
+        def _parse_classification(self, meeting):
+            """Parse or generate classification from allowed options."""
+            if "committee" in meeting["title"].lower():
+                meeting["classification"] = COMMITTEE
+            elif "commission" in meeting["title"].lower():
+                meeting["classification"] = COMMISSION
+            else:
+                meeting["classification"] = NOT_CLASSIFIED
```
pjsier, line 76: Since the overall organization is a commission it's safe to default to `COMMISSION` instead of `NOT_CLASSIFIED`

logput
```diff
@@ -0,0 +1,890 @@
+2020-08-28 21:56:35 [scrapy.utils.log] INFO: Scrapy 2.1.0 started (bot: city_scrapers)
```
pjsier, line 1: This can be removed

output
```diff
@@ -0,0 +1,20 @@
+[]
```
pjsier, line 1: This can be removed as well

tests/test_chi_ohare_noise.py
```diff
@@ -0,0 +1,451 @@
...
+
+freezer.stop()
+
+
+def test_tests():
+    assert len(parsed_items) == 45
+
+
+parsed_sub_items = []
+for i in range(5):
```
pjsier, line 35: Ideally we should try not to have more than 2 test files here, and running some logic in `parse` in a separate method that doesn't yield a request could make that easier. That way we could call something like `_parse_detail` without needing to worry about testing pagination

tests/test_chi_ohare_noise.py
```diff
@@ -0,0 +1,451 @@
...
+    assert parsed_sub_items[4]["location"] == {
+        "name": "Chicago Dept. of Aviation Building",
+        "address": "10510 W. Zemke Road, Chicago, IL, Conference Room 1",
+    }
+
+
+def test_source():
+    src = "https://www.oharenoise.org/about-oncc"
+    src += "/oncc-meetings/month.calendar/2019/02/03/-"
+    assert parsed_sub_items[0]["source"] == src
```
pjsier, line 122: This one is pretty safe, and we can shorten it a bit and get at what we're testing with `== test_response.url`

pjsier (changes requested):
Thanks for the PR! This one has been challenging for a while, so appreciate you taking it on. I left some initial comments here, let me know what you think or if any of the comments aren't clear

Issue #676, opened 2019-02-03
Spider: Chicago Southwest Home Equity Commission I


pjsier:
URL: https://swhomeequity.com/agenda-%26-minutes
Spider Name: `chi_southwest_home_equity_i`
Agency Name: Chicago Southwest Home Equity Commission I

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

mattpair:
Hi, just cloned the repo. I'm going to make a branch for this spider

mattpair:
@pjsier looks like most info besides the date is contained in a pdf. Let's discuss how to proceed but since this is my first, I think I'll start work on a more straightforward one.

pjsier:
@mattpair that approach makes sense to me, let me know if you need any help finding a clearer spider to work on but we have a good amount available

mattpair:
I'll resume work on this one

haidtang:
If this issue is unclaimed, I would like to work on that.

pjsier:
@haidtang sure! In general we like to contributors to stick to one issue at a time, so I'll assign you to this one for now and not the other. Let me know if you'd like to switch that though

haidtang:
@pjsier I got it. Could you please switch me to the other issue #566, I think that I have a better clue on how to deal with that one. Thank you so much.

pjsier:
Sure thing!

egfrank:
Hey is this issue unclaimed? I'd be happy to work on this one if so. Also, it seems like it will involve reading pdfs; has that been done within this project before?

pjsier:
It looks like it's been inactive for more than a month, so it's all yours if you're interested!

egfrank:
https://docs.scrapy.org/en/xpath-tutorial/topics/xpath-tutorial.html

egfrank:
Hi @pjsier, most of information for these meetings is included in PDFs of the minutes and agenda for each meeting. Are there any other spiders that have downloaded / parsed PDFs in this project already?

Just working on my own computer, I'm able to parse the files using the package `pdfminer.six` which I just found via googling, but was the most highly recommend Python package for reading PDFs that I could. That package also requires that the files are downloaded, so I'm using `tempfile` which should delete the files after the text is extracted from them. 

My question though is do you want to introduce that new package into the project? And is the fact that it needs to download files going to be a problem for running the spider on different computers?

pjsier:
[...]

PR #973, opened 2020-09-16
0676 spiderchi southwest home equity i

egfrank:
## Summary

**Issue:** #676

I worked on this back in February but then that spider got so outdated that I redid some of the logic and tests, here is new PR. 
## Checklist

- [x] Tests are implemented
- [x] All tests are passing


city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+
+class ChiSouthwestHomeEquityISpider(CityScrapersSpider):
+    name = "chi_southwest_home_equity_i"
+    agency = "Chicago Southwest Home Equity Commission I"
+    timezone = "America/Chicago"
+    allowed_domains = ["swhomeequity.com"]
+    start_urls = ["https://swhomeequity.com/agenda-%26-minutes"]
+    location = {
+        "name": "Southwest Home Equity Assurance office",
+        "address": "5334 W. 65th Street in Chicago, Illinois",
```
pjsier, line 28: If this is static we can add the ZIP code and standardize the formatting a bit so that it's "5334 W 65th St Chicago, IL 60638"

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    def parse(self, response):
+        table = response.xpath(
+            "body/div/div/div/div[2]/div[3]/div/div/section/div/div[2]"
+        )
+
+        for agenda_node in table.xpath(
```
pjsier, line 36: I think this approach makes sense, but their formatting doesn't seem very consistent, so we might have better luck using `defaultdict` to create a mapping of dates to links and then iterating over that to make sure we get everything. Here's an example in `chi_lsc_advisory` https://github.com/City-Bureau/city-scrapers/blob/a144119dd6c85a34ef7f18ffc01aa910274f2cd6/city_scrapers/spiders/chi_lsc_advisory.py#L37-L47

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+            "div[@data-ux=\"GridCell\"][contains(., 'Agenda')]"
+        ):
+            minutes_node = agenda_node.xpath(
+                "following-sibling::div[@data-ux=\"GridCell\"][contains(., 'Minutes')]"
+            )
+
+            if self._get_link(minutes_node):
+                minutes_contents = self.parse_pdf(self._get_link(minutes_node))
```
pjsier, line 44: Parsing the PDF adds a good amount of complexity here, including creating a new request outside of scrapy which ideally we'd like to avoid so that we can take advantage of scrapy's features like rate-limiting. I think it's fine if we just return "Governing Commission" as the title and don't parse the PDF content

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+            else:
+                minutes_contents = None
+
+            meeting = Meeting(
+                title=(
```
pjsier, line 49: Mentioned above, but this can return "Governing Commission" for all cases

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+                start=self._parse_start(agenda_node),
+                end=None,
+                all_day=False,
+                time_notes=(
+                    self._parse_time_notes(minutes_contents)
+                    if minutes_contents
+                    else None
+                ),
+                location=self.location,
+                links=self._parse_links([agenda_node, minutes_node]),
```
pjsier, line 65: Using the `defaultdict` strategy would replace this

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+        response = requests.get(url)
+        pdf_text = extract_text(BytesIO(response.content))
+        return pdf_text
+
+    def _parse_start(self, node):
+        # Try to get date from title
+        try:
+            date_str = self.find_date_in_title(node)
+            date_object = datetime.strptime(date_str, "%B %d %Y")
+        except (RuntimeError, ValueError):
```
pjsier, line 91: Is there a reason for including `RuntimeError` here? `ValueError` should cover all of our exceptions

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+                date_object = datetime.strptime(date_str, "%B %d %Y")
+            except (RuntimeError, ValueError):
+                return None
+
+        try:
+            start_time = self.find_time(node)
+            return date_object + start_time
+
+        except (RuntimeError, ValueError):
+            return date_object + timedelta(hours=18, minutes=30)
```
pjsier, line 103: A potentially simpler way of doing this than calling `timedelta` is calling `datetime.combine` with a `date` and `time` object as the arguments. That way we're more explicit about what's happening, since we're creating a datetime rather than finding the difference

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+        links = []
+        for node in nodes:
+            if node:
+                links.append(
+                    {"title": self._get_name(node), "href": self._get_link(node)}
+                )
+        return links
+
+    def _get_name(self, node):
+        name = node.xpath("string(.)").get()
```
pjsier, line 115: Since we know that the overwhelming majority of docs will be either Agenda or Minutes, we can check if either of those are present in the string and return "Agenda" or "Minutes" statically if so to keep things clean.

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+        name = name.replace("(pdf)Download", "")
+        return name
+
+    def _get_link(self, node):
+        url = node.xpath("a/@href").get()
+        if url:
+            return "https:" + url
```
pjsier, line 122: Just in case the scheme is ever added we should check if "http" is present before appending

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+            return date_object + start_time
+
+        except (RuntimeError, ValueError):
+            return date_object + timedelta(hours=18, minutes=30)
+
+    def _parse_links(self, nodes):
+        links = []
+        for node in nodes:
+            if node:
+                links.append(
```
pjsier, line 109: It looks like `_get_link` can return `None`, so we'll need to check that before adding to the list because the `href` property should never be `None`

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+        content = content.replace("\n", "")
+        time_str = self._use_time_regex(content)
+        dt = datetime.strptime(time_str, "%I:%M %p")
+        return timedelta(hours=dt.hour, minutes=dt.minute)
+
+    def _use_date_regex(self, text):
+
+        month_ind = re.search(MONTH_REGEX, text, flags=re.I)
+        if month_ind is None:
+            raise RuntimeError("No month found")
```
pjsier, line 150: Looks like these are where the `RuntimeError` from above comes. It's generally easier to read if we return `None` instead of an exception, so that would be preferable here if we can also check it in the function using this output

city_scrapers/spiders/chi_southwest_home_equity_i.py
```diff
@@ -0,0 +1,173 @@
...
+                    self._parse_title(minutes_contents)
+                    if minutes_contents
+                    else "Board Meeting"
+                ),
+                description="",
+                classification=COMMISSION,
+                start=self._parse_start(agenda_node),
+                end=None,
+                all_day=False,
+                time_notes=(
```
pjsier, line 59: Because we're assuming the time if we aren't parsing the minutes, we can add something statically like "See links for details"

pjsier (changes requested):
Thanks for sticking with this one! Made some comments, let me know if any of them aren't clear

pjsier:
@egfrank thanks for continuing to work on this! Let me know when you want me to take another look. Looks like `isort` is failing right now

Issue #718, opened 2019-04-15
Spider: Chicago State University


pjsier:
URL: https://www.csu.edu/boardoftrustees/dates.htm
Spider Name: `chi_state_university`
Agency Name: Chicago State University

Meeting title can always be "Board of Trustees" (unless there are committees). We also want to scrape the agendas, minutes, and transcripts here if possible

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

jartsavor:
I am going to give this a shot.

Also, I went through the [development](https://cityscrapers.org/docs/development/) docs, and I had some trouble at the automated test sections (for both pipenv and virtualenv). I kept getting an `UnicodeDecodeError` when running `pytest`. It turns out that I needed to add an `encoding='utf8'` argument to the `return open(file_path, 'r').read()` of line 37 in [tests/utils.py](https://github.com/City-Bureau/city-scrapers/blob/709d7c9536f030e5e2a5e0c6829d46de743bad86/tests/utils.py). Is this something that I should bring up as an issue or could this just be unique on my end? I have a windows 10 machine.

pjsier:
@jartsavor sounds great! And feel free to just include that in this PR, we've addressed it elsewhere but must have missed that section since we've deprecated that module in favor of `city_scrapers_core.utils`

jartsavor:
@pjsier okay will do, and thank for the quick reply.

jartsavor:
I have an issue related to an unknown date/time for a meeting. What should I do when there is a rescheduled time indicated on the website? For instance below,

Friday, March 1, 2019 
-RESCHEDULED to: TBD

is shown on the website, and I'm not sure  how to handle it.
I have checked [this](https://github.com/City-Bureau/city-scrapers/issues/88) issue, but it doesn't address this specific conflict.

pjsier:
@jartsavor any time something is rescheduled we're just considering it cancelled, so you could pass the text for the meeting (including where "RESCHEDULED" would be) to the `get_status` method in the `text` kwarg like this https://github.com/City-Bureau/city-scrapers/blob/bc1d5ff6163517df4e198dcd6161754a0a326d7c/city_scrapers/spiders/chi_schools.py#L42

bwalterscoding:
> @jartsavor sounds great! And feel free to just include that in this PR, we've addressed it elsewhere but must have missed that section since we've deprecated that module in favor of `city_scrapers_core.utils`

[...]

PR #974, opened 2020-09-24
0718 spider chicago state university

Rhysoshea:
## Summary

**Issue:** #718 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

### DIFF
--- a/city_scrapers/spiders/chi_state_university.py
+++ b/city_scrapers/spiders/chi_state_university.py
@@ -1,7 +1,9 @@
 import re
-from calendar import month_name
-from datetime import datetime
+from calendar import month_name, day_name
+from datetime import datetime, date
+from collections import defaultdict
 
+import scrapy
 from city_scrapers_core.constants import BOARD, COMMITTEE
 from city_scrapers_core.items import Meeting
 from city_scrapers_core.spiders import CityScrapersSpider
@@ -11,7 +13,8 @@ class ChiStateUniversitySpider(CityScrapersSpider):
     name = "chi_state_university"
     agency = "Chicago State University"
     timezone = "America/Chicago"
-    start_urls = ["https://www.csu.edu/boardoftrustees/dates.htm"]
+    start_urls = [f"https://www.csu.edu/boardoftrustees/\
+        meetingagendas/year{date.today().year}.htm"]
 
     def parse(self, response):
         """
@@ -20,14 +23,50 @@ class ChiStateUniversitySpider(CityScrapersSpider):
         Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
         needs.
         """
+        self.minutes_map = self._parse_minutes(response)
+        yield scrapy.Request(
+            "https://www.csu.edu/boardoftrustees/dates.htm",
+            callback=self._parse_meetings,
+            dont_filter=True
+        )
 
+    def _parse_minutes(self, response):
+
+        minutes_map = defaultdict(list)
+
+        for data in response.xpath("/html/body/div/div[4]/div/div[1]/div"):
+            for p in data.xpath(".//p"):
+                title = p.xpath(".//text()").extract()[0]
+                if "Meetings" in title:
+                    meetings = p.xpath("following-sibling::*")
+                    for meeting in meetings.xpath(".//li"):
+                        link = meeting.xpath(".//a").attrib["href"]
+                        dateStr = meeting.xpath(".//text()").extract()[0]
+                        dateStr = re.sub(",", " ", dateStr)
+                        dateStr = re.sub(r"\s+", " ", dateStr)
+
+                        commonStrings = \
+                            ["Standing Committee", "Special Board", "Full Board"]
+
+                        for i in commonStrings:
+                            if i in title:
+                                title = i
+
+                        minutes_map[title + " " + dateStr].append(
+                            {
+                                "title": title + " " + dateStr,
+                                "href": "https://www.csu.edu" + link,
+                            }
+                        )
+            return minutes_map
+
+    def _parse_meetings(self, response):
         for data in response.xpath("/html/body/div/div[4]/div/div[1]/div/div"):
-            titles = data.xpath(".//label/text()").extract()
-
-            for i, tab in enumerate(data.xpath('.//div[@class = "itd-tab"]')):
-
-                for item in tab.xpath(".//li"):
-
+            for label in data.xpath(".//label"):
+                title = label.xpath(".//text()").extract()[0]
+                sibling = label.xpath("following-sibling::*")
+                self._validate_location(sibling)
+                for item in sibling.xpath(".//li"):
                     months = {m.lower() for m in month_name[1:]}
                     text = item.xpath(".//text()").extract()[0]
                     monthMatch = next(
@@ -35,17 +74,18 @@ class ChiStateUniversitySpider(CityScrapersSpider):
                     )
                     if monthMatch is None:
                         continue
+                    start = self._parse_start(item)
 
                     meeting = Meeting(
-                        title=titles[i],
+                        title=title,
                         description=self._parse_description(item),
-                        classification=self._parse_classification(i),
-                        start=self._parse_start(item),
+                        classification=self._parse_classification(title),
+                        start=start,
                         end=None,
                         all_day=False,
                         time_notes=self._parse_time_notes(item),
                         location=self._parse_location(item),
-                        links=self._parse_links(item),
+                        links=self._parse_links(item, title, start),
                         source=self._parse_source(response),
                     )
                     meeting["status"] = self._get_status(meeting)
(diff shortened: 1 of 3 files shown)


Rhysoshea:
@pjsier I'm not sure why the imports are failing on the pull request tests, could you help me with this please?

alderocas:
Hi @Rhysoshea! do hope you and @pjsier don't mind me butting in (sorry!).

It looks like your imports are incorrectly formatted. Check the output of `pipenv run isort . --diff` to see which.

PS don't forget to `pipenv run` `flake8` and `black` to help with linting and formatting!

pjsier:
@alderocas I don't mind at all, thanks!

Rhysoshea:
Thanks a lot for your advice @alderocas, those were some important steps I missed out, and definitely useful tools to be aware of. All checks have passed now, so any changes you need making please let me know @pjsier

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+                    meeting["id"] = self._get_id(meeting)
+
+                    yield meeting
+
+    def _parse_description(self, item):
+        parts = item.xpath(".//text()").extract()
+        description = "".join([str(x) for x in parts])
+        description = (
+            description.replace("\xa0", " ")
+            .replace("\n", " ")
```
pjsier, line 61: One regex solution we've used for this kind of whitespace replacement is to replace all whitespace sequences with a single space like this:

```python
description = re.sub(r"\s+", " ", description)
```

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+            .replace("                                     ", " ")
+            .replace("\t", " ")
+            .strip()
+        )
+        return description
+
+    def _parse_classification(self, i):
```
pjsier, line 68: It might be a little more transparent if we check the title instead of the tab index in case they change that

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+        text = text.replace("\xa0", "").strip()
+
+        notes = ""
+        if "reschedule" in text.lower() or "postpone" in text.lower():
+            notes = text
+        if "tbd" in text.lower():
+            notes = "Meeting time TBD"
+        return notes
+
+    def _parse_location(self, item):
```
pjsier, line 130: It's fine to return a static location, but we'll want to add a `_validate_location` method to make sure that it hasn't changed. You can find an example here https://github.com/City-Bureau/city-scrapers/blob/c6771d58209fa9ade4aa3768fa3ddb23b45c09fc/city_scrapers/spiders/chi_ssa_34.py#L61-L63

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+            "name": "Room 15, 4th Floor, Gwendolyn Brooks Library Auditorium",
+        }
+
+    def _parse_links(self, item):
+
+        try:
+            link = item.xpath(".//a").attrib["href"]
+        except KeyError:
+            link = ""
+        title = ""
```
pjsier, line 143: We'll want the title to default to the link text here if it doesn't match any of the options

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+        return notes
+
+    def _parse_location(self, item):
+
+        return {
+            "address": "9501 S. King Drive Chicago, IL 60628",
+            "name": "Room 15, 4th Floor, Gwendolyn Brooks Library Auditorium",
+        }
+
+    def _parse_links(self, item):
```
pjsier, line 137: It looks like they're posting agendas and minutes as well, and we'll want to make sure to parse those https://www.csu.edu/boardoftrustees/meetingagendas/year2020.htm

You can find an example of parsing the documents first and then parsing meetings in [`chi_north_river_mental_health`](https://github.com/City-Bureau/city-scrapers/blob/5c7c9fef9e87ae71bb9c4ba7a6156b6ba1e84883/city_scrapers/spiders/chi_north_river_mental_health.py)

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,151 @@
...
+            .strip()
+        )
+        return description
+
+    def _parse_classification(self, i):
+        classifications = [BOARD, BOARD, COMMITTEE]
+        return classifications[i]
+
+    def _parse_start(self, item):
+        text = item.xpath(".//text()").extract()[0]
```
pjsier, line 73: Is there a reason for creating the datetime from each component piece rather than calling `datetime.strptime` with a format (or maybe formats)? This seems like it could return false positives if the pattern changes rather than failing directly, and it looks like the dates are generally pretty standard

pjsier (changes requested):
Thanks for the PR! Let me know if I can clarify any of my comments

city_scrapers/spiders/chi_state_university.py
Rhysoshea: In the example you have given, they appear to be parsing minutes separately to normal meetings. 
In my case, would it be reasonable to stick with what I have, then check the minutes url to see if there is an associated link to the minutes and include it?

city_scrapers/spiders/chi_state_university.py
Rhysoshea: Yes I wasn't very happy with this approach but it was what I settled on after battling with the formatting for a long time. I used `"%A, %B %d, %Y @ %H:%M %p"` and `"%B %d, %Y @ %H:%M %p"` for instances without the day at the beginning. But for some reason this wasn't working on any of the date and time strings. I tried slight variations as well as parsing some segments of the date and trying `datetime.strptime` on just those parts - this includes stripping whitespace and odd characters and so on but it wouldn't play ball. I've not come across such awkward behavior before so if you have any suggestions that would be useful.
I went with this approach also because there were instances of dates without time or minutes and so on, so as a catch all I tried to account for any kind of supplied date as much as I could.

city_scrapers/spiders/chi_state_university.py
Rhysoshea: To confirm, this would be the title of all meeting types to "Meeting Dates"?
Is it still ok to base the meeting classification on the tabs?

pjsier: Sorry, I think I might have misunderstood. I'm only talking about the link titles here, but for the comment above it would be great if we could pull the text from the tab rather than base it on the index

city_scrapers/spiders/chi_state_university.py
pjsier: Sorry about that, this is a better example of what I had in mind

https://github.com/City-Bureau/city-scrapers-akr/blob/c451c5b3b90b3f7b0e44201978b9c3e27b44947c/city_scrapers/spiders/summ_library.py#L16-L22

We could first check the minutes page, build an index of dates to documents, and then parse the schedule page attaching documents wherever they match

city_scrapers/spiders/chi_state_university.py
pjsier: Hmm thanks for clarifying! We've generally been removing the day string at the beginning because it doesn't add necessary info (and sometimes they're wrong), so that could be one place to start. It can also help to normalize the whitespace with something like:

```python
date_str = re.sub(r"\s+", " ", date_str)
```

That way you don't have to hunt down any specific hidden whitespace characters. And one other thing that has helped is removing commas and periods since they're typically unneeded and in many cases inconsistent. I'm not sure if `strptime` accepts "a.m." vs "am" either. If it's alright with you it would be great to give that another shot, but let me know if you run into specific edge cases!

city_scrapers/spiders/chi_state_university.py
Rhysoshea: Sorry my mistake, I was meant to comment on your above comment about title. I see what you mean. I tried to do this but found it difficult because those titles are written inside `<label>` objects that sit outside of the tab divs themselves.

For example
```
<input id="tabthree" type="radio" name="itd-tabs"></input> <label class="itd-label" for="tabthree">Standing Committee Meeting</label>
<div class="itd-tab">
```

This is why the titles were collected separately. What do you think?

pjsier: Gotcha, thanks for clarifying! It looks like with the way they have it set up the tab div has to be the immediate sibling of the label. You should be able to get this working by iterating over the labels instead, and inside the loop you can parse the title from the label and then access the tab with the `following-sibling` selector in XPath

Rhysoshea:
@pjsier thanks a lot for your helpful comments, I've made the changes you've requested.

I'm unsure if the way I've included the minutes in `links` is suitable so please let me know if changes need to be made.
The trouble I was having with `_parse_start` ended up being with a.m. and p.m., so removing the `.` from each worked and hopefully all cases are covered now. There's still some possibly odd looking code in there with trying to remove unnecessary words - I did it this way because there are some instances of words being present in the middle of date strings i.e. `August, 24 2020 (Board Retreat) @ 9:00 a.m.`, so that is the reasoning. Again please let me know if you'd prefer an alternate approach.

city_scrapers/spiders/chi_state_university.py
```diff
@@ -0,0 +1,200 @@
...
+        description = (
+            description.replace("\xa0", " ")
+            .replace("\n", " ")
+            .replace("\t", " ")
+            .strip()
+        )
+        description = re.sub(r"\s+", " ", description)
+        return description
+
+    def _parse_classification(self, title):
```
pjsier, line 113: Classification should always return one of the options, so we can default to `BOARD` here unless it's a committee

pjsier (changes requested):
Thanks for making these changes! Made a quick comment, but in general it looks like it isn't returning any meetings? https://github.com/City-Bureau/city-scrapers/runs/1203026923#step:10:1

Rhysoshea:
@pjsier sorry I didn't realise that! It was because of this
```
    start_urls = [
        f"https://www.csu.edu/boardoftrustees/\
        meetingagendas/year{date.today().year}.htm"
    ]
```
The linter complains that the line is too long so I included the `\` to break up the line but I've just found someone's approach to include `# noqa` on the required line. Does this stop the linter from raising an error?

pjsier:
@Rhysoshea no problem! Yeah you can add `# noqa` in cases where it makes sense to ignore the linting errors, and we typically use it for long URLs

PR #975, opened 2020-09-29
797 spider governors state university

jmelot:
## Summary

**Issue:** 797

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions


city_scrapers/spiders/il_governors_state_university.py
```diff
@@ -0,0 +1,201 @@
...
+                row_text = " ".join(row.css("* ::text").getall())
+                meeting["status"] = self._get_status(meeting, text=row_text)
+                meeting["id"] = self._get_id(meeting)
+
+                yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title. The inner html of the first column varies
+        quite a bit - brs, divs, b tags - so figuring out what is the title based on
+        line position. Sometimes the "title" is only a date, so if all else fails,
```
pjsier, line 55: Thanks for this comment! It really helps clarify things, especially since there are so many edge cases here. It seems like when the title is only a date it's a full board meeting, so we could return "Board of Trustees" in those cases.

And this is a small thing that we haven't documented well, but we're trying to remove "Meeting" from the ends of titles if it isn't needed so we can drop it from any title except for "Special Meeting"

city_scrapers/spiders/il_governors_state_university.py
```diff
@@ -0,0 +1,201 @@
...
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False. Doesn't seem to occur
+        for this website, with the possible exception of the retreats, which aren't
+        quite all day"""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
```
pjsier, line 154: It looks like this is picking up "Postponed to" as the location name, so it might be easiest to return the default if "Postponed" is present. It shows up in [this line in the output](https://github.com/City-Bureau/city-scrapers/runs/1180091014#step:10:372). Let me know if that link doesn't work, I haven't done many direct links to GitHub Actions output

I'm also seeing one where "Room D34000" shows up as the address [in this output line](https://github.com/City-Bureau/city-scrapers/runs/1180091014#step:10:832) so we might be able to use the default address if "Room" is in the address here.

city_scrapers/spiders/il_governors_state_university.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = []
+        # the links to the agenda, if present, are in the third and fourth columns
+        for col in [2, 3]:
+            for link_parent in item[col].xpath("a"):
+                link_ext = link_parent.css("::attr(href)").get()
+                if link_ext is not None:
+                    link = self.base_url + link_ext
```
pjsier, line 194: Small thing here, but instead of passing the `base_url` statically we can use the `response.urljoin` method to create the full URL path if `response` is passed to `_parse_links`

pjsier (changes requested):
Thanks for the PR, this looks great! I made a few small comments on edge cases, but once those changes are made this should be good to merge

jmelot:
Thanks for all the helpful comments! I've tried to address them, but please let me know if there's anything else you spot that needs changing.

pjsier (approved):

Issue #574, opened 2018-11-07
Spider: Chicago Special Service Area #52 51st Street


pjsier:
URL: https://www.51ststreetchicago.com/about.html
Spider Name: `chi_ssa_52`
Agency Name: Chicago Special Service Area #52 51st Street

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

lilianhj:
I'm working on this!

jusadi:
It looks like the old URL is 404-ed. Looks like a possible replacement is: [https://www.51ststreetchicago.com/about.html](https://www.51ststreetchicago.com/about.html).

pjsier:
@jusadi that looks right! Are you interested in working on this one?

jusadi:
Yeah! I'd work on it.

jusadi:
The page does not seem updated for 2020.

pjsier:
It looks like they didn't update the header but the dates are updated now @jusadi

sosolidkk:
If no one is working on this, i'd like to take it.

pjsier:
@sosolidkk all yours! Assigning you now

sosolidkk:
I'm sorry for the delay. I've run into some heavy workload at job, but I'll submit a PR soon. Just posting this comment to say that I'm still working on it

pjsier:
Closed by #983, thanks @sosolidkk!

PR #983, opened 2020-10-01
0574 spider chi ssa 52

sosolidkk:
## Summary

**Issue:** #574 

## Checklist

- [X] Tests are implemented
- [X] All tests are passing
- [ ] Code comments from template removed (Not quite sure about this)

## Questions

So, i've run into two problems. The first one was when parsing the date, the months were written incorrectly. My solution was to have a list of months and check the ratio comparing the right month string versus the wrong month string, and if the ratio is above 0.5, the incorrect string is replace with the correct one in the list. And about the test, i think i've written few tests and i would some guidance to improve them.

### DIFF
--- a/city_scrapers/spiders/chi_ssa_52.py
+++ b/city_scrapers/spiders/chi_ssa_52.py
@@ -102,16 +102,14 @@ class ChiSsa52Spider(CityScrapersSpider):
 
         try:
             date = datetime.strptime(
-                f"{item[2]} {item[1]} {item[3]} {time}",
-                "%d %B %Y %I:%M %p",
+                f"{item[2]} {item[1]} {item[3]} {time}", "%d %B %Y %I:%M %p",
             )
         except ValueError:
             for month in months:
                 ratio = SequenceMatcher(None, month, item[1]).ratio()
                 if ratio > 0.5:
                     date = datetime.strptime(
-                        f"{item[2]} {month} {item[3]} {time}",
-                        "%d %B %Y %I:%M %p",
+                        f"{item[2]} {month} {item[3]} {time}", "%d %B %Y %I:%M %p",
                     )
         return date
 
--- a/city_scrapers/spiders/chi_ssa_52.py
+++ b/city_scrapers/spiders/chi_ssa_52.py
@@ -68,7 +68,13 @@ class ChiSsa52Spider(CityScrapersSpider):
             month_day = meet[2]
             year = meet[3][:4]
             time = meet[3][4:]
-            payload = {"day_week": day_week, "month": month, "month_day": month_day, "year": year, "time": time}
+            payload = {
+                "day_week": day_week,
+                "month": month,
+                "month_day": month_day,
+                "year": year,
+                "time": time,
+            }
 
         if len(meet) == 6:
             day_week = meet[0]
@@ -119,12 +125,16 @@ class ChiSsa52Spider(CityScrapersSpider):
         ]
 
         try:
-            date = datetime.strptime(f"{item['month_day']} {item['month']}, {item['year']}", "%d %B, %Y")
+            date = datetime.strptime(
+                f"{item['month_day']} {item['month']}, {item['year']}", "%d %B, %Y"
+            )
         except ValueError:
             for month in months:
                 ratio = SequenceMatcher(None, month, item["month"]).ratio()
                 if ratio > 0.5:
-                    date = datetime.strptime(f"{item['month_day']} {month}, {item['year']}", "%d %B, %Y")
+                    date = datetime.strptime(
+                        f"{item['month_day']} {month}, {item['year']}", "%d %B, %Y"
+                    )
 
         time = item["time"].split(":")  # [0] = hours, [1] = minutes
         date = date.replace(hour=int(time[0]), minute=int(time[1]))
--- a/city_scrapers/spiders/chi_ssa_52.py
+++ b/city_scrapers/spiders/chi_ssa_52.py
@@ -1,0 +1,84 @@
+from city_scrapers_core.constants import NOT_CLASSIFIED
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa52Spider(CityScrapersSpider):
+    name = "chi_ssa_52"
+    agency = "Chicago Special Service Area #52 51st Street"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.51ststreetchicago.com/about.html"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        items = response.css("div.paragraph")[3:4]
+        titles = items.css("strong::text").getall()
+        meetings = items.css("ul")
+        items = list(zip(titles, meetings))
+
+        for item in enumerate(items):
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description=self._parse_description(item),
+                classification=self._parse_classification(item),
+                start=self._parse_start(item),
+                end=self._parse_end(item),
+                all_day=self._parse_all_day(item),
+                time_notes=self._parse_time_notes(item),
+                location=self._parse_location(item),
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return ""
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        return None
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "",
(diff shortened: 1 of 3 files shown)


city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+        item = item.replace("\xa0", "").strip()
+        return item
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
```
pjsier, line 108: We can use `COMMISSION` for all of these meeting since it seems like there aren't committees

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+        ]
+
+        try:
+            date = datetime.strptime(
+                f"{item['month_day']} {item['month']}, {item['year']}", "%d %B, %Y"
+            )
+        except ValueError:
+            for month in months:
+                ratio = SequenceMatcher(None, month, item["month"]).ratio()
+                if ratio > 0.5:
```
pjsier, line 134: This is an interesting approach! We run into this a lot, so I appreciate this as a standard library tactic we haven't considered before. Is there a reason behind 0.5 as the threshold? Haven't tested, but my only concern would be if they start using abbreviations later on

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "",
```
pjsier, line 158: It's a bit hidden but we'll want to use "220 E 51st St Chicago, IL 60615" for the address and we can leave the name blank

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+            "name": "",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        return [{"href": "", "title": ""}]
```
pjsier, line 164: More info [in the docs](https://cityscrapers.org/docs/development/#links), but we'll want this to return an empty list rather than a list with one dict with empty values if we don't have any links to parse

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+        except ValueError:
+            for month in months:
+                ratio = SequenceMatcher(None, month, item["month"]).ratio()
+                if ratio > 0.5:
+                    date = datetime.strptime(
+                        f"{item['month_day']} {month}, {item['year']}", "%d %B, %Y"
+                    )
+
+        time = item["time"].split(":")  # [0] = hours, [1] = minutes
+        date = date.replace(hour=int(time[0]), minute=int(time[1]))
```
pjsier, line 140: It looks like 1:30 is in the afternoon here, so we wouldn't want to use the integers directly. We can use `datetime.strptime` similar to the way you are above, but with a check first on the hours in case they start doing morning meetings. We could do something like "if the hour int is above 8 and less than 12 it should be AM, otherwise PM"

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        items = response.css("div.paragraph")[3:4]
+        titles = items.css("strong::text").getall()
+        meetings = items.css("ul")
```
pjsier, line 24: We only want the first `ul` here since the second one isn't related to the Commission

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+                "year": year,
+                "time": time,
+                "info": info,
+            }
+
+        return payload
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        item = item.replace("\xa0", "").strip()
```
pjsier, line 99: We can probably return "Commission" here statically since we don't need the extra info and it doesn't look like they have committees

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,168 @@
...
+        meet = meet.replace(".", "")
+        meet = meet.replace("\xa0", "")
+        meet = meet.replace("…", "")
+        meet = meet.replace(",", "")
+        meet = meet.strip()
+
+        return meet
+
+    def _clean_meet(self, meet):
+        """Clean and parse a possible meet <li> item"""
```
pjsier, line 60: Could you explain a bit more what the intent of this section is? It seems like if we remove the extra periods, commas, and whitespace characters we should be able to parse out the pattern with a regular expression like this we can simplify this section

```python
clean_str = re.sub(r"[\.,\s]+", " ", meet)
datetime_match = re.search(r"[A-Z][a-z]{2,9} \d{1,2} \d{4} \d{1,2}:\d{2}", clean_str)
```

pjsier (changes requested):
Thanks for the PR! Made a few comments here, but let me know if you have any questions!

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Thanks, i'll change it.

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: There isn't a concrete reason. Since all i did was testing the possible outcomes with permutations of the months strings. The bigger result i did get was 0.5 comparing `JUNE` with `JUEN`. So I inferred that if the ratio is bigger than 0.5, the word is probably the same. I could be wrong, but it was the only "clever" solution i've found to solve this in an almost automatic way

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: 👍🏻

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Thanks for the info. I'll change it

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Yeah, i was not quite sure if 1:30 was PM or AM, since in my country we use the 24 hour format.

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: So i don't have to parse the `GENERAL BUSINESS MEETING DATES 2019` part?

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: 👍🏻

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: My only intent is to remove the unused chars. When i parse the `li` element i get back something like this: `TUESDAY MARCH 24, 2020…………......12:00 QUARTERLY LUNCHEON`. Since i'm not that good with RE, I just added several replaces. Thanks for the suggestion anyway, i'll make the change

city_scrapers/spiders/chi_ssa_52.py
pjsier: That's a good enough reason! We might want to account for the shorter variations though like "Jun" and "Jul"

city_scrapers/spiders/chi_ssa_52.py
pjsier: Sounds good! The 24 hour format makes a lot more sense to be honest, but isn't very common here

city_scrapers/spiders/chi_ssa_52.py
pjsier: Yep, you can feel free to ignore it. These "Special Service Areas" are governmental bodies that are managed by private entities, and it looks like the "General Business Meeting" items are of the private entity

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Ok

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Ok, thanks for the tip

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Just a question, where did you saw this address? If it's not on the page itself, should i put it hard coded?

pjsier: I looked up some of their official documents, which isn't ideal, but better than nothing in this case

sosolidkk:
Hey @pjsier thanks for the tips. I've updated the code based on your suggestions!

city_scrapers/spiders/chi_ssa_52.py
```diff
@@ -0,0 +1,141 @@
...
+        minutes = time[1]
+
+        if int(hour) >= 8 and int(hour) <= 12:
+            return f"{hour}:{minutes} AM"
+        return f"{hour}:{minutes} PM"
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        # item = item.replace("\xa0", "").strip()
+        return COMMISSION
```
pjsier, line 73: It's a bit arbitrary, but even though the variable is technically the same thing there's the potential that it changes in the future and we know that we want the title here to be "Commission", so we can return the string directly. Also we can remove the comment above

pjsier (changes requested):
Looks great! Thanks for making those changes. One small tweak and then this is good to merge

city_scrapers/spiders/chi_ssa_52.py
sosolidkk: Ok, tonight i'll make this change. Thanks for the help

sosolidkk:
Done @pjsier ! Thanks for the help.

pjsier (approved):

Issue #854, opened 2019-10-03
Spider: Cook County Pharmaceutical Disposal Advisory Committee


pjsier:
URL: https://www.cookcountysheriff.org/rx/advisory-committee/
Spider Name: `cook_pharmaceutical_disposal`
Agency Name: Cook County Pharmaceutical Disposal Advisory Committee

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

yuriy-os:
Can I help with this one?

pjsier:
@yuriy-os go for it! Let me know if you have any questions

yuriy-os:
@pjsier I'm sorry, I can't find enough time to properly work on this issue, feel free to reassign it.

glowingrunes:
Hi, new here and would like to give this a shot for Hacktoberfest!

pjsier:
@glowingrunes sounds good!

KilimAnnejaro:
Hi, could I help out with this issue? Thanks!

pjsier:
@KilimAnnejaro sure! Assigning you now

KilimAnnejaro:
Just want to add a documentation note here: per discussion on Slack, we won't be using OCR to scrape the image at the top of the page. Instead we will be scraping the PDF links below that to get Meeting objects.

shf:
Hi, I would like to take this issue. Thanks.

pjsier:
@shf sounds good! Assigning you now

shf:
@pjsier They are using an image for 2020 meetings, is it Ok to use pytesseract to extract text from the image? users need to install tesseract-ocr for this to run
`sudo apt-get install tesseract-ocr`

pjsier:
@shf OCR is out of scope for what we're doing here, but I think it's still valuable to parse out past meetings in case that changes going forward

shf:
@pjsier the past meetings are also in PDF format, I found you have used pdfminer before in the code, should I continue with that?

pjsier:
I don't think it'll be necessary here since there isn't a lot of information in general, so we're mostly just trying to get dates and links

pjsier:
Closed by #984, thanks @shf!

PR #984, opened 2020-10-06
Add 0854-spide-cook-pharma

shf:
## Summary

**Issue:** #0854

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

The date format for past meetings has been inconsistent. I tried to find the dates without using the PDF files.

### DIFF
--- a/city_scrapers/spiders/cook_pharmaceutical_disposal.py
+++ b/city_scrapers/spiders/cook_pharmaceutical_disposal.py
@@ -22,7 +22,8 @@ class CookPharmaceuticalDisposalSpider(CityScrapersSpider):
 
         # Link to current meeting shown as a single image
         # In case you want to proceed with image OCR
-        # image_url = base_url + response.xpath("//div[@class='col-sm-12 ']/h1/img/@src").get()
+        # image_url = base_url +
+        #    response.xpath("//div[@class='col-sm-12 ']/h1/img/@src").get()
         # image_response = requests.get(image_url)
         # print(image_to_string(Image.open(BytesIO(image_response.content)),lang='eng'))
 
--- a/city_scrapers/spiders/cook_pharmaceutical_disposal.py
+++ b/city_scrapers/spiders/cook_pharmaceutical_disposal.py
@@ -14,21 +14,10 @@ class CookPharmaceuticalDisposalSpider(CityScrapersSpider):
     def parse(self, response):
         """
         `parse` should always `yield` Meeting items.
-
-        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
-        needs.
         """
-        self.base_url = "https://www.cookcountysheriff.org"
-
-        # Link to current meeting shown as a single image
-        # In case you want to proceed with image OCR
-        # image_url = base_url +
-        #    response.xpath("//div[@class='col-sm-12 ']/h1/img/@src").get()
-        # image_response = requests.get(image_url)
-        # print(image_to_string(Image.open(BytesIO(image_response.content)),lang='eng'))
-
         # Scraping past meetings
         for item in response.xpath("//div[@class='col-sm-12 ']/p"):
+            self._ignore = False
             meeting = Meeting(
                 title=self._parse_title(item),
                 description=self._parse_description(item),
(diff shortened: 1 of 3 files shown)


pjsier:
I can take a closer look at this once the style checks pass @shf, and you should be able to run `pipenv run black city_scrapers tests` to fix anything here

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        links = list()
+        for report_obj in item.xpath(".//a/@href"):
+            report_desc = report_obj.get().split("/")[-1]
+            links.append(
+                {
+                    "href": self.base_url + report_obj.get(),
```
pjsier, line 106: We can use `response.urljoin` if `response` is passed as an argument to the function as a shortcut for doing this

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+    start_urls = ["https://www.cookcountysheriff.org/rx/advisory-committee/"]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        self.base_url = "https://www.cookcountysheriff.org"
```
pjsier, line 21: This can be removed based on the comment in `_parse_links`

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Link to current meeting shown as a single image
```
pjsier, line 23: I think we won't plan on looking into OCR here, so we can remove this comment block, good to know though, thanks!

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Cook County Safe Disposal of Pharmaceuticals Advisory Committee"
```
pjsier, line 52: Small thing but we can leave the "Cook County" off here so that it's just "Safe Disposal of Pharmaceuticals Advisory Committee"

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+        return "See agenda to confirm times"
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "Conference Room of 704 Daley Center, Chicago, Illinois 60602",
```
pjsier, line 95: We'll want the address to be usable for geocoding here so we can return "50 W Washington St, Room 407, Chicago, IL 60602"  for address and "Daley Center" for name

city_scrapers/spiders/cook_pharmaceutical_disposal.py
```diff
@@ -0,0 +1,114 @@
...
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return ADVISORY_COMMITTEE
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        # Dates are not correct for some meetings as they
```
pjsier, line 64: To be safe here, we should ignore any meeting where we can't pull the month, date, and year

pjsier (changes requested):
Thanks for the PR! It's looking great so far, let me know if any of my comments aren't clear

city_scrapers/spiders/cook_pharmaceutical_disposal.py
shf: The only issue is that we only can get specific dates from one meeting using the agenda and minutes file names.

city_scrapers/spiders/cook_pharmaceutical_disposal.py
shf: response.urljoin uses "https://www.cookcountysheriff.org/rx/advisory-committee/" as base the base URL but they are uploading files on in this address "https://www.cookcountysheriff.org/wp-content/uploads/xyz.pdf"
How can I solve this?

pjsier: `response.urljoin` uses the same strategy as the browser for determining where a link goes, and because these start with "/wp-content" it uses the domain base. Just ran it locally from your branch and it seems to work!

city_scrapers/spiders/cook_pharmaceutical_disposal.py
pjsier: It looks like we should be able to get it from at least December and July of 2019, although they aren't super clean, which is fine with me if that works for you!

shf:
@pjsier would you add "hacktoberfest-accepted" label as well? Thanks.

pjsier:
@shf the repo has the hacktoberfest label on it already so this should count!

Issue #935, opened 2019-11-30
Looking for issues to work on? Start here!


pjsier:
The City Scrapers project is split into multiple repos for different locations. You can see a list of these projects by checking out the [city-scrapers](https://github.com/topics/city-scrapers) topic tag. Issues that are open for contributors are labeled "help wanted".

Check out our [contributing guidelines](https://github.com/City-Bureau/city-scrapers/blob/master/CONTRIBUTING.md) and [development documentation](https://cityscrapers.org/docs/development/) for more information on how to get started

idiosyncratic4:
Can I take this one

Rishabhy190:
I'd like to help

PR #987, opened 2020-10-18
Rename index.md to INDEX.md

adityaverma121:


pjsier:
Thanks for the PR! Our docs site isn't following the same conventions as some top-level markdown files would though, so we aren't interested in changing this filename right now. You're welcome to take a look through our open issues or see our [pinned issue with ways of getting started](https://github.com/City-Bureau/city-scrapers/issues/935)

PR #989, opened 2020-11-06
WIP 0880 spider chi ssa 20

guytet:
## Summary

**Issue:** #ISSUE_NUMBER

## Checklist

- [ ] Tests are implemented
- [ ] All tests are passing

## Questions
An issue I'm encountering while trying to recurs over `<h3>2019 SSA Meetings</h3>` , as this tag and its siblings are where meeting information is presented. 
I was able to move around the tree using: `h2 = response.xpath("//h2[contains(text(), 'SPECIAL SERVICE AREAS')]/following-sibling::p/text()")` - but, I understood `contains()` may not be the best route in this case (white space changes).

In current state, I'm able to parse all `h3` in the page, get the one I'm interested in `2019 SSA Meetings`, but from that point on - I cannot use `xpath` in order to traverse the tree, as the `h3` is an iterated item with no xpath siblings. 

I've tried to use relative xpath paths such as `(./)` within the iterated item, but it seems clear that when the regex matches - I'm holding a single xpath element, and possibly nothing more. 

I'd be happy to receive some guidelines as to how I can keep on solving this issue. 
Thank you.


.gitignore
```diff
@@ -133,3 +133,7 @@ src/
 logs/*.log
 
 # output files: local gitignore added to city_scrapers/local_outputs/
+
```
pjsier, line 136: Usually we'll want to keep changes like this separate from a scraper PR. You can add it to `.git/info/exclude` to exclude those files in your local repo though without changing the .gitignore

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+        """
+
+
+        for h3 in response.xpath("//h3"):
+           entry_str = h3.xpath("./text()").extract_first()
+           if entry_str and "ssa meetings" in entry_str.lower():
```
pjsier, line 21: This is looking right to me so far! You could also remove extra whitespace with something like

```python
entry_str = re.sub(r"\s+", " ", entry_str)
```

pjsier (changes requested):
Made some quick comments, let me know if there's anything in particular you want me to look at!

city_scrapers/spiders/chi_ssa_20.py
guytet: Thank you @pjsier , I can indeed work with that - but the resulting issue is that I can no longer use something like `following-siblinng` in order to traverse other items such as `<p`> that are related to my `<h3>`. Would happen to have an idea as to how to do that ?

.gitignore
guytet: >Usually we'll want to keep changes like this separate from a scraper PR. 
Sure, that's something how I should have cleaned up:)

city_scrapers/spiders/chi_ssa_20.py
guytet: I'm currently trying to create a dict (or a tuple) from the resulting `h3` and see if I can isolate the meeting dates related to `ssa 20` that way. 
Then, pass those strings for date manipulation onto `parse_start`_.

guytet:
Changed the logic and pushed. 
for white spaces and lower case, added 
```
base = [ re.sub(r"\s+", " ", item).lower() for item in base ]
```
When the spider is run, the result is now:

```
2019 ssa meetings
ssa 20:

wednesday, june 5, 9 a.m.
beverly bank & trust, 10258 s. western ave.
wednesday, july 10, 9 a.m.
beverly bank & trust, 10258 s. western ave.
```
I hope that from this point onward, I can move to processing the start times using the  `parse_start` method. 
Does this look decent enough in order to move onto `parse_start` :) ?

pjsier:
@guytet sorry for the delay, I think that's a good next step!

guytet:
> @guytet sorry for the delay, I think that's a good next step!

No problem @pjsier :) . Thank you for the feedback.

guytet:
At current state, I believe(hope),  the naive datetime object is being passed correctly `by _parse_start()`, if this seems satisfactory I can move forward, if not - please note what should I improve/change and i'll be happy to keep working on it.

guytet:
> At current state, I believe(hope), the naive datetime object is being passed correctly by` _parse_start()`, if this seems satisfactory I can move forward, if not - please note what should I improve/change and i'll be happy to keep working on it.

Hmm.. something is broken in the last commit, please ignore the above, I'm back to working on it :)

guytet:
Updates for the last commit:
- Took care of liniting requirements per `pipenv run flake8`
- The spider's run does seems to yield the expected result (from my inexperienced POV, obviously needs review)
- `pytest` exists without errors

I've looked in other spiders under `city_scrapers/spiders. Please review and let me know which items require more work.
Thank you.

pjsier:
@guytet Thanks for the updates! It looks like the test file isn't in version control though?

guytet:
> @guytet Thanks for the updates! It looks like the test file isn't in version control though?

@pjsier You're right. I will look into that. (Sorry for delayed response, for some reason I didn't see an email from github about your comment).

guytet:
Anther comment I'd like to make - I'm working on another version of this spider, however it could take more time. The main challenge  I see is - this page handles two SSA's (20 and 64), and so  - not all meetings and dates are data we'd like to process or deliver. 

I'm sure the next item is not unique, none the less it is challenging: The meeting info is posted using separate HTML tags which may not keep the same logic when newer meetings are added. (Trying to look into the future being only 2019 meetings are posted for now), and somehow make sure when future year meetings are posted, the spider is still valid. )

guytet:
Added the test file. Please let me know if anything needs improving.

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+from city_scrapers_core.spiders import CityScrapersSpider
+
+
+class ChiSsa20Spider(CityScrapersSpider):
+    name = "chi_ssa_20"
+    agency = "Chicago Special Service Area #20 South Western Avenue"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.mpbhba.org/business-resources/"]
+    location = {
+        "name": "Beverly Bank & Trust,",
```
pjsier, line 15: We can drop the comma here

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+        "address": "10258 s. Western ave.",
```
pjsier, line 16: This can be cleaned up as "10258 S Western Ave, Chicago, IL 60643

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def parse(self, response):
+
+        base = response.xpath(
+            "//*[self::p or self::strong or self::h3]/text()"
+        ).getall()
+
+        base = [re.sub(r"\s+", " ", item).lower() for item in base]
+
+        for index, line in enumerate(base):
```
pjsier, line 27: Is there a reason for iterating through this separately? A more standard way of doing this would be what you're doing below where we check a conditional and `continue` if it isn't met.

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "SSA 20"
```
pjsier, line 69: For SSAs we've been defaulting to "Commission" to minimize duplicated information

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        return ""
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return NOT_CLASSIFIED
```
pjsier, line 77: This can be the `COMMISSION` constant

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _parse_start(self, item, year):
+
+        if not any(word in item for word in ["beverly", "ssa"]):
+            item = re.sub(r"([,\.])", "", item).strip()
+            ready_date = item + " " + year
+            date_object = datetime.strptime(ready_date, "%A %B %d %I %p %Y")
```
pjsier, line 84: We've run into fewer errors ignoring the day of the week `%A` because it doesn't add info and can sometimes have typos.

city_scrapers/spiders/chi_ssa_20.py
```diff
@@ -0,0 +1,105 @@
...
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        return [{"href": "", "title": ""}]
```
pjsier, line 101: We should try and parse these from the section on the right and attempt to match them to meetings. You can see an example of this in [`chi_il_medical_district`](https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/city_scrapers/spiders/chi_il_medical_district.py)

pjsier (changes requested):
Thanks for your work on this! Let me know if any of the comments aren't clear

guytet:
> Thanks for your work on this! Let me know if any of the comments aren't clear

Thank you @pjsier , I will review your helpful input and adjust accordingly, will of course ask if anything's unclear.

city_scrapers/spiders/chi_ssa_20.py
guytet: Fixed.

city_scrapers/spiders/chi_ssa_20.py
guytet: Fixed.

city_scrapers/spiders/chi_ssa_20.py
guytet: Fixed.

city_scrapers/spiders/chi_ssa_20.py
guytet: Fixed.

city_scrapers/spiders/chi_ssa_20.py
guytet: There was no special reason, just stuck with what worked initially :neutral_face: .
I moved this down to line 40 and it's handled within the main loop `for item in base:`

city_scrapers/spiders/chi_ssa_20.py
guytet: I will try to accomplish this. I'm afraid I may have chosen a less than optimal logic if including the links should be involved,
(We had a small conversation about this on Slack, irrc). 

I will try and see how this option can be factored in.

guytet: Another item that's worth mentioning: The links on the right do not have dates posted on them but rather "1st quarter 2018" , "1st quarter 2019" and so on. 
The meetings posted for 2019 ("current" at the time, I assume) were in June and July - not necessarily corresponding to a quarterly schedule, that could be a challenge, too.

guytet:
Well, this just in: SSA 20/64 has meetings scheduled for 2021 ! (and things have changed a bit, on the page) I guess it's back to the drawing table :)
https://www.mpbhba.org/business-resources/

city_scrapers/spiders/chi_ssa_20.py
guytet: As they've added 2021 meetings, things changed up a bit. Before the change, all the meetings took place in recent history, with some of them - having matching links on the right. 

Now, there are only future meetings on the left (2021), and everything on the right is past, with no matching meetings on the left (future). 
So it's a new challenge, so it seems.

guytet: @pjsier With the added 2021 meetings:

A meeting is either current/upcoming _without_ meeting minutes, but _with_ time data or, it's past _with_ meeting minutes and _without_ time data. 
What should we do in this case ?
https://www.mpbhba.org/business-resources/

An example to a past meeting link:
```
 <div class="et_pb_text_inner"><p>SSA 20 1st Quarter 2020 meeting minutes</p></div>
```
Which does not provide any time data, more like a link to archive.

pjsier: In general if the time data isn't present on past meetings we can go with a reasonable default based on the current meetings. It looks like 8am should work?

guytet: 8am works for me. What about the date  :) ?

pjsier: Ah missed that, for some of them it looks like the date is in the URL itself, and for the others we might just need to ignore them for now

PR #990, opened 2020-11-12
Swap broken link for new Sublime linter link

egfrank:
## Summary

I found a link in one of the docs that seemed corrupted: it seemed like it was supposed to lead to someone's blog but I think the address has since changed hands. I found another link that I believe fulfills the same purpose, which is to show an example of using automatic linting in setting up Sublime.

@pjsier it might be good to check out the new link https://janikarhunen.fi/three-steps-to-lint-python-3-6-in-sublime-text just to make sure you approve of the content, and let me know if it doesn't work for the purposes of these docs.

### DIFF
--- a/docs/_docs/setuphelp.md
+++ b/docs/_docs/setuphelp.md
@@ -92,4 +92,4 @@ Most text editors can be configured to fix code style issues for you based off o
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pjsier:
Thanks for finding this and the PR! I think this link does a good job replacing the content and also adding a bit more context, so I'm merging now

Issue #568, opened 2018-11-07
Spider: Chicago Special Service Area #35 Lincoln Ave


pjsier:
URL: https://www.lincolnparkchamber.com/businesses/special-service-areas/lincoln-avenue-ssa/ssa-administration/
Spider Name: `chi_ssa_35`
Agency Name: Chicago Special Service Area #35 Lincoln Ave

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

EthanHolleman:
If this spider is still open I would like to give it go

pjsier:
@EthanHolleman it's all yours, let us know if you need any help

AlessandroVol23:
Working on that :-)

cedarfall:
May I work on this one?

pjsier:
@cedarfall sure, all yours!

zahralizadeh:
May I work on this one?

sosolidkk:
If no one is working on this one, i would like to give it a try.

pjsier:
@sosolidkk sure! Assigning you now

sosolidkk:
Hey @pjsier , i'm sorry for not giving the attention to this issue. I've run into some problems with my work deadlines, but i'll submit a pull request soon.

PR #991, opened 2020-11-29
New/add spider chi_ssa_35 and it's test case

sosolidkk:
## Summary

**Issue:** #568 

## Checklist

- [X] Tests are implemented
- [X] All tests are passing

## Questions

I am having some doubts about this spider, because it has all the meetings time, date and place displayed on the website itself, but the meeting details for the current day that will happen are inside a `.pdf` document. So what i did was to put the `.pdf` document content displayed into the description field in the spider. Anyway, i don't know if what i did was the correct approach or if the right way would be to iterate over the `.pdf` documents and parse the data inside them as meetings.

### DIFF
--- a/city_scrapers/spiders/chi_ssa_35.py
+++ b/city_scrapers/spiders/chi_ssa_35.py
@@ -1,0 +1,122 @@
+from datetime import datetime
+from io import BytesIO, StringIO
+
+import requests
+from city_scrapers_core.constants import COMMISSION
+from city_scrapers_core.items import Meeting
+from city_scrapers_core.spiders import CityScrapersSpider
+from pdfminer.high_level import extract_text_to_fp
+from pdfminer.layout import LAParams
+
+
+class ChiSsa35Spider(CityScrapersSpider):
+    name = "chi_ssa_35"
+    agency = "Chicago Special Service Area #35 Lincoln Ave"
+    timezone = "America/Chicago"
+    start_urls = [
+        "https://www.lincolnparkchamber.com/"
+        + "businesses/special-service-areas/lincoln-avenue-ssa/ssa-administration/"
+    ]
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        content_div = response.css("div.content_block.content.background_white")
+        _ = content_div.css("h4::text").getall()
+        dates = content_div.css("ol").css("li::text").getall()
+        urls = content_div.css("a")
+        content = list(zip(dates, urls[: len(dates)]))
+
+        for item in content[:1]:
+            meeting = Meeting(
+                title=self._parse_title(item),
+                description=self._parse_description(item),
+                classification=self._parse_classification(item),
+                start=self._parse_start(item),
+                end=self._parse_end(item),
+                all_day=self._parse_all_day(item),
+                time_notes=self._parse_time_notes(item),
+                location=self._parse_location(item),
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Commission"
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        item_url = item[1].css("::attr(href)").get()
+        response = requests.get(item_url)
+
+        lp = LAParams(line_margin=0.1)
+        output_str = StringIO()
+        extract_text_to_fp(BytesIO(response.content), output_str, laparams=lp)
+        pdf_text = output_str.getvalue().replace("\n", "")
+
+        return pdf_text
+
+    def _parse_classification(self, item):
+        """Parse or generate classification from allowed options."""
+        return COMMISSION
+
+    def _clean_date_item(self, date_item):
+        date_item = date_item.split()
+
+        if len(date_item) == 5:
+            date_item[-1] = date_item[-1].replace(")", "")
+            date_item[-2] = date_item[-2].replace("(", "")
+            date_item.append(f"{datetime.today().year}")
+
+        elif len(date_item) == 3:
+            date_item.append("09:00 am")
+            date_item.append(f"{datetime.today().year}")
+
+        return " ".join(date_item)
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        date_item = self._clean_date_item(item[0])
+        date_obj = datetime.strptime(date_item, "%A, %B %d %H:%M %p %Y")
+
+        return date_obj
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "Lincoln Park Chamber of Commerce, 2468 N. Lincoln, Chicago",
+            "name": "Lincoln Park Chamber of Commerce",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        href = item[1].css("::attr(href)").get()
+        title = item[1].css("::text").get()
+
+        return [{"href": href, "title": title}]
+
+    def _parse_source(self, response):
+        """Parse or generate source."""
+        return response.url
--- a/tests/files/chi_ssa_35.html
+++ b/tests/files/chi_ssa_35.html
@@ -1,0 +1,711 @@
+<!DOCTYPE html>
+<html class="no-js" lang="en-US">
+
+<head>
+	<meta charset="UTF-8" />
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<title>SSA Administration - Lincoln Park Chamber of Commerce</title>
+	<link rel="profile" href="https://gmpg.org/xfn/11">
+	
+	<!-- This site is optimized with the Yoast SEO plugin v15.0 - https://yoast.com/wordpress/plugins/seo/ -->
+	<meta name="robots" content="index, follow, max-snippet:-1, max-image-preview:large, max-video-preview:-1" />
+	<link rel="canonical" href="https://www.lincolnparkchamber.com/businesses/special-service-areas/lincoln-avenue-ssa/ssa-administration/" />
+	<meta property="og:locale" content="en_US" />
+	<meta property="og:type" content="article" />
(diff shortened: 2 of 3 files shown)


city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        content_div = response.css("div.content_block.content.background_white")
+        _ = content_div.css("h4::text").getall()
```
luchiago, line 29: Why are you ignoring this variable?

sosolidkk: I was using it for the commission titles in case i needed it later, but since they were all the same i just ignore it.

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,122 @@
...
+            yield meeting
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        return "Commission"
+
+    def _parse_description(self, item):
+        """Parse or generate meeting description."""
+        item_url = item[1].css("::attr(href)").get()
+        response = requests.get(item_url)
```
pjsier, line 60: There isn't much meaningful information in the description, so we can drop this. We would want to include PDF parsing within scrapy so we can manage rate-limiting and handling responses in a unified way instead of using `requests` here as well

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,122 @@
...
+        elif len(date_item) == 3:
+            date_item.append("09:00 am")
+            date_item.append(f"{datetime.today().year}")
+
+        return " ".join(date_item)
+
+    def _parse_start(self, item):
+        """Parse start datetime as a naive datetime object."""
+        date_item = self._clean_date_item(item[0])
+        date_obj = datetime.strptime(date_item, "%A, %B %d %H:%M %p %Y")
```
pjsier, line 90: We've had some issues in the past with agencies putting the wrong day of the week, but the correct date, so we should ignore the weekday. Also, if we're using `%p` we'll need to use `%I` for the hour format

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,122 @@
...
+        return ""
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "Lincoln Park Chamber of Commerce, 2468 N. Lincoln, Chicago",
```
pjsier, line 109: It looks like this has varied a lot according to their site. We might want to just include "Confirm with agency" in the `name` and leave `address` blank

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,122 @@
...
+            "name": "Lincoln Park Chamber of Commerce",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        href = item[1].css("::attr(href)").get()
+        title = item[1].css("::text").get()
```
pjsier, line 116: The date isn't very relevant if we're already associating it with a meeting on a given date, so we should label it `Minutes` or `Agenda` based off of the section or URL text

city_scrapers/spiders/chi_ssa_35.py
pjsier: This can be removed if it isn't being used

pjsier (changes requested):
Thanks for the PR! Left some comments here

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Ok, i'll remove

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Ok, i'll make the changes

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: 👍🏻

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Could you elaborate a little more? I didn't understand @pjsier

pjsier: Sure, if we're looking at the document in the context of a meeting we already know the meeting date, so it isn't adding information as the document title. It's more helpful to see whether we're looking at an Agenda or Minutes, so ideally we could check which section of links we're in and assign the title that way, otherwise it looks like "Agenda" or "Minutes" are often included in the URL string so we could parse them that way

sosolidkk: Oh, i see. I'll check that out and make all the changes. Thank you for the explanation

sosolidkk:
Hello @pjsier , I was updating some stopped code and I made the corrections suggested by you. I also updated the code to make the year of each item correct, since it was fixed with a `datetime.today().year`. The only problem I still have is your change suggestion to `Minutes` and `Agenda` on title. I can't think of a way to do this dynamically, since on the page all I have is a `<h4>` which is followed by several `<p>` tags that contain the links inside. I kind of have to count and make it a more hard coded process. Do you have any better suggestions?

pjsier:
@sosolidkk thanks for the changes! I mentioned in the comment, but the `href` attribute usually contains "Agenda" or "Minutes" which is one way, and you could also loop through a selector that iterates through the immediate children of `.content` and updates the document name any time it runs into an `h4`

sosolidkk:
Hey @pjsier , sorry for the delay. I've updated this PR with the changes that you request. Now i'm iterating over all the inner elements of the body and separating the items in groups based on their `<h4>` title value, that can be Agenda, Schedule or Minutes.

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,147 @@
...
+
+            element_content["description"] = _description
+            element_content["type"] = _type
+            element_content["year"] = _year
+            content.append(element_content)
+
+        self._add_year_to_date_item(content)
+        self._parse_date_to_datetime(content)
+
+        for item in content:
```
pjsier, line 61: Is there a reason for the two separate looks here? It seems like we could consolidate them and have the individual parse functions handle getting details from a parent element rather than as separate steps of a loop

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,147 @@
...
+                links=self._parse_links(item),
+                source=self._parse_source(response),
+            )
+
+            meeting["status"] = self._get_status(meeting)
+            meeting["id"] = self._get_id(meeting)
+
+            yield meeting
+
+    def _add_year_to_date_item(self, content):
```
pjsier, line 83: It's a little hard to follow this when it's modifying an object in-place and not returning a value, could this be made more explicit in a function that returns a value instead of modifying one?

pjsier (changes requested):
Sorry for the delay, and thanks for the updates! I had a few general questions I wanted to address first, let me know if I can clarify anything

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Sure, I could do that.

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: I actually don't have a reason for using two for loops. For me it was more simpler to save all data into a list and the just run the data that is into the list, but i can change that too. Anyways, the suggestion that you made about search for the elements and iterating over them was just like the way i did? Or did you have something else in mind?

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Done!

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Done!

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,144 @@
...
+                meeting_content["date"] = element.css("::text").get()
+                if "href" in element.get():
+                    meeting_content["url"] = element.css("a::attr(href)").get()
+
+            if "<p><a" in element.get():
+                meeting_content["date"] = element.css("::text").get()
+                meeting_content["url"] = element.css("a::attr(href)").get()
+
+            meeting_content["description"] = _description
+            meeting_content["type"] = _type
```
pjsier, line 53: It looks like this isn't currently being used?

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,144 @@
...
+            meeting_content["year"] = _year
+
+            if "date" not in meeting_content:
```
pjsier, line 56: There may be some meetings that are both listed for agendas and minutes, and we'll need to combine those so that this doesn't return duplicates.

pjsier (changes requested):
Thanks for these updates, it's coming along! I left a few comments here

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: It isn't. I was looking for somewhere to put this info.

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,144 @@
...
+        """Parse start datetime as a naive datetime object."""
+        return item["date_obj"]
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return item["date_obj"] + timedelta(minutes=90)
+
+    def _parse_time_notes(self, item):
+        """Parse any additional notes on the timing of the meeting"""
+        return ""
```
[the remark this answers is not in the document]
sosolidkk: Are you talking about the `contact us` section?

```
Contact Us
Lincoln Park Chamber of Commerce
2468 N. Lincoln
Chicago, IL 60614
773 880 5200
```

city_scrapers/spiders/chi_ssa_35.py
```diff
@@ -0,0 +1,144 @@
...
+    def _parse_location(self, item):
+        """Parse or generate location."""
+        return {
+            "address": "",
+            "name": "Confirm with agency",
+        }
+
+    def _parse_links(self, item):
+        """Parse or generate links."""
+        return [{"href": item.get("url"), "title": item["date"]}]
```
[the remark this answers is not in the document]
sosolidkk: Can i just replace the title key with the info needed? Agenda or minutes

city_scrapers/spiders/chi_ssa_35.py
sosolidkk: Ok

simran-2501 (approved):

Issue #807, opened 2019-07-21
Spider: Illinois Sex Offender Management Board


pjsier:
URL: https://www2.illinois.gov/idoc/Pages/SexOffenderManagementBoard.aspx
Spider Name: `il_sex_offender_management`
Agency Name: Illinois Sex Offender Management Board

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

BobbyKarabinakis:
Hi There,

i would like to work on this issue

pjsier:
@BobbyKarabinakis sounds good!

mmachado95:
Hi, if this is still up for grabs I would like to start working on it.

pjsier:
@BobbyKarabinakis are you still working on this? If not it's all yours, @mmachado95

mmachado95:
@pjsier Can I try to move forward with this?

pjsier:
@mmachado95 go for it!

mmachado95:
Hey @pjsier quick question the meetings location and time is provided by the pdf that opens up after you click the meeting link.
In this case do I try to get that data from the PDF or do we ignore it?

pjsier:
Ideally I think we should try and parse these PDFs since I'm also seeing cancellation notices inside https://www2.illinois.gov/idoc/Documents/SOMB%20Meeting%20Agenda%202019%20August.pdf. It might only be necessary to actually parse the agenda content since the minutes content shouldn't be different

mmachado95:
Yeah some even have cancellation notes. 
Thanks for the clarification 👍  will look into parsing the pdf.

pjsier:
We've been using PyPDF2 for that mostly, and there's an example in `chi_human_relations`

mmachado95:
Oh that will make it easier, thanks for the help!

luchiago:
Is this issue still need help?

mmachado95:
Hey sorry I've been a bit delayed with this issue, work stuff got a bit in the way. 
I'm working on it still

luchiago:
@mmachado95 that's ok, no problem. Good job.

cvafaian:
Hi, I'd like to work on this if it's still available.

pjsier:
@cvafaian sure! Assigning you now

cvafaian:
Hello! 
[...]

PR #993, opened 2020-12-14
0807 spider il sex offender management

cvafaian:
## Summary

**Issue:** #807 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

I was able to extract all meeting info except for the one under SOMB Meeting Agendas labeled "September 16, 2020 – WebEx Meeting". This agenda meeting link redirects to a government account protected meeting recording that I could not access. I have excluded that meeting for now, but please let me know if you would like me to implement it differently!


pjsier:
Made a comment on #807, but I can take a look at this once we sort that out

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+from pdfminer.layout import LAParams
+
+
+class IlSexOffenderManagementSpider(CityScrapersSpider):
+    name = "il_sex_offender_management"
+    agency = "Illinois Sex Offender Management Board"
+    timezone = "America/Chicago"
+    custom_settings = {"ROBOTSTXT_OBEY": False}
+    start_urls = [
+        "https://www2.illinois.gov/idoc/" + "Pages/SexOffenderManagementBoard.aspx"
```
pjsier, line 20: Is there a reason for the "+" here? If the URL is longer than the allowed character limit it's usually more readable to add `#noqa` at the end of the line so that it doesn't trigger errors with `flake8` or `black` to override the checks there

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def parse(self, response):
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+
+        for link in response.css(
```
pjsier, line 31: `.soi-article-content` is a more general selector that's less likely to change and still seems to pull the same content, does that get at what you're looking for here?

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+                yield scrapy.Request(
+                    response.urljoin(link.attrib["href"]),
+                    callback=self._parse_documents,
+                    dont_filter=True,
+                )
+            # special case where link to video was uploaded rather than agenda
+            if "WebEx" in link_text:
+                yield scrapy.Request(
+                    link.attrib["href"], callback=self._parse_meta_video
+                )
```
pjsier, line 48: We'll want to get the dates from minutes as well so that we can link them back to agendas if they're available. We have an example of this in [`chi_lsc_advisory`](https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/city_scrapers/spiders/chi_lsc_advisory.py)

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+                continue
+            if "Agenda" in link_text:
+                # scrape agenda pdf for meeting information
+                yield scrapy.Request(
+                    response.urljoin(link.attrib["href"]),
+                    callback=self._parse_documents,
+                    dont_filter=True,
+                )
+            # special case where link to video was uploaded rather than agenda
+            if "WebEx" in link_text:
```
pjsier, line 45: To be safe, we can make this and the other title conditionals case-insensitive with `.lower()`. It seems possible that someone could type in "Webex" instead of "WebEx" as some point

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+        ).extract_first()
+        if not meta_text:
+            return
+        meta_text = meta_text.strip()
+        clean_text_vid = json.loads(meta_text)
+
+        start_time, end_time = self._video_start_end_time(clean_text_vid)
+
+        meeting = Meeting(
+            title="SEX OFFENDER MANAGEMENT BOARD",
```
pjsier, line 63: Since a lot of information is already in the agency name we can return "Management Board" to avoid duplicating info. It looks like there are a few for the "Training Committee", but since that hasn't happened in years we might be able to ignore it

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+            # special case where link to video was uploaded rather than agenda
+            if "WebEx" in link_text:
+                yield scrapy.Request(
+                    link.attrib["href"], callback=self._parse_meta_video
+                )
+
+    def _parse_meta_video(self, response):
+        """Parse through meta of response in JSON if link provided rather than PDF"""
+        meta_text = response.css(
+            "[type='application/json']:not(#bootstrapData)::text"
```
pjsier, line 53: To simplify this we could use the ID `extendedData` instead of negating the ID

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+        meeting = Meeting(
+            title="SEX OFFENDER MANAGEMENT BOARD",
+            description="",
+            classification=BOARD,
+            start=start_time,
+            end=end_time,
+            all_day=False,
+            time_notes="",
+            location={
+                "address": response.urljoin(""),
```
pjsier, line 71: Is there a reason for calling `response.urljoin` here? If there isn't a physical location we can leave this blank and use `response.url` in `source`

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+                "name": "WebEx Meeting (See meeting links to confirm)",
+            },
+            links={"title": "WebEx Meeting Link", "href": response.urljoin("")},
```
pjsier, line 74: `links` should always return a list (more info [in the docs](https://cityscrapers.org/docs/development/#links)), but in this case we can include the meeting link in `source` instead of `links` because it's where we parsed the content and not a separate attachment

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+        start_time = datetime.strptime("{}".format(date_time), "%Y-%m-%d %H:%M")
+
+        end_time = start_time + timedelta(minutes=scheduled_duration)
+
+        return start_time, end_time
+
+    def _parse_documents(self, response):
+        """Parse meeting information from agenda PDF"""
+
+        links_dict = {
```
pjsier, line 101: Repeating the comment from above that `links` should always be a list of dicts or an empty list

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+            "title": "Meeting Agenda",
+            "href": response.urljoin(""),
+        }
+        clean_text = self._cleanUpPDF(response)
+
+        if self._parse_status(clean_text) == CANCELLED:
+            meeting = Meeting(
+                title="SEX OFFENDER MANAGEMENT BOARD",
```
pjsier, line 109: Can be "Management Board"

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+    def _parse_documents(self, response):
+        """Parse meeting information from agenda PDF"""
+
+        links_dict = {
+            "title": "Meeting Agenda",
+            "href": response.urljoin(""),
+        }
+        clean_text = self._cleanUpPDF(response)
+
+        if self._parse_status(clean_text) == CANCELLED:
```
pjsier, line 107: Is there a reason for splitting this with a conditional? The location can be the same regardless of the status, and it looks like the rest of the info is unchanged

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+            meeting = Meeting(
+                title="SEX OFFENDER MANAGEMENT BOARD",
+                description="",
+                classification=BOARD,
+                start=self._parse_start(clean_text),
+                end=self._parse_end(clean_text),
+                all_day=False,
+                time_notes="See agenda to confirm details",
+                location={"address": "", "name": "Meeting Cancelled"},
+                links=links_dict,
```
pjsier, line 117: This one can contain the current agenda link at `response.url` in the list, since the agenda is a separate attachment that we're trying to pull information from

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+                source=self.start_urls[0],
+            )
+            meeting["status"] = self._get_status(meeting, text="Meeting is cancelled")
```
pjsier, line 120: Instead of hard-coding text to pass, the `_get_status` function is already checking whether any form of cancellation shows up in the text, so we can pass `clean_text` in the `text` kwarg.

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+    def _get_start_end_time(self, clean_text):
+        """Parse start and end time of meeting from agenda and returns group"""
+        pattern = re.compile(
+            r"[0-2]?[0-9]:[0-9][0-9][a-zA-Z][a-zA-Z]-"
+            r"[0-2]?[0-9]:[0-9][0-9][a-zA-Z][a-zA-Z]"
+        )
+        if pattern.search(clean_text) is not None:
+            return pattern.search(clean_text).group()
+        else:
+            # in a few specific agendas there are times formatted as '9-10am'
```
pjsier, line 200: It could be quicker to try multiple datetime formats or using `dateparse` instead of doing a lot of manual cleanup here, but let me know what you think!

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,249 @@
...
+        # there were different locations depending on if the meeting was
+        # a video conference, a phone conference, or a physical location.
+        # I hardcoded in the most frequent addresses as there were quite a few
+        # different locations, but I have made sure to note that one should conirm
+        # the address is indeed correct before using.
+
+        name = ""
+        address = ""
+
+        if "video conference" in clean_text.lower():
```
pjsier, line 233: If the location changes this much we can leave `address` blank and replace `name` with "See meeting details"

pjsier (changes requested):
Thanks for the PR! I left some comments here, let me know if anything is unclear

city_scrapers/spiders/il_sex_offender_management.py
cvafaian: Would the meeting link be in addition to the original source (the website itself) or would it replace it?

city_scrapers/spiders/il_sex_offender_management.py
cvafaian: I looked through all of the agendas, and for the most part (besides the video link) the PDFs have similar datetime formats! I feel comfortable with leaving it as is, but let me know otherwise.

city_scrapers/spiders/il_sex_offender_management.py
cvafaian: I have implemented it such that agenda dict with response.url as the href is part is part of links list. Please let me know if this is the correct interpretation!

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,234 @@
...
+        ).extract_first()
+        if not meta_text:
+            return
+        meta_text = meta_text.strip()
+        clean_text_vid = json.loads(meta_text)
+
+        start_time, end_time = self._video_start_end_time(clean_text_vid)
+
+        meeting = Meeting(
+            title="MANAGEMENT BOARD",
```
pjsier, line 72: We want these to be formatted for display, so we can do "Management Board" instead of all-caps here

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,234 @@
...
+                if "training" in agenda_dict["href"].lower():
+                    if "training" in meeting[1].lower():
+                        minutes_dict = {"title": "Meeting Minutes", "href": meeting[1]}
+                        links.append(minutes_dict)
+                elif "training" not in meeting[1].lower():
+                    minutes_dict = {"title": "Meeting Minutes", "href": meeting[1]}
+                    links.append(minutes_dict)
+        return links
+
+    def _cleanUpPDF(self, response):
```
pjsier, line 152: Missed this on the first pass, but generally Python follows snake case vs camel case, so this would be "_clean_up_pdf". It's a convention thing, but following it consistently makes code easier to follow

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,234 @@
...
+        """
+        `parse` should always `yield` Meeting items.
+
+        Change the `_parse_title`, `_parse_start`, etc methods to fit your scraping
+        needs.
+        """
+        # first go through and gather list of lists for meeting minutes
+        for link in response.css(".soi-article-content a"):
+            link_text = " ".join(link.css("*::text").extract())
+            if ("meeting minutes") in link_text.lower():
```
pjsier, line 34: Small thing but we don't need the parentheses around "meeting minutes"

tests/test_il_sex_offender_management.py
```diff
@@ -0,0 +1,94 @@
...
+test_response = file_response(
+    join(dirname(__file__), "files", "il_sex_offender_management.html"),
+    url="https://www2.illinois.gov/idoc/Pages/SexOffenderManagementBoard.aspx",
+)
+spider = IlSexOffenderManagementSpider()
+
+freezer = freeze_time("2020-12-12")
+freezer.start()
+
+parsed_items = [item for item in spider.parse(test_response)]
```
pjsier, line 29: To keep tests consistent we want to make sure that they're only parsing fixture files and not making any network requests. Usually the way we'll do this is to split out parsing methods from methods that `yield` requests so that we can parse fixture responses without having to chain requests together. One way to test this is to try and run your test file with wifi turned off on your computer and see if it still runs.

You can see an example of this setup in the tests for  [`chi_human_relations`](https://github.com/City-Bureau/city-scrapers/blob/main/tests/test_chi_human_relations.py)

city_scrapers/spiders/il_sex_offender_management.py
```diff
@@ -0,0 +1,234 @@
...
+        meeting["id"] = self._get_id(meeting)
+
+        yield meeting
+
+    def _make_links(self, agenda_dict, date_time_obj):
+        """Make list of agenda and minutes(if applicable)."""
+        links = []
+        if agenda_dict is not None:
+            links.append(agenda_dict)
+        for meeting in self.meeting_minutes:
```
pjsier, line 140: We've used `collections.defaultdict` for this in the past which has made it easier to check for links and return an empty list if they aren't present. We're doing that in [`chi_lsc_advisory`](https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/city_scrapers/spiders/chi_lsc_advisory.py)

pjsier (changes requested):
Thanks for making these changes! I have a few more comments here, and the tests might be the most significant part so let me know if I can help more there

tests/test_il_sex_offender_management.py
cvafaian: Hello, I actually was referencing `chi_human_relations` when I first wrote the test file! Like `chi_human_relations` test file, I downloaded the pdf in addition to the html that was already downloaded and ran the response on those files. Running pytest without wifi seems to work as it should with the current setup as well.  Could you provide some more details on where the problem area may be?

cvafaian:
I did a quick run of the program before pushing to make sure everything is in order, and it seems like the metadata for the WebEx video has since disappeared! While there is a proper response for the link, there is only an empty dictionary returned for where the metadata once was. I tried checking through the other information returned in the response to see if the metadata may have moved, but I wasn't able to find anything. I have pushed changes for the revisions suggested, but the current spider is no longer making a meeting for the latest WebEx meeting. How would you like me to proceed?

tests/test_il_sex_offender_management.py
pjsier: Gotcha, thanks for clarifying! Ordinarily we're looking to check the items from the first run of `parse`, but you're right that since you're only checking the count and then parsing the PDF separately that this works without it. I'll resolve this

pjsier:
Thanks for the updates! It was working before, so I think it's fine to leave the code as-is in case it's added back later

pjsier (approved):

Issue #847, opened 2019-10-03
Spider: Illinois Department of Corrections Advisory Board


pjsier:
URL: https://www2.illinois.gov/idoc/aboutus/advisoryboard/Pages/default.aspx
Spider Name: `il_corrections`
Agency Name: Illinois Department of Corrections Advisory Board

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

Some information will need to be parsed from PDFs, so it could be useful to look at [`chi_human_relations`](https://github.com/City-Bureau/city-scrapers/blob/fbddf7cf785f261ea1829706dd58fc7ea5777a9a/city_scrapers/spiders/chi_human_relations.py) for an example of how to handle this

jairathnem:
I am working on this one. PDF documents for the different dates are inconsistent and don't always follow a format. Any suggestions on that?

zuzaman:
I just wanted to check this issue and saw that the https://www2.illinois.gov/ website is not reachable. do you possible know if they moved to a new host?

pjsier:
@zuzaman we've had issues with people outside of Illinois and/or the US accessing that URL. Not sure if that applies to you, but it's showing up for me still

zuzaman:
@pjsier  Ok, thanks. Well then I will leave this issue as I am from the EU. :)

SubtleHyperbole:
I am trying to see whether I could do this as my first contribution, but I’m running into a problem.  I follow the instructions you provided on cloning the git locally and then using pipenv to create a virtual environment on a new branch with this issue # in tfw format suggested, with all the necessary dependencies, 

I am running into two issues.  The first is that, at this point, I open up the spider file in my programming environment of choice, Jupiter notebook, and quickly discover that the dependencies installed in the virtual environment on the command line don’t apply to the Jupiter kernels I can create there.  Ok, so I try to instead open up the files in Atom but run into the same issue.

This led me to discover my 2nd issue.  I decided ok what if I just install the missing decencies so I can work on the code outside of the command terminal.  I see that I am missing city_scrapers_core but when I try to install that I see that I am missing a dependency called Legistar.  I try to install legistar and learn that it is not available me to Mac OSX users.  

[...]

PR #994, opened 2020-12-14
0847 spider corrections advisory board

cherdeman:
## Summary
Completed spider for Illinois Department of Corrections Advisory Board meetings. 

**Issue:** #847 

## Checklist

- [x] Tests are implemented
- [x] All tests are passing

## Questions

I keep getting generator objects returned from the scrapy requests rather than the actual object that I want, particularly `pdf_text`. This is certainly a Scrapy usage issue - do ya'll have input on how I can either get the behavior I expect or rework to be more in line with the package?

Edit 12/31: All tests are passing but I'm getting a timezone warning, not sure how big a concern that should be.


city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+            )
+
+            if date is not None:
+                if date not in self.links:
+                    self.links[date] = {}
+                for item in ["Notice", "Agenda", "Minutes"]:
+                    if item in link.attrib["href"]:
+                        self.links[date][item] = self.url_base + link.attrib["href"]
+
+        return self.links
```
pjsier, line 153: It looks like this is both setting and returning `self.links`. It could be more clear to create the dictionary in here and return it, setting it to `self.links` so that it's clear that we aren't expecting it to contain anything.

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+        for link in response.css("a"):
+            date = link.re_first(
+                r"""((Jan(uary)?|Feb(ruary)?|Mar(ch)?|Apr(il)?|
+            May|Jun(e)?|Jul(y)?|Aug(ust)?|Sep(tember)?|Oct(ober)?|Nov(ember)?|
+            Dec(ember)?)\s+\d{1,2},\s+\d{4})|((1[0-2]|0?[1-9])/(3[01]|
+            [12][0-9]|0?[1-9])/(?:[0-9]{2})?[0-9]{2})"""
+            )
+
+            if date is not None:
+                if date not in self.links:
```
pjsier, line 147: This could be simplified by using `defaultdict` instead of a regular `dict`. Creating this as `defaultdict(dict)` will automatically set the value of any key to a dictionary when it's accessed so you can set `links[date][key]` directly without having to check whether it exists first

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+                    self.links[date] = {}
+                for item in ["Notice", "Agenda", "Minutes"]:
+                    if item in link.attrib["href"]:
+                        self.links[date][item] = self.url_base + link.attrib["href"]
```
pjsier, line 151: We should be able to use `response.urljoin` instead of setting `url_base` statically at the beginning to combine the domain with any partial URL

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+
+class IlCorrectionsSpider(CityScrapersSpider):
+    name = "il_corrections"
+    agency = "Illinois Department of Corrections Advisory Board"
+    timezone = "America/Chicago"
+    start_urls = [
+        "https://www2.illinois.gov/idoc/aboutus/advisoryboard/Pages/default.aspx"
+    ]
+
+    def __init__(self):
```
pjsier, line 21: It looks like we don't need to set any of these on initialization, and other than `links` we may not need to set any of these directly on the spider itself because it looks like they could be handled as normal variables. Let me know if I'm missing something there though!

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+                )
+            elif "Minutes" in self.links[date].keys():
+                yield scrapy.Request(
+                    self.links[date]["Minutes"],
+                    callback=self._meeting,
+                    cb_kwargs=dict(date=date),
+                )
+
+    def _meeting(self, response, date):
+        self.pdf_text = self._parse_pdf(response)
```
pjsier, line 51: Mentioned above, but it could be more clear to set this as `pdf_text` rather than `self.pdf_text` and pass it as an argument to the functions where it's needed so that this is a little more clear

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+            source=response.url,
+        )
+
+        meeting["status"] = self._get_status(meeting)
+        meeting["id"] = self._get_id(meeting)
+        yield meeting
+
+    def _parse_pdf(self, response):
+        """Parse dates and details from schedule PDF"""
+        print("prase pdf)")
```
pjsier, line 71: Guessing this is left over from testing but we can drop it

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+                for item in ["Notice", "Agenda", "Minutes"]:
+                    if item in link.attrib["href"]:
+                        self.links[date][item] = self.url_base + link.attrib["href"]
+
+        return self.links
+
+    def _parse_links(self, date):
+        """Parse or generate links."""
+        link_list = []
+        for key, value in self.links[date].items():
```
pjsier, line 158: Related to some of the other comments, but if we use `defaultdict(list)` instead of `defaultdict(dict)` we could access this as `self.links[date]` without having a `_parse_links` function at all. It seems like we don't need the dict functionality other than to establish date uniqueness, which we should have at this point

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,160 @@
...
+            start=self._parse_times(date),
+            end=self._parse_times(date, False),
+            all_day=False,
+            time_notes="",
+            location=self._parse_location(),
+            links=self._parse_links(date),
+            source=response.url,
+        )
+
+        meeting["status"] = self._get_status(meeting)
```
pjsier, line 65: If cancellation is ever mentioned in meeting text, we can pass `pdf_text` to this as the `text` kwarg since we have some built-in handling of cancelling meetings based on text provided there

tests/test_il_corrections.py
```diff
@@ -0,0 +1,99 @@
...
+from city_scrapers.spiders.il_corrections import IlCorrectionsSpider
+
+test_pdf_response = file_response(
+    join(dirname(__file__), "files", "il_corrections.pdf"),
+    url="https://www2.illinois.gov/idoc/aboutus/advisoryboard/Documents/"
+    "Agenda%20-%20November%204th-2019.pdf",
+    mode="rb",
+)
+
+print(test_pdf_response)
```
pjsier, line 17: Looks like this is left over from debugging and can be removed

tests/test_il_corrections.py
```diff
@@ -0,0 +1,99 @@
...
+freezer.start()
+
+parsed_dates = spider._parse_all_links(test_response)
+test_generator = spider._meeting(test_pdf_response, "November 4, 2019")
+test_meeting = next(test_generator)
+
+freezer.stop()
+
+
+def meeting_count(parsed_dates):
```
pjsier, line 36: I think the test runner might not treat functions not starting in `test_` as tests, so this can become `test_meeting_count`

pjsier (changes requested):
Thanks for making these updates! I left some comments but this is coming along really well

city_scrapers/spiders/il_corrections.py
cherdeman: Ahhh yeah some of these are from me going back and forth to figure out passing the responses through correctly. I'll clean this up!

city_scrapers/spiders/il_corrections.py
cherdeman: This one is because the spider originates at `https://www2.illinois.gov/idoc/aboutus/advisoryboard/Pages/default.aspx` but the relative links for each start at the `idoc` section, so if I use `response.urljoin` I get something like `https://www2.illinois.gov/idoc/aboutus/advisoryboard/Pages/idoc/aboutus/advisoryboard/Documents/20090713_Advisory_Board_Minutes.pdf`. I figured it was easier to define statically than parse the response url, or is there a built in way to select just the domain portion from the response?

city_scrapers/spiders/il_corrections.py
cherdeman: I'm also using the dictionary structure to decide which document to parse based on existence - I want Agenda if it's there, or else minutes. That functionality was a little easier with the dictionary structure than going straight to the final list of dicts

city_scrapers/spiders/il_corrections.py
cherdeman: Based on other commends I'm leaving links and url_base for now

city_scrapers/spiders/il_corrections.py
cherdeman: Hmmm I don't think this would be the case? Since we're only parsing Agendas or Minutes they're for meetings that happened, I'm worried that the word cancel might be in the minute text for some other reason and falsely change the status. Though not a big change if we want to try it out!

cherdeman:
> Thanks for making these updates! I left some comments but this is coming along really well

Thanks for the review @pjsier! I think I've changed or responded to all of your comments - let me know if there are other issues to resolve.

city_scrapers/spiders/il_corrections.py
pjsier: It's pretty rare that they include it, but not a big deal! We can leave it out

city_scrapers/spiders/il_corrections.py
pjsier: Gotcha, `response.urljoin` should handle this though. It uses a similar setup to the browser to parse relative links including the current page subdirectory. It's a wrapper around a standard library function with more info [in the Python docs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin). When I run this locally with this line, it's showing the URLs as equal, but let me know if I'm missing some!

```python
print(
    self.url_base + link.attrib["href"]
    == response.urljoin(link.attrib["href"])
)
```

cherdeman: Yeahhh you're right you're right, made this change!

city_scrapers/spiders/il_corrections.py
```diff
@@ -0,0 +1,157 @@
...
+class IlCorrectionsSpider(CityScrapersSpider):
+    name = "il_corrections"
+    agency = "Illinois Department of Corrections Advisory Board"
+    timezone = "America/Chicago"
+    start_urls = [
+        "https://www2.illinois.gov/idoc/aboutus/advisoryboard/Pages/default.aspx"
+    ]
+
+    def __init__(self):
+        self.links = {}
```
pjsier, line 23: Small thing but this doesn't need to be set on initialization since it'll be overwritten on the call to `parse` anyway

cherdeman: Oh interesting, I thought it had to be instantiated at the outset. thanks!

pjsier:
Thanks for the changes! Hopefully last thing, I just noticed that there's an error in the output logs for running the scraper on the current site. More info here https://github.com/City-Bureau/city-scrapers/runs/1652131797#step:10:346, but it should show up in the output if you merge the latest changes from `main` (to hide the pdfminer logs) and run the scraper

cherdeman:
> Thanks for the changes! Hopefully last thing, I just noticed that there's an error in the output logs for running the scraper on the current site. More info here https://github.com/City-Bureau/city-scrapers/runs/1652131797#step:10:346, but it should show up in the output if you merge the latest changes from `main` (to hide the pdfminer logs) and run the scraper

Ah yup I see - it's looking like most of these are coming from missing times - do you have a preference for making 10:30am or 12:00am the default start? It looks like the others usually start at 10:30, though maybe we'd rather make it clear that its a guess with 12:00. Ill make the end 1.5 hours later.

pjsier:
Right now any meetings with `None` as the end time will automatically set it to 2 hours after the start so that part's fine to ignore. For the start time, it's up to you! If it seems like 10:30 is a safe assumption that works for me, otherwise midnight is probably best

pjsier (approved):

Issue #914, opened 2019-10-24
Spider: Illinois Finance Authority


pjsier:
URL: https://www.il-fa.com/
Documents URL: https://www.il-fa.com/public-access/board-documents/
Spider Name: il_finance_authority
Agency Name: Illinois Finance Authority

See the [contribution guide](https://cityscrapers.org/docs/development/#contribute) for information on how to get started

aneesh404:
I'd like to work on this issue

pjsier:
@aneesh404 sounds good!

aneesh404:
Hi! I'm sorry I'm not getting time to work on this issue. Please feel free to assign it to someone else.

janeskim:
Hi! It looks like this issue isn't claimed. Is it ok if I work on this issue?

pjsier:
@janeskim all yours!

mesterhammerfic:
If this one is open I'm going to work on it

pjsier:
@mesterhammerfic sounds great! Assigning you now

ledaliang:
Hi, I was wondering If I could work on this issue if it hasn't been active recently. Thanks!

pjsier:
@ledaliang thanks for your interest! We try to limit contributors to [one issue at a time](https://github.com/City-Bureau/city-scrapers/blob/master/CONTRIBUTING.md#1-find-an-open-issue-to-work-on), but once your other PR is merged you can feel free to work on this one

PatrickKlingler:
Hi there I've only just asked for a Slack invite, but could I start working on this now?

pjsier:
@PatrickKlingler sure! Marking it claimed now

PatrickKlingler:
Hey Patrick would it be possible to add another PDF parser?

The PyPDF2 parser does not seem to work for the PDFs on IFA's website, i.e. it returns an empty string. I copied this code to parse the PDF: https://github.com/City-Bureau/city-scrapers/blob/main/city_scrapers/spiders/il_pollution_control.py#L103

Apparently PyPDF2 is limited to certain kinds of PDF encodings: https://stackoverflow.com/questions/30272269/python-text-extraction-does-not-work-on-some-pdfs

I ended up using `pdfplumber` and that works but it would introduce another dependency.

pjsier:
[...]

PR #995, opened 2020-12-20
914 spider il finance authority

solisedwin:
## Summary

**Issue:** #914

## Checklist

- [ ] Tests are implemented
- [ ] All tests are passing

## Comments
Glad to say after months I'm finally doing working on this spider as my first ever pull request. Thanks for the opportunity as I learned a lot. I used scrapy meta arguments for each crawl since the self keyword didn't make each item unique and would repeat values. The code for tests might seem "weird" but scarped information can vary when working with PDFs so I tried a general approach.


.gitignore
```diff
@@ -22,6 +22,7 @@ parts/
 sdist/
 var/
 wheels/
+city_scrapers/get-pip.py
```
pjsier, line 25: Usually we'll want to keep unrelated .gitignore changes out of PRs. If you want to ignore something locally you can add it to `.git/info/exclude` in your repo

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+from pdfminer.pdfparser import PDFSyntaxError
+
+
+class IlFinanceAuthoritySpider(CityScrapersSpider):
+    name = "il_finance_authority"
+    agency = "Illinois Finance Authority"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.il-fa.com/public-access/board-documents/"]
+
+    def __init__(self, *args, **kwargs):
```
pjsier, line 20: Is there a reason for leaving this in? Right now it has no effect, but not sure if it was used earlier

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+    def _parse_location(self, pdf_content):
+        try:
+            """Parse or generate location."""
+            address_match = re.search(
+                r"(?:in\s*the|at\s*the) .*(\. | \d{5})", pdf_content
+            )
+            address = address_match.group(0)
+            name = re.findall(r"(?:in\s*the|at\s*the).*?,", pdf_content)[0]
+        except Exception:
+            address = "Address Not Found"
```
pjsier, line 138: We can just return blank strings if that's the case, but it looks like the location is pretty consistent so it could be easier to use `_validate_location` and raise an exception if the location isn't a set address like we do here https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/city_scrapers/spiders/chi_ssa_61.py#L65

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+
+            pdf_content = out_str.getvalue().replace("\n", "")
+            # Remove duplicate spaces
+            clean_text = re.sub(r"\s+", " ", pdf_content)
+            # Remove underscores
+            clean_text = re.sub(r"_*", "", clean_text)
+            return clean_text
+
+        except PDFSyntaxError as e:
```
pjsier, line 75: It looks like this can return `None`, does this happen often? If so we should allow it to raise the error normally

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def _parse_end(self, item):
+        """Parse end datetime as a naive datetime object. Added by pipeline if None"""
+        return None
+
+    def _parse_all_day(self, item):
+        """Parse or generate all-day status. Defaults to False."""
+        return False
+
+    def _parse_links(self, link, title):
```
pjsier, line 165: It looks like there are multiple links here, so ideally we would want to pull the notice, minutes, and any other information as links even if we aren't parsing them.

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+    name = "il_finance_authority"
+    agency = "Illinois Finance Authority"
+    timezone = "America/Chicago"
+    start_urls = ["https://www.il-fa.com/public-access/board-documents/"]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def parse(self, response):
+        for item in response.css("tr:nth-child(1n+2)"):
```
pjsier, line 24: We don't have to use this much, but because there are so many meetings here it would be good to use our `CITY_SCRAPERS_ARCHIVE` setting so that we don't have to pull the entire meeting list every time. Here's an example https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/city_scrapers/spiders/chi_zoning_board.py#L33, we would probably want to check within 6 months to a year from the current date

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+        pdf_text = self._parse_agenda_pdf(response)
+        location = self._parse_location(pdf_text)
+        time = self._parse_start(pdf_text)
+        meeting_dict = dict()
+        meeting_dict["title"] = response.meta["title"]
+        meeting_dict["date"] = response.meta["date"]
+        meeting_dict["location"] = location
+        meeting_dict["time"] = time
+
+        yield scrapy.Request(
```
pjsier, line 49: Is this last request necessary? It looks like we're trying to `yield` the meeting after parsing the PDF, but is there other information here that we need to get from `_parse_meeting`?

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+            title = item.css("td:nth-child(3)::text").extract_first()
+            return title
+        except TypeError:
+            return ""
+
+    def _parse_classification(self, title):
+        """Parse or generate classification from allowed options."""
+        if "Comm" in title:
+            return COMMITTEE
+        if "Board" in title:
```
pjsier, line 125: We can default to `BOARD` instead of Commission since it seems like that's the majority of these

tests/test_il_finance_authority.py
```diff
@@ -0,0 +1,69 @@
...
+
+def test_name():
+    location = spider._parse_location(pdf_text)
+    assert "Authority’s Chicago Office" in location["name"]
+
+
+def test_location():
+    location = spider._parse_location(pdf_text)
+    assert (
+        "in the Authority’s Chicago Office, 160 North LaSalle Street, Suite S-1000, Chicago, Illinois 60601"
```
pjsier, line 46: We'll want to simplify this so that only the street address is included in the `address` value of the location dict. This could be fixed by using a static location checked by `_validate_location` mentioned in another comment here

city_scrapers/spiders/il_finance_authority.py
```diff
@@ -0,0 +1,171 @@
...
+    def _get_pdf_link(self, item):
+        pdf_tag = item.css("td:nth-child(4) > a")
+        if not (pdf_tag):
+            return None
+        pdf_link = pdf_tag[0].attrib["href"]
+        return pdf_link
+
+    def _parse_title(self, item):
+        """Parse or generate meeting title."""
+        try:
```
pjsier, line 115: Small thing, but if possible it would be good to replace "Comm" with "Committee" for committee meetings. To be safe we'd want to replace it only when it's at the end of the title, so something like this could work

```python
title = re.sub(r"Comm$", "Committee", title)
```

tests/test_il_finance_authority.py
```diff
@@ -0,0 +1,69 @@
...
+
+parsed_items = [item for item in spider.parse(test_response)]
+
+pdf_text = spider._parse_agenda_pdf(test_pdf_response)
+
+freezer.stop()
+
+
+def test_time():
+    time = spider._parse_start(pdf_text)
```
pjsier, line 34: This should work, but is there a reason for not testing against a Meeting item? Usually that's the way we've done it to make sure the whole process works

pjsier (changes requested):
Thanks for the PR, and sorry for the delay! This is looking great so far, especially when the PDF scrapers tend to be pretty tricky. I left some comments here, let me know if anything is unclear

tests/test_il_finance_authority.py
solisedwin: To be honest, I'm not sure how to test the id since it  doesn't appear from parsed_items. Perhaps I'm missing the obvious from my novice coding skills, but I can't seem to think of way to get all the information to recreate//test the id. Since the process of yielding a a meeting object is very specific in order and 'tricky' for this spider.

city_scrapers/spiders/il_finance_authority.py
solisedwin: Are you talking about the the other pdf links on the website? Could you elaborate more? 
Is there an example you could provide? Thanks

city_scrapers/spiders/il_finance_authority.py
solisedwin: The last request is necessary for the code to work.

city_scrapers/spiders/il_finance_authority.py
solisedwin: Can you explain more about the meaning behind this idea? I added it to my code and now it doesnt work. Also more about the settings object? Thank you

city_scrapers/spiders/il_finance_authority.py
pjsier: Sure, I'm seeing up to 4 PDF links in each row for the Agenda, Board Book, Minutes, and Voting Record. Ideally we'll want to include all of those in the `links` list

city_scrapers/spiders/il_finance_authority.py
pjsier: Sure! We want to be careful about spamming sites with a ton of requests at once, and typically when we scrape a site we're only interested in the last few past meetings and next few upcoming. To reduce the amount of requests we make, as well as simplify the output for anyone using the feeds directly, we try to set ranges of time relative to the current date that we're interested like everything in the past year in that example.

Scrapy's settings are a way for managing configuration across spiders like where the output is written or how quickly requests should be made. You can find more info on them in the [scrapy documentation on settings](https://docs.scrapy.org/en/latest/topics/settings.html).

Could you explain more about what isn't working? It's hard for me to debug without an example, but in general all the `CITY_SCRAPERS_ARCHIVE` settings is doing is giving us a boolean that can be put inside a conditional, so it doesn't have to work the same way as the example

city_scrapers/spiders/il_finance_authority.py
pjsier: Could you explain more about what you mean? When you create a new request to `response.url`, it looks like you're submitting a second request to the URL you parsed the response from rather than going to a new page

tests/test_il_finance_authority.py
pjsier: Gotcha, we aren't typically using the `meta` attribute on requests, but it seems like the right approach here and testing it can be tricky.

Here's an example of how we're doing that for the `chi_buildings` scraper where we're using it in a similar way. Let me know if that helps!

https://github.com/City-Bureau/city-scrapers/blob/c42620480f9184ca74885e2a3c6ef576faba3eb7/tests/test_chi_buildings.py#L22-L32

pjsier (changes requested):
Thanks for the responses! I batched some replies here

PR #1150, opened 2025-04-23
Remove keep alive step to prevent blocking/failing cron

duongcao-ea:
This PR removes the keepalive-workflow step from the cron workflow.


lamle-ea (approved):
LGTM

PR #1161, opened 2025-12-26
Add keepalive job to scheduled workflows

duongcao-ea:
Add workflow-keepalive job to cron.yml and archive.yml to prevent GitHub from automatically disabling scheduled workflows after 60 days of inactivity.


lamle-ea (approved):
LGTM

PR #1181, opened 2026-09-04
🏗️ Build spider: Chicago Police District Councils

msrezaie:
## What's this PR do?
This PR adds a scraper for Chicago Police District Councils.

## Why are we doing this?
Requested by [City Bureau](https://airtable.com/appawQXzMQREfmVEW/tblV2uXZZOeQSqjXx/viwiCr3krf5HzWpFZ?blocks=hide).

After installing the project using `pipenv`:

1. Activate the virtual env and enter the pipenv shell:
```
pipenv shell
```
2. Install project dependencies:
```
pipenv sync --dev
```
3. Run the spider:
```
scrapy crawl chi_police_district_councils -O test_output.csv
```
4. Monitor the output and ensure no errors are raised.
5. Inspect `test_output.csv` to ensure the data looks valid.
6. Ensure all tests pass

PR #1183, opened 2026-09-16
Add Airtable spider sync workflows

lamle-ea:
## Why

Every new spider PR in the other `city-scrapers-*` repos (Indianapolis, Philadelphia, Atlanta, Dallas, Minneapolis) triggers a **Parse spiders** workflow and then **Sync spiders to Airtable**, which creates the spider's row in the Airtable Slug table and links it to the backlog record. This repo has neither, so Chicago spiders never get a Slug row.

Concrete effect: the backlog record for Chicago Police District Councils (#1181) has no linked spider, so the DOC-2706 staging automation reports it as "record has no linked spiders" and can never flip its status, even after the PR merges and the Chicago leg is unblocked by City-Bureau/documenters#2384.

## What

Two stub workflows, copied byte-for-byte from `city-scrapers-indianapolis`:

- `.github/workflows/parse-spiders.yml` — runs on PRs touching `city_scrapers/spiders/**.py`, calls the shared parse workflow in `city-scrapers-core`.
- `.github/workflows/sync-airtable.yml` — runs when the parse succeeds, calls the shared sync workflow. It never reads PR code, only the parse artifact.

## Before this works

The sync needs three Actions variables and one secret visible to this repo: `CS_AIRTABLE_BASE_ID`, `CS_SLUGS_TABLE_ID`, `CS_BACKLOG_TABLE_ID`, and `CS_AIRTABLE_PAT`. They are not repo-level here, and I can't read org-level config to confirm whether they're org variables scoped to selected repos. Whoever provisioned them for Indianapolis should confirm this repo is included. The first sync run after merge is the check: it fails at the sync step if any are missing.

## After merge

Pushing any commit to #1181 (or reopening it) triggers the parse and creates the `chi_police_district_councils` row automatically. Until then the row can be created by hand.

Note: this PR touches `.github/`, so the staging automation will correctly skip it as a protected path; it needs a human merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lamle-ea:
Closing for now. The two stubs are ready on the add-airtable-spider-sync branch, but the sync needs the CS_* Actions variables and CS_AIRTABLE_PAT secret visible to this repo first, and that provisioning hasn't been agreed yet. Will reopen once it is.

msrezaie:
Thanks @lamle-ea for adding this PR. On the question of provisioning variables and secrets, they don't need to exist per repo (further explained [here](https://github.com/City-Bureau/city-scrapers-core/pull/34)). They only need to be set as org level secrets, which is already done (related [discussion](https://codethedream.slack.com/archives/C0833NP35RN/p1778080790982079)).

There is an open [PR](https://github.com/City-Bureau/city-scrapers/pull/1182) which adds the consumer stub YML files, so you can merge that one. Thanks.

